PR #1, opened 2016-01-03
Various Improvements

mborgerson:
I've made several changes that make NXDK a bit more robust. In summary:
* Added extract-xiso to the tools for easy XISO creation.
* Restructured some things:
  * Moved library code into the lib directory.
  * Moved the 0ldschoo1 sample code into the samples directory.
* Added changes to enabled out-of-tree builds.
* Added a "Hello World" sample.
* Added Fixes to support building on Linux (specifically, Ubuntu).
* Added finer dependency tracking (via generation of .d files using the compiler).
* Fixed some code issues causing warnings, etc.
* Added a different debug console font.

Thanks,
Matt


espes:
Lots of good stuff, thanks! I'll take a look when I have access to a real internet connection again.

mborgerson:
Sure, thanks! And sorry about the jumbo-sized pull-request.

Makefile
```diff
@@ -1,92 +1,145 @@
...
-%.inl: %.vp
-	$(VP20COMPILER) '$<' > '$@'
-
-%.inl: %.fp
-	$(FP20COMPILER) '$<' > '$@'
+ifneq ($(GEN_XISO),)
+TARGET += $(GEN_XISO)
+endif
+
+V = 0
```
espes: I'd really prefer keeping the Makefiles simple - get rid of the $(VE) and $(STDOUT_TO_NULL) debugging stuff, and ditch distclean?

README.md
```diff
@@ -1,73 +1,108 @@
-nxdk - the *new* open source xdk
+NXDK - *the new open source XDK*
```
espes: I prefer nxdk to NXDK...

README.md
```diff
@@ -1,73 +1,108 @@
...
-- A (windows) copy of the NVIDIA Cg compiler 1.3 from the NVIDIA SDK is bundled.
+### Generate XISO
+To generate an ISO file that can be burned to a disc, or passed to the XQEMU
+emulator via the `-drive index=1,media=cdrom,file=/path/to/your.iso` parameter,
+define the `GEN_XISO` variable with the name of the ISO to be created in your
+project Makefile. For example:
 
-Code Overview
--------------
+    GEN_XISO=$(XBE_TITLE).iso
```
espes: confusing. Just generate it by default?

Makefile
```diff
@@ -1,92 +1,145 @@
...
+	@echo "[ DEP      ] $@"
+	$(VE) set -e; rm -f $@; \
+	$(CC) -M -MM -MG -MT '$*.obj' -MF $@ $(CFLAGS) $<; \
+	echo "\n$@ : $^\n" >> $@
+
+%.inl: %.vs.cg
+	@echo "[ CG       ] $@"
+	$(VE) $(CGC) -profile vp20 -o $@.$$$$ $< $(STDOUT_TO_NULL) && \
+	$(VP20COMPILER) $@.$$$$ > $@ && \
+	rm -rf $@.$$$$
```
espes: keep the intermediate around?

README.md
```diff
@@ -1,73 +1,108 @@
...
-- A recent version of [clang](http://clang.llvm.org/) and [lld](http://lld.llvm.org/)
+#### Linux (Ubuntu)
+On Ubuntu, the requried packages can be downloaded from the standard repositories:
 
-- [GNU bison](https://www.gnu.org/software/bison/) and [flex](http://flex.sourceforge.net/)
+    sudo apt-get install build-essential flex bison g++ clang wine binutils-mingw-w64 git
 
-- *Temporarilly because I'm lazy*, [wine](https://www.winehq.org/) to invoke the NVIDIA Cg compiler
+### Download NXDK
+    git clone https://github.com/xqemu/nxdk.git
```
espes: --recursive, why not

mborgerson: Hah, I can take this out if you really want

Makefile
mborgerson: Possibly. The $@.$$$$ will generate a new file each time it runs bc $$$$ gets translated to the pid. We could possibly do a string substitution to keep better track of the intermediates.

README.md
mborgerson: I don't really think we should generate ISOs by default. It's pretty simple though, just add GEN_XISO=filename.iso to your Makefile. Or you can do it when you run make. e.g. `make GEN_XISO=test.iso`.

README.md
mborgerson: Fair enough. I was feeling pedantic at the time. I'll change it back.

Makefile
mborgerson: I agree with keeping the Makefile simple. I'm ok with ditching distclean if you really want.

That said, the reason I added `$(VE)` and `$(STDOUT_TO_NULL)` is as we keep adding library code the number of files that go into a build shoots up, and when you're compiling it generates heaps of output.

The `$(VE)` bit is nice because if you want to see all the build commands when your build you just run `make V=1`. (Same goes with `$(STDOUT_TO_NULL)` which squelches noisy tools like extract-xiso.) Otherwise, it'll just spit out the filename name of what's currently building. This is pretty common practice nowadays (Linux, coreboot, etc.).

If you feel strongly about it, we can remove this feature.

README.md
espes: sorry like, add --recursive since we have submodules now

mborgerson: Oh, yeah, sure.

mborgerson: Fixed

README.md
mborgerson: Fixed

espes:
Thanks!

PR #2, opened 2016-02-04
Networking

mborgerson:
This PR adds networking support to NXDK via [PktDrv](http://sourceforge.net/projects/openxdk/files/Patches%20and%20Drivers/AFL%20packet%20driver/) and the [LwIP](http://savannah.nongnu.org/projects/lwip/) networking libraries. A driver was created to interface LwIP and PktDrv. A simple HTTP server is included in the samples.


espes:
Thanks!

PR #4, opened 2017-04-16
Fix warnings

JayFoxRox:
I've fixed some warnings (or errors even) to make using NXDK easier.
Changes should be pretty self-explanatory

(There's a ton of warnings left - this was just the most annoying / important changes)

### DIFF
--- a/lib/net/nforceif/src/sys_arch.c
+++ b/lib/net/nforceif/src/sys_arch.c
@@ -91,7 +91,7 @@ sys_now(void)
   asm __volatile__ ("rdtsc; divl %1"
                     : "=a" (now)
                     : "r" (freq)
-                    : "%eax", "%edx");
+                    : "%edx");
   return now;
 }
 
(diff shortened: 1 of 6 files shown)


espes:
thanks!

lib/net/nforceif/src/sys_arch.c
```diff
@@ -91,7 +91,7 @@ sys_now(void)
   asm __volatile__ ("rdtsc; divl %1"
                     : "=a" (now)
                     : "r" (freq)
-                    : "%eax", "%edx");
+                    : "%edx");
```
JayFoxRox: This change actually causes problems:

clang won't let me use eax as dirty register so I removed it, thinking that it would not touch it until it's used as output.
I've now checked the behaviour and eax will be used for the `freq` input. So the `freq` will be broken and the timer output will be very bizarre values.

A fix for this is to use `"b" (freq)` to force the `freq` into a non-eax/edx register.
Another alternative approach is to ditch this inline assembly altogether and use the `KeTickCount` kernel export which is probably exactly what this function asks for: ms since reboot.

PR #8, opened 2017-04-23
Fix networking code bug / crash introduced by #4

JayFoxRox:
#4 caused random crashes / hangs after about 10 minutes if the network code was active (See https://github.com/xqemu/nxdk/pull/4/files#r112821109 ).
This fixes it by implementing `sys_now` using `KeTickCount`.

---

Additionally this activates the `NtClose` to avoid leaking created semaphores.
The original commented out code had the `NT_SUCCESS` check negated (accidentally?) and always caused `abort`.
I'm assuming this negated condition was the original reason for the code being commented out.

---

(Only tested in xqemu yet, I'll test on hw later)

### DIFF
--- a/lib/net/nforceif/src/sys_arch.c
+++ b/lib/net/nforceif/src/sys_arch.c
@@ -401,12 +401,10 @@ sys_sem_signal(struct sys_sem **s)
 static void
 sys_sem_free_internal(struct sys_sem *sem)
 {
-  // FIXME! This will leak.
-
-  // NTSTATUS status = NtClose((void*)sem->handle);
-  // if (NT_SUCCESS(status)) {
-  //     abort();
-  // }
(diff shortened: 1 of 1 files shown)


lib/net/nforceif/src/sys_arch.c
```diff
@@ -83,16 +83,7 @@ int errno;
...
-
-  /* Read TSC into EDX:EAX, then divide TSC by processor freq / 1000 to
-   * get elapsed milliseconds since reset (roughly).
-   */
-  asm __volatile__ ("rdtsc; divl %1"
-                    : "=a" (now)
-                    : "r" (freq)
-                    : "%edx");
-  return now;
+  return KeTickCount;
```
JayFoxRox: Should have used `XGetTickCount()` (which I wasn't aware off at the time)

PR #9, opened 2017-04-29
Demo/triangle

mborgerson:
I wanted to get familiar with 3d rendering using pbkit, so I boiled down the other pbkit demos to create this one.

This patch adds a new demo which simply draws a triangle using pbkit. It is a highly simplified version of what is in the other pbkit demos. It's relatively small and linear, but still covers enough to show how shaders can be set up, attributes can be set, and geometry can be rendered.

Tested on real Xbox and XQEMU.

### DIFF
--- a/samples/triangle/main.c
+++ b/samples/triangle/main.c
@@ -1,0 +1,244 @@
+/*
+ * This sample provides a very basic demonstration of 3D rendering on the Xbox,
+ * using pbkit. Based on the pbkit demo sources.
+ */
+#include <hal/video.h>
+#include <hal/xbox.h>
+#include <math.h>
+#include <pbkit/pbkit.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <xboxkrnl/xboxkrnl.h>
+#include <xboxrt/debug.h>
+
+/* Restart after a number of milliseconds (0 for never) */
+const int reboot_after = 10*1000;
+
+static uint32_t *alloc_vertices;
+static uint32_t  num_vertices;
+static float     m_viewport[4][4];
+
+typedef struct {
+    float pos[3];
+    float color[3];
+} __attribute__((packed)) ColoredVertex;
+
+static const ColoredVertex verts[] = {
+    //  X     Y     Z       R     G     B
+    {{-1.0, -1.0,  1.0}, { 0.1,  0.1,  0.6}}, /* Background triangle 1 */
+    {{-1.0,  1.0,  1.0}, { 0.0,  0.0,  0.0}},
+    {{ 1.0,  1.0,  1.0}, { 0.0,  0.0,  0.0}},
+    {{-1.0, -1.0,  1.0}, { 0.1,  0.1,  0.6}}, /* Background triangle 2 */
+    {{ 1.0,  1.0,  1.0}, { 0.0,  0.0,  0.0}},
+    {{ 1.0, -1.0,  1.0}, { 0.1,  0.1,  0.6}},
+    {{-1.0, -1.0,  1.0}, { 1.0,  0.0,  0.0}}, /* Foreground triangle */
+    {{ 0.0,  1.0,  1.0}, { 0.0,  1.0,  0.0}},
+    {{ 1.0, -1.0,  1.0}, { 0.0,  0.0,  1.0}},
+};
+
+#define MASK(mask, val) (((val) << (ffs(mask)-1)) & (mask))
+
+static void matrix_viewport(float out[4][4], float x, float y, float width, float height, float z_min, float z_max);
+static void init_shader(void);
+static void set_attrib_pointer(unsigned int index, unsigned int format, unsigned int size, unsigned int stride, const void* data);
+static void draw_arrays(unsigned int mode, int start, int count);
+
+/* Main program function */
+void main(void)
+{
+    uint32_t *p;
+    int       i, status;
+    int       width, height;
+    int       start, last, now;
+    int       fps, frames, frames_total;
+
+    if ((status = pb_init())) {
+        debugPrint("pb_init Error %d\n", status);
+        XSleep(2000);
+        XReboot();
+        return;
+    }
+
+    pb_show_front_screen();
+
+    /* Basic setup */
+    width = pb_back_buffer_width();
+    height = pb_back_buffer_height();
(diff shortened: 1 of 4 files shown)


mborgerson:
image

samples/triangle/main.c
```diff
@@ -0,0 +1,281 @@
...
+        if (num_batches > 1) {
+            n = MAX_VERTICES_PER_BATCH;
+        } else {
+            n = count % MAX_VERTICES_PER_BATCH;
+        }
+
+        uint32_t *p = pb_begin();
+        pb_push1(p, NV097_SET_BEGIN_END, mode); p += 2;
+
+        uint32_t size=(n+251)/252;
```
JayFoxRox: Where do these numbers come from?

samples/triangle/main.c
```diff
@@ -0,0 +1,281 @@
...
+    //  X     Y     Z       R     G     B
+    {{-1.0, -1.0,  1.0}, { 0.1,  0.1,  0.6}},
+    {{-1.0,  1.0,  1.0}, { 0.0,  0.0,  0.0}},
+    {{ 1.0,  1.0,  1.0}, { 0.0,  0.0,  0.0}},
+    {{-1.0, -1.0,  1.0}, { 0.1,  0.1,  0.6}},
+    {{ 1.0,  1.0,  1.0}, { 0.0,  0.0,  0.0}},
+    {{ 1.0, -1.0,  1.0}, { 0.1,  0.1,  0.6}},
+    {{-1.0, -1.0,  1.0}, { 1.0,  0.0,  0.0}},
+    {{ 0.0,  1.0,  1.0}, { 0.0,  1.0,  0.0}},
+    {{ 1.0, -1.0,  1.0}, { 0.0,  0.0,  1.0}},
```
JayFoxRox: Why does this triangle use 9 vertices?

samples/triangle/main.c
```diff
@@ -0,0 +1,281 @@
...
+
+    pb_show_front_screen();
+
+    /* Basic setup */
+    width = pb_back_buffer_width();
+    height = pb_back_buffer_height();
+
+    /* Load constant rendering things (shaders, geometry) */
+    init_shader();
+    alloc_vertices = MmAllocateContiguousMemoryEx(sizeof(verts), 0, 0x3ffb000, 0, 0x404);
```
JayFoxRox: More symbolic constants would be nice or more docs so it's easier to change it in the future once we introduce names for these values

samples/triangle/main.c
mborgerson: 2 triangles for background gradient, 1 for foreground (see picture)

samples/triangle/main.c
```diff
@@ -0,0 +1,281 @@
...
+
+        pb_push1(p,NV097_SET_BEGIN_END, NV097_SET_BEGIN_END_OP_END); p += 2;
+        pb_end(p);
+
+        overflow_estimation+=n;
+        if (overflow_estimation>252*8*16000) //depends on push buffer size
+        {
+            overflow_estimation=0;
+            pb_reset(); //jump back to push buffer head to avoid buffer overflow
+        }
```
JayFoxRox: This construct looks pretty awkward. Can't we just reset more often or after a couple of iterations?
If not: Shouldn't pbkit handle this for us somehow?

samples/triangle/main.c
mborgerson: This function comes direcr from pbkit demos, looks like it just limits the number of verts sent at a time.

samples/triangle/main.c
mborgerson: Not sure which constants you're referring to. Could you be more specific?

samples/triangle/main.c
JayFoxRox: I think you should seperate those then (seperate arrays or seperate draw calls even) or at least add a comment so users know which vertices belong to which triangle

samples/triangle/main.c
mborgerson: Agree, again, this is a duplicate from pbkit demo sources. Could definitely stand to be refactored.

samples/triangle/main.c
JayFoxRox: I think it tries to avoid buffer overruns in the pushbuffer, while still trying to limit the number of jumps

samples/triangle/main.c
```diff
@@ -0,0 +1,281 @@
...
+        pb_push(p++, NV20_TCL_PRIMITIVE_3D_VP_UPLOAD_CONST_X, 16);
+        memcpy(p, m_viewport, 16*4); p+=16;
+
+        pb_end(p);
+        p = pb_begin();
+
+        /* Clear all attributes */
+        pb_push(p++,NV097_SET_VERTEX_DATA_ARRAY_FORMAT,16);
+        for(i = 0; i < 16; i++) {
+            *(p++) = 2;
```
JayFoxRox: Why 2?

samples/triangle/main.c
JayFoxRox: 0x404 and 0x3ffb000; I assume one is the permissions or flags and the other one is a min/max address?

samples/triangle/main.c
mborgerson: Not quite sure. This comes from graphics.c@clear_attribs. NV097_SET_VERTEX_DATA_ARRAY_FORMAT_TYPE_F maybe?

samples/triangle/main.c
mborgerson: Comes from flag.c.

0x3ffb000 is max address to allocate. Maybe @espes knows why this address is important? DMA related?

0x404 looks like:
#define PAGE_READWRITE 0x4
#define PAGE_EXECUTE_READWRITE 0x40

Is this critical?

JayFoxRox: did you mean 0x400 or did you pick the wrong flag?

I'm guessing the max address is because the GPU can't access the full memory (the only Xbox which can access more is Chihiro, which has (almost) full 128MiB access) and this is 63.98MiB in.

samples/triangle/main.c
JayFoxRox: We should figure this out and document it

samples/triangle/main.c
mborgerson: Oops.. I meant PAGE_WRITECOMBINE | PAGE_READWRITE.

PR #10, opened 2017-05-12
Add Support for compiling on Windows (using MSYS2)

LukeUsher:
This allows nxdk to be used on Windows using the MSYS2 shell.

Known Issues:
 - Currently LLVM needs to be installed separately as the MSYS2 package is still on LLVM 3.9, as soon as LLVM 4.0 ships in MSYS2, this will no longer be required (this is why it is not mentioned in the readme)
 - extract-xiso does not compile under MSYS2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 *.d
 _*.cpp
 _*.h
+!tools/cg/win/cgc.exe
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,12 @@ CC           = /usr/local/opt/llvm/bin/clang
 CXX          = /usr/local/opt/llvm/bin/clang++
 CGC          = $(NXDK_DIR)/tools/cg/mac/cgc
 endif
+ifneq (,$(findstring MSYS_NT,$(UNAME_S)))
+LD           = lld-link
+CC           = clang
+CXX          = clang++
+CGC          = $(NXDK_DIR)/tools/cg/win/cgc
+endif
 
 TARGET       = $(OUTPUT_DIR)/default.xbe
 CXBE         = $(NXDK_DIR)/tools/cxbe/cxbe
--- a/tools/vp20compiler/Makefile
+++ b/tools/vp20compiler/Makefile
@@ -8,7 +8,7 @@ SRCS = nvvertparse.c \
 
 OBJS = $(SRCS:.c=.o)
 
-CFLAGS = -std=c99
+CFLAGS = -std=gnu99
 
 $(MAIN): $(OBJS)
 	$(CC) -o '$@' $(OBJS)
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@
 *.d
 _*.cpp
 _*.h
+!tools/cg/win/cgc.exe
```
JayFoxRox: newline at end of file please

Makefile
```diff
@@ -23,6 +23,12 @@ CC           = /usr/local/opt/llvm/bin/clang
 CXX          = /usr/local/opt/llvm/bin/clang++
 CGC          = $(NXDK_DIR)/tools/cg/mac/cgc
 endif
+ifneq (,$(findstring MSYS_NT,$(UNAME_S)))
```
JayFoxRox: why is this `ifneq`? also is there a better way to find the platform maybe?

tools/vp20compiler/Makefile
```diff
@@ -8,7 +8,7 @@ SRCS = nvvertparse.c \
 
 OBJS = $(SRCS:.c=.o)
 
-CFLAGS = -std=c99
+CFLAGS = -std=gnu99
```
JayFoxRox: motivation for change not stated in PR description.

Makefile
```diff
@@ -23,6 +23,12 @@ CC           = /usr/local/opt/llvm/bin/clang
...
+LD           = lld-link
+CC           = clang
+CXX          = clang++
+CGC          = $(NXDK_DIR)/tools/cg/win/cgc
```
JayFoxRox: just wondering: is this the 32 or 64 bit version?

tools/vp20compiler/Makefile
JayFoxRox: just FYI: I split this into a seperate commit in my WIP branches "vp20compiler: Use gnu99 for strdup support"

LukeUsher: I included the 64-bit version, but I just had a thought: Perhaps I should replace it with the 32-bit version for maximum compatibility. What do you think?

Makefile
LukeUsher: I included the 64-bit version, but I just had a thought: Perhaps I should replace it with the 32-bit version for maximum compatibility. What do you think?

JayFoxRox: I think we should include both and select on the running environment (this would be a seperate change / PR).
Ideally we'd have our own `cgc` like tool in open-source (but we don't have developer capacity for this).

However, 32 bit is almost dead and we don't know how long Microsoft will support 32 bit still.
So 64 bit only is fine. People can create an issue if they run into problems.
(No need to waste time on changing this now)

PR #11, opened 2017-06-21
Makefile fixes

JayFoxRox:
* ~~I had issues where the deps were outdated or my project folders were full of *.d files. This branch cleans the dep on `make clean`.~~
* Code should have a way to identify that it's being build using NXDK. This adds a NXDK preprocessor macro, we should also consider defining XBOX or something in the future.
* As the makefile is included, we need a way to add more CFLAGS (to add more include directories, preprocessor defines etc.). This branch renames our CFLAGS to NXDK_CFLAGS and appends existing CFLAGS to the end.

The integration of the last change is not the best. But I hate makefiles and can't be bothered to figure out how to do this properly - this get's the job done (I think).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,14 +30,16 @@ VP20COMPILER = $(NXDK_DIR)/tools/vp20compiler/vp20compiler
 FP20COMPILER = $(NXDK_DIR)/tools/fp20compiler/fp20compiler
 EXTRACT_XISO = $(NXDK_DIR)/tools/extract-xiso/extract-xiso
 TOOLS        = cxbe vp20compiler fp20compiler extract-xiso
-CFLAGS       = -target i386-pc-win32 -march=pentium3 \
+NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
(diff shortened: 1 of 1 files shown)


JayFoxRox:
The dep clean was slightly annoying: When doing `make clean` it would re-generate all deps first. So in order to solve this mess I've removed said commit.

Please review / merge soon. I have a handful of PRs (for nxdk-rdt) depending on this.

Makefile
```diff
@@ -30,15 +30,17 @@ VP20COMPILER = $(NXDK_DIR)/tools/vp20compiler/vp20compiler
 FP20COMPILER = $(NXDK_DIR)/tools/fp20compiler/fp20compiler
 EXTRACT_XISO = $(NXDK_DIR)/tools/extract-xiso/extract-xiso
 TOOLS        = cxbe vp20compiler fp20compiler extract-xiso
-CFLAGS       = -target i386-pc-win32 -march=pentium3 \
+NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
```
mborgerson: Why not just do `CFLAGS += `

PR #16, opened 2017-07-08
Add support for vertex state programs (VSP)

JayFoxRox:
This is the first step towards better vertex shader support.
It will allow for testing and researching vertex shaders in the future.

Some changes are necessary / wise:

* pbkit internal functions are exposed so it's possible to wait for the GPU to be idle (necessary for reading PGRAPH RDI to verify VSP results)
* The shader config for the vertex programs in vp20compiler was changed to represent the NV2A
* Error output has been added so shaders are debuggable better
* Removal of `assert(false);` for GL_VERTEX_STATE_PROGRAM_NV and addition of constant writes
* Addition of nv2a registers for launch programs / VSPs in pbkit

Issues to be created after merge (as these were not too easy to test until now and still need a bit of work / research / discussion):

* Confirm R12 still works correctly (is it writeable in our implementation?)
* Confirm if index for constant writes is correct (also check if relative indexing through A0.x works; with stand-alone indexed writes and if two things happen at once: indexed read + indexed write)
* Should constant indexing work from 0 to 192 or from -96 to +95? (What happens in the different GPU modes?)


lib/pbkit/nv_objects.h
```diff
@@ -1219,8 +1219,11 @@ NV20_TCL_PRIMITIVE_3D      [0x1d78/4] = 0x00000001 | UNKNOWN = 00000001
 #define NV20_TCL_PRIMITIVE_3D_SHADER_TYPE				0x00001e94
 #define 	SHADER_TYPE_INTERNAL	4
 #define 	SHADER_TYPE_EXTERNAL	6
+#define 	SHADER_TYPE_INTERNAL_LIMITED	0
+#define 	SHADER_TYPE_EXTERNAL_LIMITED	2
 #define NV20_TCL_PRIMITIVE_3D_SHADER_SUBTYPE				0x00001e98
 #define		SHADER_SUBTYPE_REGULAR	0 						//not allowed to write into constants: faster
+#define		SHADER_SUBTYPE_WRITE		1 	  		  	//can write into constants: slower?!
```
JayFoxRox: change tabs to spaces?
(existing code used tabs, but it looks like a giant mess)

PR #23, opened 2017-07-16
Fix mixup in video-mode list and use kernel Av-constant names

### LINKED ISSUE #22: Incorrect AV pack constants lead to crash on PAL-Xbox
I found this while investigating why all the nxdk-samples make my Xbox (PAL, 1.0 I think since it used to have that GPU-fan) lock up.

I fired up my Debug-BIOS, which showed that somewhere in the code, MmAllocateContiguousMemoryEx was called with a size-parameter of zero, which triggers an assertion in the kernel. I managed to track down this allocation to [line 3489 of pbkit.c](https://github.com/xqemu/nxdk/blob/master/lib/pbkit/pbkit.c#L3489). Further investigation showed that no videomode was set, and neither pb_init() nor XboxCRT() handle that case. I dug in the video-code and found out that no mode was set because all were rejected based on my AV pack - the constants go up to 5, my Xbox reports a 6. Removing that check (not a clean solution) resolved that for me, and the httpd-sample stopped crashing and worked nicely (didn't test others yet).

So the AV-pack constants that nxdk has seem to be incomplete. I managed to get my hands on a list of constants the kernel uses for them, which is the following:
```
#define AV_PACK_NONE                      0x00000000
#define AV_PACK_STANDARD                  0x00000001
#define AV_PACK_RFU                       0x00000002
#define AV_PACK_SCART                     0x00000003
#define AV_PACK_HDTV                      0x00000004
#define AV_PACK_VGA                       0x00000005
#define AV_PACK_SVIDEO                    0x00000006
#define AV_PACK_MAX                       0x00000007
#define AV_PACK_MASK                      0x000000FF
```
These differ a bit from the ones nxdk uses. Interestingly, these constants claim that my AV pack is SVIDEO, but it's a standard composite-one.

To resolve this, I'd suggest that pb_init() gets a check for a set video-mode. However, I'm not sure about what exactly to do about the constant-issue, which is the core-problem.

JayFoxRox:
Fixes #22  (also describes the solution). Summary of changes:

* The mapping of the adapter constants to the video-modes has been corrected (was responsible for the crash)
* The order of the video modes is now in the order of constants
* I chose to include the AV constants listed in #22 as they are commonly found on the internet anyway. I don't think anybody will mind.


thrimbor:
Works fine for me now, thanks!

On a side-note, I found out what made my Xbox think it had an S-Video cable attached: I have an original cable, but I also have a TOSLINK-adapter to get digital audio from the xbox (looks like [this](https://abload.de/img/img_20170717_003116kks1a.jpg)) which you plug between the original cable and the xbox. Without the adapter, it works both with and without the patch. With the adapter, I need the patch to make nxdk work.

JayFoxRox:
Tested with all my available adapters with Xbox configured to NTSC-M with 480p and 720p HDTV enabled. Respective "Value" is returned by `AvSendTVEncoderOption` with Option=6:

---

Value is: 0x004A0101
Cable type: 0x01 (Standard AV Cable (Composite))
Video standard: 0x01 (NTSC-M)
60 Hz

---

Value is: 0x004A0103
Cable type: 0x03 (Advanced SCART Cable)
Video standard: 0x01 (NTSC-M)
60 Hz

(Third-party "Brooklyn" cable with composite out, only tested composite)

---

Value is: 0x004A0104
Cable type: 0x04 (High Definition AV Pack (Component))
Video standard: 0x01 (NTSC-M)
60 Hz

---

Untested: anything PAL (might do that some other time), VGA, S-Video, RF adapter

PR #26, opened 2017-07-25
Replaced xboxkrnl.exe.def and libxboxkrnl.lib with more complete ones

### LINKED ISSUE #14: Update kernel imports
The kernel imports are not complete. This should be changed.
Unfortunately it's also not possible to easily generate a new libxboxkrnl.lib (I believe I ran into some issues with the existing one a while back, but I don't remember the specifics currently)

Regarding the headers / signatures and datatypes, @PatrickvL said:

> Almost all kernel signatures have been added to the OpenXDK copy in Cxbx-Reloaded, here : https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/tree/master/import/OpenXDK/include/xboxkrnl
> 
> I don't have the time to contribute to either OpenXDK nor NXDK, but you people shouldn't let a resource like this go to waste! ;)

However, I looked into this after naively moving the code to nxdk:
* There is some incompatibility as to where certain types are declared (but this is easily fixed).
* There is also an issue with the datatypes (structs) only being correctly put in memory for MSVC. clang / gcc use a different packing, so explicit padding fields and explicit packing has to be used. Even mingw does not pack some structures correctly I believe.
* Some signatures and data types are very obviously copied from copyrighted code. There are legal means to figure out these signatures, so we should use them, instead of just copying them around. While I think it's perfectly fine to copy the function signatures, at least whitespaces and comments should be removed / changed.

Eitherway, we should update our signatures and datatypes.
Contributions welcome.

thrimbor:
This replaces the xboxkrnl.exe.def with the one I've been creating for my own XDK-project. This xboxkrnl.exe.def is not derived from the one in OpenXDK or nxdk, which is why I was able to license it under the CC0.

This new file includes, afaik, all known functions with the proper stdcall/fastcall-annotations. The libxboxkrnl.lib was built by using the command as listed in make_import_lib.bat. I tested the new library by building and running all samples on both xqemu and my Xbox.

This, together with the already merged xboxkrnl.h-update, should fix #14.

I would've liked to remove the need to include a binary file all-together, but that doesn't seem to be possible at the moment. I have been able to use mingw for that in the past (i686-w64-mingw32-dlltool -m i386 -D xboxkrnl.exe -l libxboxkrnl.lib -d xboxkrnl.exe.def), but it seems that a mingw-update broke that a while back. LLVMs lld gained the ability to generate import-libs from definitions, but that's only available in SVN right now and has problems with "NONAME".


espes:
thanks!

PR #28, opened 2017-12-29
Added missing definition (@ExInterlockedAddLargeStatistic@8)

thrimbor:
This fixes a missing function in the new xboxkrnl definition file. I regenerated the import library and used it to run the hello-sample on my Xbox to make sure it didn't break.

The missing function was reported by @mborgerson


espes:
thanks!

PR #29, opened 2018-01-02
Added a couple of useful string handling functions

Luca1991:
Hello nxdk people!!

I'm currently writing a program to test xbox kernel apis (to test CxBx-R HLE kernel implementation), and while writing the configuration file parser, I realized that a couple of useful string functions were missing from nxdk.

So, this PR add these functions. Every of them are adapted from FOSS projects and contains a reference in comment.

I hope that everything is ok, I'm waiting for your verdict :P

Happy New Year,
Luca

### DIFF
--- a/lib/xboxrt/stdlib.c
+++ b/lib/xboxrt/stdlib.c
@@ -1,5 +1,6 @@
 #include "stdlib.h"
 #include "string.h"
+#include "ctype.h"
 #include <xboxkrnl/xboxkrnl.h>
 
 static const char digits[] = "0123456789abcdef";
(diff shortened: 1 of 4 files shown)


mborgerson:
Always good to see improvements like these!

PS: While you're at it... we could use an improved `printf`. 😉

lib/xboxrt/stdlib.c
```diff
@@ -179,3 +180,85 @@ void *realloc(void *ptr, size_t size)
     }
     return new;
 }
+
+// This is adapted from libc
+// Ref: https://github.com/gcc-mirror/gcc/blob/master/libiberty/strtol.c
```
JayFoxRox: wouldn't the current proposal violate their license?

espes:
thanks! would be good to make sure these all have compatible licenses in a follow-up PR.

PR #30, opened 2018-01-02
Add /debug linker flag for debug builds

mborgerson:
This patch adds the `/debug` flag to the linker when performing a debug build. This is necessary to instruct the linker to keep the debug symbol sections in the final binary.


espes:
thanks! feel free to merge yourself in future...

PR #44, opened 2018-08-10
Add mesh sample

mborgerson:
This PR adds a new sample which draws real 3D geometry, with textures and lights. Based on my triangle sample, and in part on the 0ldskoo1 sample (which is in-turn based on pbkit samples). The code here is a little easier to navigate over the other 3d samples. Simple text mesh of 'nxdk' and planar mapping of the texture.

screenshot

The code needs a little cleanup, specifically replacing some of the magic numbers with their symbolic counterparts, but I think it's in good addition as-is. Cleanup can happen in follow-up PR.


dracc:
I've built and tested both in xqemu and on real hardware.
Works just fine, apart from a few flickers on my box.

Screenshot

JayFoxRox:
> Cleanup can happen in follow-up PR.

I mostly agree. However, I'd like to see the following changes before a merge:

1. Document where the texture.h came from, or how to create one.
2. Document where the verts.h came from, or how to create one (rename to mesh.h?).
3. Remove reboot after 600 frames, because it feels like a crash.

The reboot feels too much like a crash. The 0ldskoo1 already confuses people by creating a solid rectangle ontop of the 3D visuals. If they run the mesh sample next they'll assume nxdk is very broken.

Not being able to generate new textures / meshes is a big problem, too. Simply because people can't tinker with it then. Especially beginners will likely start with artwork-swaps.

> The code here is a little easier to navigate over the other 3d samples.

I still feel it's much too complicated to be honest.
I feel like this should be a more simple example using the fixed-function pipeline.
We should also hide all the `pb_push` behind a slightly higher level API (but below something like GL).
Anyhow, all of this should be done in a follow-up PR.
For now, having another functional sample is a lot better than not having it.

---

I ran this sample on my Xbox and it worked fine when starting it from evox. It did crash when trying to run it using XBDM on a debug BFM kernel (not sure which one I used; it's probably also a more generic nxdk issue).

Anyhow, LGTM, but I'd like to see the 3 things listed above documented / changed.

JayFoxRox:
Merged. Please address those things I've mentioned in my review as a follow-up, or create an issue @mborgerson

mborgerson:
Issue #45

Issue #35, opened 2018-03-28
Fix networking packet size issue


mborgerson:
Seems like packets of size greater than 128 bytes cause issues, making the network stack *almost* useless. It's probably not too complicated, just needs investigation.

JayFoxRox:
We have a couple of network users by now, and nobody seems to have problems with this. I don't think this problem ever existed (I assume it was bugs in how lwip was used).

If this problem really exists, create a new issue, or leave a message so we can re-open this discussion.

Closed.

PR #48, opened 2018-09-10
libc and xboxkrnl fixes

JayFoxRox:
Changes necessary for xbe-loader, nxdk-rdt and other projects:

* This *should* fix our `free()` function which should make all nxdk programs more stable. Our existing `free` does not actually free memory. Could affect #35.
* It also fixes `strlen()` which broke many things for me. It used to return 1 for `strlen("")`. I've fixed and re-enabled `strdup` as a result.
* I also decided to enable LWIP DNS support, which required `atoi`.
* I also added a `bool` type to our stdbool.
* The mem* functions are now handled using the kernel Rtl functions. This should improve performance.
* I've addressed all sort of bugs in the xboxkrnl which prevented people from writing new drivers (this includes changing calling conventions and making `BOOL` a smaller type - most of our xboxkrnl structs were the wrong size).

See commit titles for more infos.

Ideally, someone should test this - I assume most of it works, but I'm not sure either:
I added whatever I needed to make errors disappear. So I just assumed my code was working (even if it might only be a workaround, eventually leading to other errors).

### DIFF
--- a/lib/net/nforceif/include/lwipopts.h
+++ b/lib/net/nforceif/include/lwipopts.h
@@ -315,7 +315,7 @@ extern int errno;
  * LWIP_DNS==1: Turn on DNS module. UDP must be available for DNS
  * transport.
  */
-#define LWIP_DNS                        0
+#define LWIP_DNS                        1
 
 /*
    ---------------------------------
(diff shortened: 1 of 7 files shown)


mborgerson:
Thanks for taking the time to fix these issues!

PR #50, opened 2018-09-12
build: Fix mingw64 build

dracc:
It seems as if `g++` and `Clang++` has some discrepancies when it comes to what headers are to be included and what overloads to create for some functions.
This hinders building nxdk under MinGW64 currently.

Submodule extract-xiso now tracks XboxDev/extract-xiso instead of @mborgerson's repo, as it's up to date and successfully builds in MinGW64.

Changes tested in MinGW64, WSL (Debian) and barebone Debian testing.


.gitmodules
```diff
@@ -1,10 +1,10 @@
 [submodule "tools/extract-xiso"]
 	path = tools/extract-xiso
-	url = https://github.com/mborgerson/extract-xiso.git
+	url = https://github.com/xboxdev/extract-xiso.git
 [submodule "lib/net/lwip"]
 	path = lib/net/lwip
 	url = git://git.savannah.nongnu.org/lwip.git
```
JayFoxRox: Please put the lwip update in a separate commit, possibly even a separate PR.

If lwip breaks, we'd have to revert *all* the changes, including the Windows fix (which might also just be a workaround?).

Same goes for other stuff like changing the upstream for the other submodules, or fixing code in cxbe.
However, they are probably smaller changes which are not as critical

Only group those changes which are absolutely necessary to reach your goal (PR). And each unrelated step towards that goal should also be separate (commit).

README.md
```diff
@@ -34,9 +34,10 @@ Then install lld from an [LLVM nightly package](http://apt.llvm.org/)
 #### Windows
 Install [MSYS2](http://www.msys2.org/)
 
-In MSYS2 Shell:
+In mingw64 Shell:
 
-    pacman -S make git gcc bison flex llvm clang
+    pacman -S make git mingw-w64-x86_64-gcc bison flex mingw-w64-x86_64-llvm \
+    mingw-w64-x86_64-clang mingw-w64-x86_64-lld
```
JayFoxRox: nit: This feels a bit like a workaround, rather than a fix.

Did you consult the MSYS people about this and asked why this would be necessary?
What did we conclude when we discussed this on Discord when Chris had the problems you are addressing?

The reason I'm asking, is that this now restricts us to 64 bit machines (although going to 32 bit is still easy). Whereas the MSYS line should have worked on 32 and 64 bit (and also worked on other pacman based environments).

*You might also want to move bison and flex to the left, so the unreadable part starts later.*

Makefile
```diff
@@ -29,6 +29,11 @@ CC           = clang
 CXX          = clang++
 CGC          = $(NXDK_DIR)/tools/cg/win/cgc
 endif
+ifneq (,$(findstring MINGW,$(UNAME_S)))
+LD           = lld-link
+CC           = clang
+CXX          = clang++
```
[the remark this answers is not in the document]
dracc: I'll add CGC.
We can't use the same case as MinGW uses `lld-link` and MSYS uses `ld.lld`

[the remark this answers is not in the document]
dracc: Seeing as I can't seem to get a straight answer other than "Don't use MSYS tools as noone seems to do that", I'll combine these as you suggested.

JayFoxRox:
Waiting for feedback before merge:

- I'm surprised there are no README changes, when packages for MSYS and MinGW are typically named differently
- Do you want to make the commit naming style to be more uniform?

I want this merged soon. So please no larger changes in this one.
For a possible follow-up:

- If MSYS exposes the mingw binaries, we could also use them directly? (like /usr/bin/mingw-blabla/lld)
- MSYS does not work, right? Maybe we should [just throw an error if MSYS is used as platform](https://stackoverflow.com/questions/1864385/how-to-force-an-error-in-a-gnumake-file) ("Please use a MinGW64 shell"?).

README.md
```diff
@@ -34,9 +34,10 @@ Then install lld from an [LLVM nightly package](http://apt.llvm.org/)
 #### Windows
 Install [MSYS2](http://www.msys2.org/)
 
-In MSYS2 Shell:
+In MinGW64 Shell:
```
mborgerson: We also give this guidance on xqemu.com...and I made it very bold because I spent a couple minutes scratching my head as to why I couldn't build anything. So, I think we need to be more clear here too because people (like me) might miss it. Here's what I wrote on xqemu.com:

> After installing MSYS2, you'll need to open MSYS2 MinGW 64-bit to perform building. Otherwise, you may see build failures for cc.exe.

So maybe something along these lines. I'll let you put it in your own words.

tools/cxbe/Xbe.cpp
```diff
@@ -1071,7 +1072,7 @@ void Xbe::DumpInformation(FILE *x_file)
     fprintf(x_file, "Size of Headers                  : 0x%.08X\n", m_Header.dwSizeofHeaders);
     fprintf(x_file, "Size of Image                    : 0x%.08X\n", m_Header.dwSizeofImage);
     fprintf(x_file, "Size of Image Header             : 0x%.08X\n", m_Header.dwSizeofImageHeader);
-    fprintf(x_file, "TimeDate Stamp                   : 0x%.08X (%s)\n", m_Header.dwTimeDate, BetterTime(ctime((const long*)&m_Header.dwTimeDate)));
+    fprintf(x_file, "TimeDate Stamp                   : 0x%.08X (%s)\n", m_Header.dwTimeDate, BetterTime(ctime((const time_t*)&m_Header.dwTimeDate)));
```
mborgerson: Hmm.. Is this right? Can we be sure that type `time_t` is a 32-bit value?

dracc: Seeing as `time_t` is undefined in the C standard according to [cppreference.com](https://en.cppreference.com/w/c/chrono/time_t), you're probably right that we can't. `clang` would not compile with just a regular `long int*` though.
Suggestions welcome!

dracc: After a quick discussion on Discord, @mborgerson suggested that a conversion of the `long int` was done before sending a pointer to `ctime()`, instead of converting the pointer.

mborgerson:
Thanks!

PR #53, opened 2018-09-25
Create a basic assert.h

dracc:


lib/xboxrt/assert.h
```diff
@@ -0,0 +1,11 @@
+#ifndef __ASSERT_H
+#define __ASSERT_H
+
+#ifndef NDEBUG
+#include "debug.h"
+#define assert(x) if (!(x)) { debugPrint("\nAssertion failed! '%s' in %s:%d\n\n", #x, __FILE__, __LINE__); }
+#else
+#define assert(x)
```
JayFoxRox: This is not ideal, because it will break some C code such as:

```c
if (false)
  assert(true);
exit();
crash();
```

lib/xboxrt/assert.h
```diff
@@ -0,0 +1,11 @@
+#ifndef __ASSERT_H
+#define __ASSERT_H
+
+#ifndef NDEBUG
+#include "debug.h"
+#define assert(x) if (!(x)) { debugPrint("\nAssertion failed! '%s' in %s:%d\n\n", #x, __FILE__, __LINE__); }
```
JayFoxRox: nit: I don't like our `debugPrint` solution because we should not expect a framebuffer, serial connection, ... . I think we should force users to provide their own `debugPrint` handlers (and the default should be a no-operation function)

I'm fine with keeping this for now though, it's not your fault and an overall problem in nxdk.

dracc:
Closed in favor for @thrimbor's much more elegant code. :+1:

PR #54, opened 2018-09-29
Add libc runtime functions

### LINKED ISSUE #32: Nxdk fails to link __chkstk when allocating a large number of local variables on the stack
When trying to create a large number of local variables on the stack in a function, nxdk will fail to link because it is trying to use the __chkstk symbol.

lld-link: error: /mnt/e/Windows/Programming/nxdk/my_progs/xbox_kernel_test_suite/rtl_tests.obj: undefined symbol: __chkstk
/mnt/e/Windows/Programming/nxdk/my_progs/xbox_kernel_test_suite/../../Makefile:94: recipe for target 'main.exe' failed
make: *** [main.exe] Error 1

Code that causes this:

void foo() {
    char buffer[2000];
}

thrimbor:
This adds functions required at runtime by compiler generated function calls. Most of the 64-bit math code was taken from LLVM's compiler_rt library (required license files are included) and modified to fit the required calling convention.

I originally intended to use LLVM's __chkstk implementation too, but their implementation expects the OS to grow the stack when the guard page is hit during probing. However, on the Xbox, it seems that stacks are statically allocated, and while there is a guard page, it is only used to catch stack overflows - dynamic growth is not supported. So I wrote my own __chkstk implementation, which uses a C helper function to check whether the requested amount fits into the current thread's stack and prints an error if it doesn't.

I quickly tested it by running a sample program which creates a large array on the stack, and it seemed to work correctly.

Should fix #32.

### DIFF
--- a/lib/xlibc-rt/_alldiv.s
+++ b/lib/xlibc-rt/_alldiv.s
@@ -1,0 +1,158 @@
+// This file is dual licensed under the MIT and the University of Illinois Open
+// Source Licenses. See LICENSE.TXT for details.
+
+// di_int __divdi3(di_int a, di_int b);
+
+// result = a / b.
+// both inputs and the output are 64-bit signed integers.
+// This will do whatever the underlying hardware is set to do on division by zero.
+// No other exceptions are generated, as the divide cannot overflow.
+//
+// This is targeted at 32-bit x86 *only*, as this can be done directly in hardware
+// on x86_64.  The performance goal is ~40 cycles per divide, which is faster than
+// currently possible via simulation of integer divides on the x87 unit.
+//
+// Stephen Canon, December 2008
+
+// Modified for stdcall calling convention for use in nxdk
+
+.text
+.balign 4
+.globl __alldiv
+__alldiv:
+
+/* This is currently implemented by wrapping the unsigned divide up in an absolute
+   value, then restoring the correct sign at the end of the computation.  This could
+   certainly be improved upon. */
+
+	pushl		%esi
+	movl	 20(%esp),			%edx	// high word of b
+	movl	 16(%esp),			%eax	// low word of b
+	movl		%edx,			%ecx
+	sarl		$31,			%ecx	// (b < 0) ? -1 : 0
+	xorl		%ecx,			%eax
+	xorl		%ecx,			%edx	// EDX:EAX = (b < 0) ? not(b) : b
+	subl		%ecx,			%eax
+	sbbl		%ecx,			%edx	// EDX:EAX = abs(b)
+	movl		%edx,		 20(%esp)
+	movl		%eax,		 16(%esp)	// store abs(b) back to stack
+	movl		%ecx,			%esi	// set aside sign of b
+
+	movl	 12(%esp),			%edx	// high word of b
+	movl	  8(%esp),			%eax	// low word of b
+	movl		%edx,			%ecx
+	sarl		$31,			%ecx	// (a < 0) ? -1 : 0
+	xorl		%ecx,			%eax
+	xorl		%ecx,			%edx	// EDX:EAX = (a < 0) ? not(a) : a
+	subl		%ecx,			%eax
+	sbbl		%ecx,			%edx	// EDX:EAX = abs(a)
+	movl		%edx,		 12(%esp)
+	movl		%eax,		  8(%esp)	// store abs(a) back to stack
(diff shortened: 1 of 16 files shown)


lib/xlibc-rt/check_stack.c
```diff
@@ -0,0 +1,25 @@
+#include <debug.h>
+#include "xboxkrnl/xboxkrnl.h"
+
+#ifdef DEBUG_CONSOLE
+    #define _print DbgPrint
+#else
+    #define _print debugPrint
+#endif
+
+void _xlibc_check_stack (unsigned int eax, unsigned int esp)
```
JayFoxRox: Given that this is called from the assembly, you might want to add an explicit calling convention here

lib/xlibc-rt/check_stack.c
```diff
@@ -0,0 +1,25 @@
...
+{
+    // eax is the amount of requested memory
+    // esp is the stack before the call to __chkstk
```
JayFoxRox: I'd suggest renaming the function parameters, especially because `eax=$eax`, but `esp=$ecx`.
Also this function itself has little to do with x86 anymore: it really only checks for stack bounds. So I'd rename `esp` to `stack-pointer`; and rename `eax` to `requested_size`.

*(Under the assumption this function was not written to be more similar to an existing implementation elsewhere; if so, I'd suggest to link to it in a comment)*

lib/xlibc-rt/check_stack.c
```diff
@@ -0,0 +1,25 @@
...
+    PKTHREAD current_thread = KeGetCurrentThread();
+
+    if (esp - eax < (DWORD)current_thread->StackLimit || eax >= esp)
```
JayFoxRox: nit: Might want to check `eax >= esp` first, before using them in the subtraction (which might give bogus results we don't care about)

lib/xlibc-rt/chkstk.s
```diff
@@ -0,0 +1,29 @@
...
+    pushl %eax
+    call __xlibc_check_stack
+    addl $8, %esp
+
+    popl %eax
+    popl %ecx
+    subl %eax, %ecx     // ecx is the new stack pointer
+    leal 4(%esp), %eax  // eax is a pointer to the return address variable
+    movl %ecx, %esp     // load the new stack address
+    movl -4(%eax), %ecx // restore ecx
```
JayFoxRox: What's the calling convention of this function? Do we really have to restore ecx?

(Might also be worth noting the context on entry in a comment for the function: requested size is in eax etc.)

lib/xlibc-rt/_alldiv.s
```diff
@@ -0,0 +1,158 @@
...
+// both inputs and the output are 64-bit signed integers.
+// This will do whatever the underlying hardware is set to do on division by zero.
+// No other exceptions are generated, as the divide cannot overflow.
+//
+// This is targeted at 32-bit x86 *only*, as this can be done directly in hardware
+// on x86_64.  The performance goal is ~40 cycles per divide, which is faster than
+// currently possible via simulation of integer divides on the x87 unit.
+//
+// Stephen Canon, December 2008
+// Stefan Schmidt, October 2018
```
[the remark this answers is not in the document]
thrimbor: I didn't write the code, I only modified it to fit the required calling convention (cdecl->stdcall) and adjusted the name (divdi3->_alldiv).
My intention was to make sure that people see that I've modified the code, so the original author doesn't get blamed in case somethings broken.
I see it's misleading and I'll remove it.

lib/xlibc-rt/check_stack.c
thrimbor: Will do.

lib/xlibc-rt/check_stack.c
thrimbor: I agree, that's a better way to name things.

lib/xlibc-rt/chkstk.s
thrimbor: chkstk doesn't really comply to any standardized calling convention (for example, %esp gets modified). As far as I understood it, a call to chkstk is supposed to have the same effect as "subl %eax, %esp", so we're not allowed to modify %eax or trash other registers (actually, I forgot to also save and restore %edx, which is specified as caller-saved in cdecl, when calling _xlibc_check_stack).
I'll add a comment to the function to clarify things a bit.

PR #58, opened 2018-10-08
samples: Do not restart

### LINKED ISSUE #45: Remove reboot on triangle and mesh samples
I added it for convenience (so I wouldn't have to reach over and restart my Xbox), but most people seem to perceive it as a crash since they haven't read the source code--so it should probably just be removed.

dracc:
Closes #45.

Small fix to make sure the samples doesn't automatically restart.

### DIFF
--- a/samples/mesh/main.c
+++ b/samples/mesh/main.c
@@ -15,9 +15,6 @@
 #include <xboxrt/debug.h>
 #include "math3d.h"
 
-/* Restart after a number of milliseconds (0 for never) */
-const int reboot_after = 10*1000;
-
 static uint32_t *alloc_vertices;
 static uint32_t *alloc_indices;
 static uint32_t  num_vertices;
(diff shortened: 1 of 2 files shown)


samples/mesh/main.c
```diff
@@ -265,17 +261,7 @@ void main(void)
...
             last = now;
         }
-
-        /* Exit the demo after timeout */
-        if ((reboot_after > 0) && ((now-start) > reboot_after)) {
-            break;
-        }
     }
-
-    MmFreeContiguousMemory(alloc_vertices);
```
mborgerson: May want to keep these even if they are unreachable so people know what cleanup would look like.

JayFoxRox: Agree. keep this and `pb_kill`.
Then add a comment to tell people that this is not reachable before the cleanup code.

JayFoxRox:
Thanks!

PR #60, opened 2018-10-12
libc additions

thrimbor:
These are the first bits and pieces of my libc work. Additions mostly, without much replacing going on yet.

a6e0ec8cf355af94c8a1d6d9056dbca12a05ca26 adds the prerequisites for the later additions (include dir and source paths).
f3c9cbab073df536e18de2df525fd98db32524ca adds _restrict.h, which is meant as a libc-internal header to allow using the "restrict"-specifier while staying C++ compatible.
d89f109400cd4fe07d9290e8a6de6ac82d22b051 adds assert(). The standard actually says to call abort() (which would exit the program) which doesn't make much sense here, so the Xbox is halted instead. When compiled for a debug console (this isn't integrated with the makefile yet), RtlAssert is used instead of printing to the screen, and the Xbox is not halted in order to support tools like xbWatson, which allows the user to resume execution. Both codepaths were tested.
0efd6b70a73becf198561ca6e6bb5c99cdd9c8c6 adds stdbool.h, replacing the pre-existing incomplete one.
2640201c3f6bf0a6ca07e0bbf424834a5e03e4da adds stdnoreturn.h, not the most useful header, but the standard demands it.
95a84ddf6f6e2bc3826c376240d0ae7b998dbfae adds setjmp.h and implements setjmp and longjmp.
652cf683de0dff62f0ab5502b52319d0802e0f3b adds stdarg.h which provides va_start, va_end, va_arg and va_copy (we've actually been using the host's stdarg.h till now).

@JayFoxRox I tested setjmp/longjmp with your [code sample](https://github.com/JayFoxRox/nxdk/blob/8da92893d1cc94f43828bce9b6b9114b67bd6dd1/samples/setjmp/main.c) and it seemed to work fine, but you might want to double-check.

I previously wasn't sure if setjmp/longjmp should also touch FPU registers because the standard only says "it does not include the state of the floating-point status flags[...]", but looking at other libc's it seems normal to not store FPU registers on x86.

My assert-macro contains a loop that might seem pointless, but it's there to prevent users from accidentally doing things like `assert(0==1)
 else {
     print("This is broken\n");
 }`

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -21,6 +21,8 @@ USB_SRCS := \
 
 SRCS += \
 	$(wildcard $(NXDK_DIR)/lib/xboxrt/*.c) \
+	$(shell find $(NXDK_DIR)/lib/xlibc/src/ -name "*.c") \
+	$(shell find $(NXDK_DIR)/lib/xlibc/src/ -name "*.s") \
 	$(wildcard $(NXDK_DIR)/lib/xlibc-rt/*.c) \
 	$(wildcard $(NXDK_DIR)/lib/xlibc-rt/*.s) \
 	$(wildcard $(NXDK_DIR)/lib/hal/*.c) \
--- a/lib/xlibc/include/assert.h
+++ b/lib/xlibc/include/assert.h
@@ -1,0 +1,41 @@
+#ifndef _XLIBC_ASSERT_H
+#define _XLIBC_ASSERT_H
+
+/* based on _PDCLIB_symbol2value from PDCLIB */
+#ifndef _XLIBC_symbol2value
+#define _XLIBC_symbol2value(x) #x
+#endif
+
+/* based on _PDCLIB_symbol2string from PDCLIB */
+#ifndef _XLIBC_symbol2string
+#define _XLIBC_symbol2string(x) _XLIBC_symbol2value(x)
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#undef assert
+#ifdef NDEBUG
+    #define assert(ignore) ((void)0)
+#else
+    void _xlibc_assert(char const * const expression, char const * const file_name, char const * const function_name, unsigned long line);
+
+    #define assert(expression) \
+        do { \
+            if(!(expression)) { \
+                _xlibc_assert(_XLIBC_symbol2string(expression), \
+                              __FILE__, \
+                              __func__, \
+                              __LINE__); \
+            } \
+        } while(0)
+#endif
+
+#define static_assert _Static_assert
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* end of include guard: _XLIBC_ASSERT_H */
--- a/lib/xlibc/include/stdarg.h
+++ b/lib/xlibc/include/stdarg.h
@@ -1,0 +1,19 @@
+#ifndef _XLIBC_STDARG_H
+#define _XLIBC_STDARG_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef __builtin_va_list va_list;
+
+#define va_start(v,l) __builtin_va_start((v), (l))
+#define va_end __builtin_va_end
+#define va_arg __builtin_va_arg
+#define va_copy(d,s) __builtin_va_copy((d),(s))
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* end of include guard: _XLIBC_STDARG_H */
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ TOOLS        = cxbe vp20compiler fp20compiler extract-xiso
 NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
                -ffreestanding -nostdlib -fno-builtin -fno-exceptions \
                -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt \
+               -isystem $(NXDK_DIR)/lib/xlibc/include \
                -Wno-ignored-attributes -DNXDK
 NXDK_ASFLAGS = -target i386-pc-win32 -march=pentium3 \
                -nostdlib -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt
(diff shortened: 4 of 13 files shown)


lib/Makefile
```diff
@@ -21,6 +21,8 @@ USB_SRCS := \
 
 SRCS += \
 	$(wildcard $(NXDK_DIR)/lib/xboxrt/*.c) \
+	$(shell find $(NXDK_DIR)/lib/xlibc/src/ -name "*.c") \
+	$(shell find $(NXDK_DIR)/lib/xlibc/src/ -name "*.s") \
```
JayFoxRox: why is this done differently than all the others?

lib/xlibc/include/assert.h
```diff
@@ -0,0 +1,39 @@
+#ifndef _XLIBC_ASSERT_H
+#define _XLIBC_ASSERT_H
+
+#ifndef _xlibc_sym2val
+#define _xlibc_sym2val(x) #x
+#endif
+
+#ifndef _xlibc_sym2str
+#define _xlibc_sym2str(x) _xlibc_sym2val(x)
+#endif
```
JayFoxRox: This macro name is bad in my opinion. Is this a standard name anywhere else?
I feel it implies a lookup from some symbol table. When all it does is "stringify" (for lack of a better term).

lib/Makefile
thrimbor: All other libs have a single directory containing all source files, so wildcard is enough for them. For xlibc, I have several subdirectories (xlibc/src/assert, xlibc/src/setjmp) plus additional source files that don't reside in a subdirectory (not part of this PR), so need to look up these source files recursively. Afaik doing this with wildcard is possible but far uglier than using find.

For the future it might be better to list each file individually (like UBS_SRC) as looking files up on each run of make can be quite slow.

lib/xlibc/include/assert.h
thrimbor: I've named it similar to the macro from PDCLIB (_PDCLIB_symbol2string) which is where I got the idea of how to convert the expression to a string from.
Would you consider renaming it to `_XLIBC_EXPR2STR` (expression instead of symbol, and uppercase to show it's a macro) acceptable?

lib/Makefile
JayFoxRox: Yes, I also agree regarding listing files manually. Not only is the wildcard potentially slower, but it might even include files we don't want to include (like files accidentally left behind, not managed by git).

I'm fine with this solution for now. It would be nice if you could create an issue about listing files manually (I assume other maintainers agree).

lib/xlibc/include/assert.h
JayFoxRox: I think it's better to keep traces to the origin: `_XLIBC_symbol2string` and add a comment that this is actually just stringifying expression, but is named after `_PDCLIB_symbol2string`from PDCLIB.

Alternatively, try to fix this upstream? The underscore suggests this was not meant to be used by outside programs, and the name *is* bad. I personally like your expr2str a lot more - but if it means diverging from upstream it's a bad idea (just confuses people: "why did they change this?!").

Consider this nitpicking. Giving your reasoning, I'm fine with the current solution.

lib/xlibc/include/stdarg.h
```diff
@@ -0,0 +1,19 @@
+#ifndef _XLIBC_STDARG_H
+#define _XLIBC_STDARG_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
JayFoxRox: nit: looks like a bug? In C++ we need cstdarg and it should have the stuff in `std::`?

JayFoxRox:
First post says we use the host stdarg so far. How is that even possible? Isn't the compiler told to ignore the host? If not: Shouldn't it?

lib/xlibc/include/assert.h
```diff
@@ -0,0 +1,41 @@
...
+#define _XLIBC_symbol2string(x) _XLIBC_symbol2value(x)
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#undef assert
+#ifdef NDEBUG
+    #define assert(ignore) ((void)0)
```
JayFoxRox:
```
void foo() {}
```

and later:

```
foo() // Missing ; here!
assert(z == 10);
```

causes a syntax error in my host clang, but not in nxdk.

Error in my clang is:
```
test.c:27:5: error: called object type 'void' is not a function or function
      pointer
    assert(z == 10);
    ^
/usr/include/assert.h:50:24: note: expanded from macro 'assert'
# define assert(expr)           (__ASSERT_VOID_CAST (0))
                                ^
1 error generated.
```

It is independent of NDEBUG in my host.

thrimbor:
> First post says we use the host stdarg so far. How is that even possible? Isn't the compiler told to ignore the host? If not: Shouldn't it?

I too would've thought that `-ffreestanding -nostdlib` was enough to make clang ignore host headers, but apparently not. I just looked through the clang options list and tried out a few options, and `-nobuiltininc` seems to do the trick (meaning clang does complain about stdarg.h not being found on current master).

lib/xlibc/include/stdarg.h
thrimbor: C++ assumes va_start, va_arg, va_copy and va_end are macros (which they are in this case) so they're global even when using cstdarg. Looking at libc++, all cstdarg does is include stdarg.h and import va_list into the std-namespace (so there's va_list and std::va_list).

Since we're using builtins here, we *might* not need to use `extern "C"` (I don't know how it relates to builtins and haven't experimented with it to find out), but I've seen musl use it with builtins in stdarg.h, too.

JayFoxRox:
Even with the following code, I couldn't make `assert` work

```c
#undef NDEBUG
#include <assert.h>
```

`assert(i++)` does not increment `i`. asserting on something `false` would not trigger the assert for me. I have trouble figuring out why that is (maybe my test setup is broken somehow?).

lib/Makefile
JayFoxRox: We can do this later. Someone create an issue please.

lib/xlibc/include/stdarg.h
JayFoxRox: Okay, ignore my comment then. We can review it when there's trouble.

JayFoxRox:
I believe I've tested everything except for stdnoreturn.

Merging.

PR #62, opened 2018-10-16
Deinitialize pbkit on shutdown

### LINKED ISSUE #61: pbkit does not kill itself when application quits
Once pbkit is initialized, it's the users sole responsibility to unload it eventually.
If this does not happen, the GPU is left in a broken state.

I consider this a critical bug, because it might break tools like nxdk-rdt or dashboards made in nxdk.
Basically when you leave an nxdk app uncleanly, it will corrupt the state of the Xbox until a cold-reboot.
So this will not trigger problems in nxdk apps themselves, but applications ran much later.

---

I've prototyped a small change in nxdk-rdt to test this. I hope someone else implements it properly.

Provide this callback and notification routine:

```c
static HAL_SHUTDOWN_REGISTRATION pb_shutdown_notification;

static NTAPI VOID pb_shutdown_notification_routine(PHAL_SHUTDOWN_REGISTRATION shutdown_notifiction) {
  pb_show_debug_screen();
  pb_kill();
}
```

*(Maybe we want to do less or more there. I don't know)*

Do this when `pb_init` is done, to register the above routine (D3D also uses priority 0):

```c
pb_shutdown_notification.NotificationRoutine = pb_shutdown_notification_routine;
pb_shutdown_notification.Priority = 0;
HalRegisterShutdownNotification(&pb_shutdown_notification, TRUE);
```

We also have to unregister the notification if the user manually kills pbkit. I have not implemented this step in the prototype.

This also needs an update in our xboxkrnl.h to fix the calling convention of `PHAL_SHUTDOWN_NOTIFICATION` which should probably be an stdcall (handled by `NTAPI` macro).

thrimbor:
This makes pbkit register a shutdown notification after initialization is done. The registered notification routine then calls pb_kill to deinitialize pbkit. pb_kill now also unregisters the shutdown notification (this is safe to do even when a shutdown is already in progress) to avoid trying to deinitialize twice.

It's based on @JayFoxRox's code from #61 and implements his suggestions.

I tested the code on my Xbox and can confirm that it fixes the issues I've been experiencing previously (stuck screen output after warm reboots on debug bios).

Fixes #61.


lib/pbkit/pbkit.c
```diff
@@ -229,8 +229,11 @@ static  float           pb_BiasTable[7]={
                     1.907f,
                     2.0f    };
 
+static HAL_SHUTDOWN_REGISTRATION pb_shutdown_notification;
```
JayFoxRox: nit: My original code was rather sloppy, I'd probably use `pb_shutdown_registration` now (pbkit is not sure about it's code style? `pb_ShutdownRegistration` is also an option).

thrimbor:
Updated the PR to address the review.

PR #64, opened 2018-10-21
cxbe: Make cxbe compatible with default Xbox debug installations

### LINKED ISSUE #63: cxbe is incompatible with default installations on debug Xboxes
VerifierX (in dxt/vx.dxt) is incompatible with applications generated by cxbe.

The issue is that there's no libraries listed in our XBEs and this trips VerifierX.

I've sent a PR to https://github.com/OpenXDK/cxbe/pull/1 some time ago, but we don't even use that version of cxbe.

Someone should clean up my patch and apply it to nxdk.

dracc:
Copy-Paste of @JayFoxRox's upstream PR with very few 'cleanups'.

Closes #63 .

### DIFF
--- a/tools/cxbe/Xbe.cpp
+++ b/tools/cxbe/Xbe.cpp
@@ -40,6 +40,7 @@
 #include <stdlib.h>
 #include <time.h>
 #include <cstring>
+#include <cstdio>
 
 // construct via Xbe file
 Xbe::Xbe(const char *x_szFilename)
@@ -389,9 +390,10 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail)
         // TODO: generate valid addr if necessary
         m_Header.dwNonKernelImportDirAddr = 0;
 
+        // Need one or more library version, otherwise VerifierX on XDKs crashes
+        m_Header.dwLibraryVersions = 1;
+
         // TODO: generate these values
-        m_Header.dwLibraryVersions = 0;
-        m_Header.dwLibraryVersionsAddr = 0;
         m_Header.dwKernelLibraryVersionAddr = 0;
         m_Header.dwXAPILibraryVersionAddr = 0;
     }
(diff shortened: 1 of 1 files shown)


tools/cxbe/Xbe.cpp
```diff
@@ -686,6 +697,36 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail)
...
 
             hwc = hwc_secn;
+            szBuffer = m_HeaderEx + hwc - (m_Header.dwBaseAddr + sizeof(m_Header));
+        }
+
+        // Write library versions
+        {
+            m_LibraryVersion = new LibraryVersion[m_Header.dwLibraryVersions];
+
+            for(uint32_t v = 0; v < m_Header.dwLibraryVersions; ++v)
```
JayFoxRox: I believe this codebase uses a different type (probably `int`, `unsigned int` or `uint32` [without the `_t`]). Please check surrounding code.

tools/cxbe/Xbe.cpp
```diff
@@ -686,6 +697,36 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail)
...
+            {
+                char tmp[9];
+
+                std::snprintf(tmp, sizeof(tmp), "CXBE%d", v);
```
JayFoxRox: Rest of codebase doesn't eem to care about `std::`

tools/cxbe/Xbe.cpp
```diff
@@ -686,6 +697,36 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail)
             }
 
             hwc = hwc_secn;
+            szBuffer = m_HeaderEx + hwc - (m_Header.dwBaseAddr + sizeof(m_Header));
+        }
+
+        // Write library versions
+        {
```
JayFoxRox: There should pobably be a comment that these are fake libraries.

dracc: Would "Write (placeholder) library versions" work?

JayFoxRox:
Merged, because it seems to work fine for me, and there hasn't been activity for a week already.
If there's any regressions, send a PR to revert or fixup.

tools/cxbe/Xbe.cpp
```diff
@@ -433,8 +435,17 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail)
...
+        // make room for library versions
         {
+            if (m_Header.dwLibraryVersions == 0)
+            {
+                m_Header.dwLibraryVersionsAddr = 0;
+            }
+            else
+            {
+                m_Header.dwLibraryVersionsAddr = mrc;
+                mrc += m_Header.dwLibraryVersions * static_cast<uint32>(sizeof(*m_LibraryVersion));
```
JayFoxRox: nit: Codebase seems to use C style casts.

We should eventually clean the entire code up.
The whole allocation logic and code style is horrible anyway.

PR #70, opened 2018-10-29
Small fix for longjmp return value

thrimbor:
This fixes a small oversight I noticed this morning: longjmp isn't allowed to make setjmp return zero, instead it should return one if the val-parameter was zero.


dracc:
I've no idea where to find the specification for this function, but the assembly code checks out. Lgtm.

Issue #36, opened 2018-03-28
Setup CI build system


mborgerson:
Something like [Travis CI](https://travis-ci.org/) would be great to have to test the build on PR. Maybe [AppVeyor](https://www.appveyor.com/) for Windows nightly build downloads.

JayFoxRox:
I believe we can close this? We have Windows, macOS and Linux CI via Travis now (See #84 and #157).

We only have to add CI badges to README at this point.
Distributing binaries wouldn't be beneficial for this project, as most compile deps are also required later on.

thrimbor:
#246 added badges, so I think it's fine if I close this now.

PR #72, opened 2018-10-30
CI: Add Appveyor support

thrimbor:
This adds basic CI support, building the nxdk-samples on Ubuntu 18.04. I've used Appveyor for this because the travis Ubuntu image is very old and complicates lld installation.
This also adds a small CI helper script to work around a known problem with Appveyor where calling `exit` in code contained in appveyor.yml would hang the build.
The artifacts-part currently packages the whole samples-directory and could surely be improved to not include unnecessary files (like code and .obj-files), but I've decided to keep it simple for now.
I tested the Appveyor support with my own copy of the repo to make sure it actually works.

Of course, Appveyor also has to be enabled for the repository by someone with the necessary access rights (like @mborgerson did for xqemu).

Addresses #36.

### DIFF
--- a/.ci_build_samples.sh
+++ b/.ci_build_samples.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+cd samples
+
+for dir in */
+do
+    cd "$dir"
+    make
+    cd ..
+done
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,14 @@
+image: ubuntu1804
+
+install:
+    - sh: git submodule update --init --recursive
+    - sh: sudo apt-get update >/dev/null
+    - sh: sudo apt-get install -y build-essential flex bison clang lld >/dev/null
+
+build_script:
+    - sh: ./.ci_build_samples.sh
+
+artifacts:
+    - path: samples
+      name: nxdk-samples
+      type: zip


JayFoxRox:
I personally won't be merging this, but I remain neutral in the final decision; if another maintainer accepts this, that's fine.

The reason for staying out of it, is that I *personally* don't like AppVeyor.

I only use it for Windows projects, and I even consider switching away from AppVeyor for those.

My experience with AppVeyor has been mostly negative:

- Build image breaks often, and it takes weeks to fix, that will often lead to issues with PRs not being reviewed because everyone just sees the red cross.
- Only the previous build image is available, but they sometimes fix minor bugs, so suddenly you are only left with 2 almost identical build images (both non-functional).
- AppVeyor organization accounts are absolutely horrible; there's many permission issues with naive setups, and you need an e-mail address for your organization.
- Build-times and queue-length on AppVeyor are often considerable. I had even simple projects waiting on AppVeyor build for *hours*.
- Ubuntu / Linux support on AppVeyor is pretty recent, and we can't know what might happen in the future.

I also dislike that Travis images are old, but I still consider Travis a *much* better platform.
There's also acceptable solutions like PPAs or even Docker (which I'm not too fond of).

That said: the work has been done already, and not everyone dislikes AppVeyor as much as me. So rather than rejecting this outright, I'm open to someone else reviewing (and maintaining) this CI platform.

PR #74, opened 2018-11-04
[Don't merge] C11 threads and thread-local storage

thrimbor:
**Note: I plan to submit a PR that will supersede this soon.**

~~**This shouldn't get merged yet, it needs more testing.** I still decided to create this PR to potentially gather some feedback.
 I'll make sure to write some test code in the upcoming days and also to ensure that no commit breaks building (they were rearranged quite a bit). I do already have local testing code for threads and static thread-local storage.~~
**PR should be ready now. All changes are tested and testcode is linked.**

This PR adds the threading part of C11(call_once, threads, mutexes and thread-specific storage) and static thread-local storage (`thread_local` and `__declspec(thread)`). One time.h function is also implemented as it is required for `thrd_sleep`, and `main()` is now run in a separate thread because TLS does not get set up for the thread running the entry point. Also, the entry function is replaced with a slightly stripped version of my own code, although the initial version I added looks very similar to the pre-existing one for obvious reasons.

The most recent and complex part is TLS support. It's based on what I learned from [this blog series](http://www.nynaeve.net/?p=190), the "Windows Internals" book and a lot of Trial&Error to figure out the Xbox specifics.

**One important change that isn't included in this PR** is that XCreateThread will have to get removed, as it is fundamentally incompatible with properly setting up the new thread (TSS). Afaik the network code is the only place it is internally used, anyway.

/edit:
Test progress:

- [x] Threads
    - [x] [thrd_create](https://gist.github.com/thrimbor/3b34f1a6bd1d9c874ad9cc499c9f10b9)
    - [x] [thrd_current](https://gist.github.com/thrimbor/8054503d3894ddab36a1fa548a12898f)
    - [x] thrd_detach (Not really testable, but should work)
    - [x] [thrd_sleep](https://gist.github.com/thrimbor/23285c4f9ccc0900cf9d41f87b3589a2)
    - [x] thrd_yield (Not really testable, but simply calls NtYieldExecution)
    - [x] [thrd_equal](https://gist.github.com/thrimbor/8054503d3894ddab36a1fa548a12898f)
    - [x] [thrd_join](https://gist.github.com/thrimbor/3b34f1a6bd1d9c874ad9cc499c9f10b9)
    - [x] [thrd_exit](https://gist.github.com/thrimbor/ec6a8b0e7b1f7d6846fe4c72fb2f6186)
- [x] [thread_local](https://gist.github.com/thrimbor/f45184e6fade24fdf0f121d04c1cdd8e) (static TLS)
- [x] [call_once](https://gist.github.com/thrimbor/e133d15a2be1ce60fd62592d3b63aa28)
- [x] Mutexes
    - [x] mtx_destroy (not really testable, but should work)
    - [x] [mtx_init](https://gist.github.com/thrimbor/84bb003cc9b2b3f5f9a5b7d37e0682f9)
    - [x] [mtx_lock](https://gist.github.com/thrimbor/84bb003cc9b2b3f5f9a5b7d37e0682f9)
    - [x] [mtx_timedlock](https://gist.github.com/thrimbor/7247bd1d57a503507412c81f2fdcb6d2)
    - [x] [mtx_trylock](https://gist.github.com/thrimbor/8bc90eabd0968ee6858d36c14a101cd1)
    - [x] [mtx_unlock](https://gist.github.com/thrimbor/647fd2febedcda19da809f80c6afbeb3)
- [x] TSS
    - [x] [tss_create](https://gist.github.com/thrimbor/7d4654271c5e632d4f73d1dc0d072cc7)
    - [x] [tss_delete](https://gist.github.com/thrimbor/f60d5cf39e6e3ea7778bd96b44a94d7e)
    - [x] [tss_get](https://gist.github.com/thrimbor/c41c174dee1abf41b915eb2f4b90b58a)
    - [x] [tss_set](https://gist.github.com/thrimbor/c41c174dee1abf41b915eb2f4b90b58a)
- [x] Condition variables
    - [x] [cnd_broadcast](https://gist.github.com/thrimbor/74a6d1a26eea30a52e964fbd0d72d2b2)
    - [x] cnd_destroy (Not really testable, but should work)
    - [x] [cnd_init](https://gist.github.com/thrimbor/59ff937d3997d04c829c1eb874b79985)
    - [x] [cnd_signal](https://gist.github.com/thrimbor/d0e28760a43eb979ca97dc1c8768a963)
    - [x] [cnd_timedwait](https://gist.github.com/thrimbor/59ff937d3997d04c829c1eb874b79985)
    - [x] [cnd_wait](https://gist.github.com/thrimbor/74a6d1a26eea30a52e964fbd0d72d2b2) (Forwards to cnd_timedwait internally)
- [x] Time
    - [x] [timespec_get](https://gist.github.com/thrimbor/7247bd1d57a503507412c81f2fdcb6d2)


lib/xlibc/src/threads/tss_create.c
```diff
@@ -0,0 +1,37 @@
+#include <threads.h>
+#include <assert.h>
+#include <_tss.h>
+
+mtx_t tss_lock;
+// Note: This uses range-based initialization, which is a GNU C extension
```
JayFoxRox: Is this really necessary?

lib/xlibc/include/threads.h
```diff
@@ -75,10 +75,10 @@ int thrd_join (thrd_t thr, int *res);
...
 void thrd_yield (void);
 
-int tss_create (tss_t *key, tss_dtor_t dtor) __unimplemented;
-void tss_delete (tss_t key) __unimplemented;
-void *tss_get (tss_t key) __unimplemented;
-int tss_set (tss_t key, void *val) __unimplemented;
+int tss_create (tss_t *key, tss_dtor_t dtor);
+void tss_delete (tss_t key);
+void *tss_get (tss_t key);
+int tss_set (tss_t key, void *val);
```
JayFoxRox: Did you consider implementing a standard API for this instead? I think using win32 function names / signatures would make a lot of sense here.

lib/xlibc/include/time.h
```diff
@@ -0,0 +1,58 @@
+#ifndef _XLIBC_TIME_H
+#define _XLIBC_TIME_H
+
+#include <_restrict.h>
+#include <_unimplemented.h>
```
JayFoxRox: This might be displayed in wrong-order on GitHub, but this include file does not exist in this commit.

JayFoxRox:
" xlibc: Add new XBoxCRTEntry entry point including TLS index initializ… " commit title is cut-off

lib/xlibc-rt/_tls_array.s
```diff
@@ -0,0 +1,5 @@
+.data
+
+// _tls_array is a constant describing the offset of StackBase inside the NT TEB
+.globl __tls_array
+.equ __tls_array, 04h
```
JayFoxRox: What's the purpose of this being in assembly?

lib/xlibc-rt/tls.c
```diff
@@ -0,0 +1,20 @@
+#include "xboxkrnl/xboxkrnl.h"
+#include "_tls.h"
+
+ULONG _tls_index = 0;
+
+#pragma data_seg(".tls")
+char _tls_start = 0;
+#pragma data_seg("tls$ZZZ")
+char _tls_end = 0;
```
JayFoxRox: Does our linker script (which one do we even use?) guarantee preservation of this section ordering? What's the "$ZZZ" suffix? Why isn't it ".tls$ZZZ"? Do you have any docs for this?

Also, does cxbe preserve these sections (or do they get merged elsewhere)?

lib/xlibc/src/threads/thrd_create.c
```diff
@@ -1,13 +1,18 @@
...
 int thrd_create (thrd_t *thr, thrd_start_t func, void *arg)
 {
     NTSTATUS ntstatus;
     ULONG stacksize;
+    ULONG tlssize;
+    // FIXME: We're directly reading the XBE StackCommit field here.
+    //        A cleaner way with proper structs would be nice.
     stacksize = *((ULONG *)0x00010130);
-    ntstatus = PsCreateSystemThreadEx(thr, 0, stacksize, 0, NULL, (PKSTART_ROUTINE)func, arg, FALSE, FALSE, _xlibc_thread_startup);
+    tlssize = _tls_used.SizeOfZeroFill + _tls_used.EndAddressOfRawData - _tls_used.StartAddressOfRawData + 4;
```
JayFoxRox: Where does the `+ 4` come from? Even if there's no TLS data, shouldn't `(EndAddressOfRawData - StartAddressOfRawData)` be 1 (or larger due to alignment)?
I'd expect `- 1` or something.

lib/xlibc/src/threads/_xlibc_thread_startup.c
```diff
@@ -1,8 +1,31 @@
...
+    assert(((DWORD)TlsData & 15) == 0);
+
+    // copy initial values of TLS area
+    DWORD TlsDataSize;
+    TlsDataSize = _tls_used.EndAddressOfRawData - _tls_used.StartAddressOfRawData;
+    memcpy(TlsData, (void *)_tls_used.StartAddressOfRawData, TlsDataSize);
+
+    // zero-initialize the rest
+    if (_tls_used.SizeOfZeroFill != 0)
+        RtlZeroMemory((char *)TlsData + TlsDataSize, _tls_used.SizeOfZeroFill);
```
JayFoxRox: Explicit scope please.

Not sure if we do it all over the codebase yet, but *I* think we should.

lib/xlibc/src/threads/_xlibc_thread_startup.c
```diff
@@ -1,8 +1,31 @@
...
+    // 16 byte alignment required
+    assert(((DWORD)TlsData & 15) == 0);
+
+    // copy initial values of TLS area
+    DWORD TlsDataSize;
+    TlsDataSize = _tls_used.EndAddressOfRawData - _tls_used.StartAddressOfRawData;
+    memcpy(TlsData, (void *)_tls_used.StartAddressOfRawData, TlsDataSize);
+
+    // zero-initialize the rest
+    if (_tls_used.SizeOfZeroFill != 0)
```
JayFoxRox: nit: I prefer `> 0`; this condition also shouldn't even be necessary?

lib/xlibc/include/time.h
thrimbor: GitHub seems to order the commits by time, which is all mixed up since I resorted the commits. When I look at my branch, the commits are in the correct order.

lib/xlibc/src/threads/tss_create.c
thrimbor: Every element of these arrays needs to be initialized before using any TSS-touching code, so my options were:
1. Introduce an initialization function which gets run before creating the thread for main() (ugly imho and would shift initialization to runtime)
2. Use humongous amounts of dirty macro magic, which expands to a proper initializer list
3. Use the GNU extension

I went with 3 because it seemed the least messy to me. Also I would have expected C to have this as a properly standardized feature already.

lib/xlibc-rt/tls.c
thrimbor: > Why isn't it ".tls$ZZZ"?

It should be ".tls$ZZZ", the missing dot is a typo. Interesting that it still worked, though.

> What's the "$ZZZ" suffix?

The "$ZZZ" suffix is a way to put stuff at the end of a section.
I can't find MS documentation about it right now, but [this](http://www.nynaeve.net/?p=183) contains a description of the behavior near the end.

> Also, does cxbe preserve these sections (or do they get merged elsewhere)?

When inspecting the exe-file, it seems that lld merges the sections (into rdata, I assume), which is fine, since we really only need it to generate IMAGE_TLS_DIRECTORY.

lib/xlibc/src/threads/thrd_create.c
thrimbor: Right, there's a self-reference. I'll do the reordering and add a comment.

lib/xlibc/src/threads/_xlibc_thread_startup.c
thrimbor: Will do.

lib/xlibc/src/threads/_xlibc_thread_startup.c
thrimbor: Right, I'll get rid of it.

lib/xlibc/src/threads/_xlibc_thread_startup.c
```diff
@@ -1,8 +1,31 @@
...
+
+    // Retrieve TLS area address and point first entry to itself
+    DWORD *TlsData;
+    TlsData = ((DWORD *)KeGetCurrentThread()->TlsData) + 1;
+    TlsData[-1] = (DWORD)TlsData;
+
+    // 16 byte alignment required
+    assert(((DWORD)TlsData & 15) == 0);
+
+    // copy initial values of TLS area
```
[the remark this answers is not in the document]
thrimbor: I'll go through and change them to uppercase.

lib/xlibc-rt/tls.c
```diff
@@ -0,0 +1,20 @@
...
+
+#pragma data_seg(".rdata")
+const IMAGE_TLS_DIRECTORY _tls_used =
+{
+    (DWORD) &_tls_start,
+    (DWORD) &_tls_end,
+    (DWORD) &_tls_index,
+    (DWORD) NULL,
+    (DWORD) 0,
+    (DWORD) 0
```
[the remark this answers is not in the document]
thrimbor: Yes, the cast is unnecessary, I only kept it because I found the formatting to be easier on the eye this way. I can remove it though I you prefer so.

lib/xlibc/src/crt0.c
```diff
@@ -0,0 +1,30 @@
+#include <xboxkrnl/xboxkrnl.h>
+#include <hal/video.h>
+#include <threads.h>
+#include <_tss.h>
+#include <_tls.h>
+
+extern int main(int argc, char **argv);
+
+static int main_wrapper ()
```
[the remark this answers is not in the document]
thrimbor: I usually put in a space when declaring or defining functions, but it seems this one slipped my mind. I'll change it.

thrimbor:
PR is updated.

lib/xlibc/src/threads/thrd_current.c
```diff
@@ -0,0 +1,8 @@
+#include <threads.h>
+#include "xboxkrnl/xboxkrnl.h"
+
+thrd_t thrd_current (void)
+{
+    // This is just a pseudohandle (enough for thrd_equal)
+    return (thrd_t) 0xFFFFFFFE;
```
mborgerson: Why not `KeGetCurrentThread` or something otherwise unique?

thrimbor: That would only solve the problem (`thrd_current()` only being useful for `thrd_equal()`) if I could use the KTHREAD pointer for everything else as well (like `thrd_join()`) or somehow use it to get the original handle (although that's probably not how handles are meant to be used). Also, I'd have to somehow get the KTHREAD pointer upon thread creation.
I thought about using the thread ID (a.k.a. `UniqueThread`, which you can get from `PsCreateSystemThreadEx()`), but that suffers the same problem.

JayFoxRox:
This mostly landed in https://github.com/XboxDev/nxdk-pdclib/commit/0dc4023d6ebf138950bd5237126ad5742d2eef51 right?

- Can this be closed?
- Do we need a follow-up for the cxbe changes?

thrimbor:
Right, #85 includes everything this PR provides (it includes the cxbe changes as well), so I'm closing this.

PR #76, opened 2018-11-11
[Don't merge] Add improved printf-style functions

thrimbor:
**Note: I plan to submit a PR that will supersede this soon.**

This adds the stdio.h header and brings over PDCLib's formatted printing with minimal changes. I built and ran all the nxdk-samples on my Xbox to make sure nothing broke due to the changes, as well as [this small sample](https://gist.github.com/thrimbor/348acca0932f1f10e2650e7d845517d3) based on [cppreference example code](https://en.cppreference.com/w/c/io/fprintf). The indentation in my sample code is off, but this is a separate issue in `debugPrint`.
This does not support printing floats yet.

~~**Important:** This requires 845f4056caac25355fe725a10728b41774ffe3cd from #74. One solution would be to merge #74 first and rebase this PR, but I'd suggest cherry-picking the relevant commit into master and rebasing both PRs.~~

### DIFF
--- a/lib/xboxrt/string.c
+++ b/lib/xboxrt/string.c
@@ -47,6 +47,16 @@ size_t strlen(const char *s1) {
   return i;
 }
 
+size_t strnlen( const char * s, size_t maxlen )
+{
+    for( size_t len = 0; len != maxlen; len++ )
+    {
+        if(s[len] == '\0')
+            return len;
+    }
+    return maxlen;
+}
+
 char *strdup(const char *s1) {
     char *out = malloc(strlen(s1) + 1);
     strcpy(out, s1);
--- a/lib/xboxrt/stdlib.c
+++ b/lib/xboxrt/stdlib.c
@@ -3,148 +3,6 @@
 #include "ctype.h"
 #include <xboxkrnl/xboxkrnl.h>
 
-static const char digits[] = "0123456789abcdef";
-
-static unsigned int find_divisor(unsigned long val, unsigned int base)
-{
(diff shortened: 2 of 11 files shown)


lib/xboxrt/string.c
```diff
@@ -47,6 +47,16 @@ size_t strlen(const char *s1) {
   return i;
 }
 
+size_t strnlen( const char * s, size_t maxlen )
+{
+    for( size_t len = 0; len != maxlen; len++ )
```
JayFoxRox: nit: I always feel safer with "< maxlen" style.

thrimbor:
Using PDCLib directly was my actual plan ~2 years ago. Getting file I/O working was one of the first few things I tried (I created a copy of the win32 backend and went from there), but I was experiencing seemingly random crashes when writing to files, which, despite working on it for almost two weeks, I couldn't fix.

Considering how much I learned in the meantime about how Xbox/libc/MSVC stuff works internally, it might actually be worth it giving it another shot. For now, I've sent the PDCLib author a mail with a few questions.
We'd still require some additions though, mainly the stuff currently residing in xlibc-rt plus a libm/math.h (which is what I'm looking at right now, because I tried porting a game).

JayFoxRox:
I assume this will be replaced by #85 and has already partially landed in https://github.com/XboxDev/nxdk-pdclib

Can we close this?

thrimbor:
This PR included a subset of the PDCLib code, so I'm closing this as the functionality is also provided by #85.

PR #77, opened 2018-11-11
Update BOOL type to be correct size according to Microsoft Documentation

Fisherman166:
BOOL should be a typedef of size int according to Microsoft Documentation: https://docs.microsoft.com/en-us/windows/desktop/winprog/windows-data-types#BOOL

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -34,7 +34,8 @@ extern "C"
 #define CONST const
 
 typedef unsigned int SIZE_T, *PSIZE_T;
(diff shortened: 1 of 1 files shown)


0DaveX:
The linked Documentation states a byte type 
> typedef BYTE BOOLEAN;
The current code (unsigned char) looks correct.

JayFoxRox:
We are using a prefix style for our commit names now (make sure it isn't too long though).

I'd recommend: `xboxkrnl: BOOL is byte-sized now` or something.

Fisherman166:
Amended commit message to fit prefix format.

JayFoxRox:
Thanks a lot!

*(Also kudos for paying attention - I had actually messed up in my recommended name)*

Issue #13, opened 2017-07-08
nxdk has no proper libc


JayFoxRox:
* `printf` is absolutely horrible and doesn't respect most format specifiers (not Xbox specific).
A better minimalistic alternative is even present in cromwell
* `fopen` and friends are missing (Xbox specific).
* I'd appreciate C11 thread compatibility too (Xbox specific).

I'd like to hear about peoples opinion which libc we should use or if we should implement our own.
I have never ported a libc and I'm not sure how C++ compatibility would play into this in the future.

I'm also not sure how a libc would handle the platform specific parts in contrast to the non-platform specific parts (such as the `printf` stuff)

However, this is an important issue which causes a lot of trouble in any Xbox related homebrew I have.

thrimbor:
Maybe this is the right time to talk about what I've been doing.
Around the same time as nxdk was started, I started doing my own XDK, but instead of basing it on OpenXDK, I started from scratch. It's far from being complete, but it has the following features:

1. A 100% complete xboxkrnl.h (related to #14). It's monolithic (but could be split up), not fully documented (large parts of it are documented using Doxygen) and may have a constant or two missing, but when it comes to the function declarations, I'm pretty confident it's complete and correct.
2. An xboxkrnl.lib built from "source" (i.e. a definition-file)
3. A libc. It's far from being complete (fopen isn't working yet, printf isn't implemented) and may not be bug-free, but it also has some improvements over what nxdk is currently using:
    **1.** Correct string and memory comparison. The code you took from lib43 may be working for you, but it's not fulfilling the requirements the specification states.
    **2.** Far more math.h functions.
    **3.** More efficient memory allocation (nxdk's malloc always reserves complete 4KiB pages which I consider wasteful considering memory is pretty tight on the XBox)
    **4.** The ability to use a standards-compliant main() function
    **5.** C11 threads, mutexes and call_once.

Missing parts like printf could get picked from the CC0-licensed "pdclib", which some of my code is already based on.

There are several more things I worked on or experimented with like an OpenAL-implementation with some tricks to make it run faster than existing ones, C++ support and support for thread-specific storage (which is especially tricky because cxbe simply ignores TSS).

However, there are reasons why I didn't already contribute to nxdk or base my XDK on OpenXDK code.
[...]

PR #79, opened 2018-11-13
xlibc: Add wchar.h

Fisherman166:
Add a subset of wchar.h functions that I need for the xbox kernel test suite. Most functions were copied from string.h/string.c and modified to work with wchar_t.

### DIFF
--- a/lib/xboxrt/wchar.c
+++ b/lib/xboxrt/wchar.c
@@ -1,0 +1,43 @@
+#include "wchar.h"
+
+#include <stdint.h>
+
+wchar_t *wcscpy(wchar_t *dst, const wchar_t *src) {
+    return wcsncpy(dst, src, SIZE_MAX);
+}
+
+wchar_t *wcsncpy(wchar_t *dst, const wchar_t *src, size_t n) {
+    size_t i = 0;
+    do dst[i] = *src; while (*src++ != L'\0' && i++ < n);
+    return dst;
+}
+
+wchar_t *wcscat(wchar_t *s1, const wchar_t *s2) {
+    return wcsncat(s1, s2, SIZE_MAX);
+}
+
+wchar_t *wcsncat(wchar_t *s1, const wchar_t *s2, size_t n) {
+    wcsncpy(s1 + wcslen(s1), s2, n);
+    return s1;
+}
+
+size_t wcslen(const wchar_t *s1) {
+    size_t len = 0;
+    while(s1[len] != L'\0') { len++; }
+    return len;
+}
+
+int wcscmp(const wchar_t *s1, const wchar_t *s2) {
+    return wcsncmp(s1, s2, SIZE_MAX);
+}
+
+int wcsncmp(const wchar_t *s1, const wchar_t *s2, size_t n) {
+    for (size_t pos = 0; pos < n; pos++) {
+        wchar_t c1 = *s1++;
+        wchar_t c2 = *s2++;
+        if (c1 < c2) return -1;
+        if (c1 > c2) return 1;
+        if (c1 == 0 || c2 == 0) break;
+    }
+    return 0;
+}
--- a/lib/xboxrt/wchar.h
+++ b/lib/xboxrt/wchar.h
@@ -1,0 +1,14 @@
+#ifndef XBOXRT_WCHAR
+#define XBOXRT_WCHAR
+
+#include <stddef.h>
+
+wchar_t *wcscat(wchar_t *s1, const wchar_t *s2);
+wchar_t *wcscpy(wchar_t *dst, const wchar_t *src);
+wchar_t *wcsncat(wchar_t *s1, const wchar_t *s2, size_t n);
+wchar_t *wcsncpy(wchar_t *dst, const wchar_t *src, size_t n);
+size_t wcslen(const wchar_t *s1);
+int wcscmp(const wchar_t *s1, const wchar_t *s2);
+int wcsncmp(const wchar_t *s1, const wchar_t *s2, size_t n);
+
+#endif


JayFoxRox:
Some questions for @Thrimbor which could affect my decision to merge this:

- What's the difference between xboxrt and xlibc?
    I curently assume xlibc will be the new libc which is clean, while xboxrt is the legacy, dirty version that will eventually be removed?
- Should such edits be prefixed with xboxrt or xlibc?
- Should the stdbool.h also be moved to xlibc (it has an XLIBC include guard)?
    Your stdbool was prefixed xlibc but went to xboxrt - xlibc was created afterwards I believe.
- What's your plan on moving further libc functions? if you do the string stuff *very soon* then I'd probably vote to reject / postpone this.

---

Personally I'm fine with adding more dirty-libc - it's still better than master. But we also need a plan to get rid of this dirty code.

A couple of weeks ago, I had to fix bugs in `strlen()` - now this code will be duplicated. It will become an unmaintainable mess if we keep working like this.

Ideally we'd use an external libc (see #13) with as few changes as possible, so we can stay in synchronization with upstream, and benefit from a fully tested standard C implementation.

lib/xboxrt/wchar.c
```diff
@@ -0,0 +1,43 @@
...
+    do dst[i] = *src; while (*src++ != L'\0' && i++ < n);
+    return dst;
+}
+
+wchar_t *wcscat(wchar_t *s1, const wchar_t *s2) {
+    return wcsncat(s1, s2, SIZE_MAX);
+}
+
+wchar_t *wcsncat(wchar_t *s1, const wchar_t *s2, size_t n) {
+    wcsncpy(s1 + wcslen(s1), s2, n);
```
JayFoxRox: Just to prove my point: Shouldn't the last parameter be something like `n - wcslen(s1)` ?

Even if this is not your bug, it will be harder to fix them across our codebase. Especially because this poor / old libc isn't tested at all.

I'm really worried about stability of the software that has been compiled with these "frankensteined" versions of nxdk.

*You should not fix this yet; it's just food for thought. If we want to fix it, we should fix it in the other string function, too.*

lib/xboxrt/wchar.c
```diff
@@ -0,0 +1,43 @@
...
+#include <stdint.h>
+
+wchar_t *wcscpy(wchar_t *dst, const wchar_t *src) {
+    return wcsncpy(dst, src, SIZE_MAX);
+}
+
+wchar_t *wcsncpy(wchar_t *dst, const wchar_t *src, size_t n) {
+    size_t i = 0;
+    do dst[i] = *src; while (*src++ != L'\0' && i++ < n);
+    return dst;
```
JayFoxRox: `man wcsncpy`

> Exactly n wide characters are written at dest. If the length wcslen(src) is smaller than n, the remaining wide characters in the array pointed to by dest are filled with null wide characters.

Another error that propagated: we currently don't fill exactly `n` bytes (and neither could we, because our users don't respect that - see `wcscpy` which calls with `SIZE_MAX`)

*Same as other instance: don't fix, we should fix this all at once.*

lib/xboxrt/wchar.c
JayFoxRox: Yes indeed. I was wrong.

*My argument still holds with the other comment below (I hope): errors might propagate*

thrimbor:
> What's the difference between xboxrt and xlibc?
> I curently assume xlibc will be the new libc which is clean, while xboxrt is the legacy, dirty version that will eventually be removed?

Yes, that's the plan. xboxrt has untested code with mixed licenses, while xlibc's code comes from PDCLib (which is CC-0 and has tests) or was written by me (like the threads part).

> Should such edits be prefixed with xboxrt or xlibc?

Since this adds code to the old libc, it should be xboxrt.

> Should the stdbool.h also be moved to xlibc (it has an XLIBC include guard)?
> Your stdbool was prefixed xlibc but went to xboxrt - xlibc was created afterwards I believe.

[It should already be in xlibc, shouldn't it?](https://github.com/XboxDev/nxdk/commit/aa35bcd1a6b57a54c3a77cda41119383eb4db19f)

> What's your plan on moving further libc functions? if you do the string stuff _very soon_ then I'd probably vote to reject / postpone this.

I'll probably tackle ctype.h, math.h, stdlib.h and string.h soon. I have (mostly PDCLib-based) code for these already, but I need to finish and polish it. wchar.h doesn't have a very high priority for me, since it's not as critical as the rest for getting more code working with nxdk out of the box.
Personally, I'm fine with this PR adding that code (but it might get replaced in the future, and I wouldn't necessarily trust test-results based on it).

JayFoxRox:
>> Should the stdbool.h also be moved to xlibc (it has an XLIBC include guard)?
>> Your stdbool was prefixed xlibc but went to xboxrt - xlibc was created afterwards I believe.

> It should already be in xlibc, shouldn't it?

Yes. I was wrong.

*GitHub shows the commit log in the tree-view on the folder where the old stdbool.h was deleted, so I got confused.*

---

Given the explanation given by @thrimbor, I'm fine with accepting this as-is. I see the value of having this.

Any other comments @XboxDev ? If nobody opposes I'll "Rebase and merge" this later.

@Fisherman166 could you please add an issue about the bug I've located? If you feel like it, even send a separate PR to address it.

Fisherman166:
Issue filed here: https://github.com/XboxDev/nxdk/issues/81

Fisherman166:
This pull request should likely be killed in favor of this new one which is a port from pdclib and is stored correctly in the xlibc folder.

https://github.com/XboxDev/nxdk/pull/82

mborgerson:
Closed per @Fisherman166

PR #82, opened 2018-11-18
xlibc: Add wchar.h

Fisherman166:
Using the shepherd branch of pdclib, I ported over many of the wchar functions. This should replace my previous pull request which was looking to add a legacy wchar implementation in xboxrt.

### DIFF
--- a/lib/xboxrt/wchar.c
+++ b/lib/xboxrt/wchar.c
@@ -1,0 +1,325 @@
+#include "wchar.h"
+
+wchar_t * wcscat( wchar_t * XBOXRT_RESTRICT s1, const wchar_t * XBOXRT_RESTRICT s2 )
+{
+    wchar_t * rc = s1;
+    if ( *s1 )
+    {
+        while ( *++s1 );
+    }
+    while ( (*s1++ = *s2++) );
+    return rc;
+}
+
+wchar_t *wcschr(const wchar_t * haystack, wchar_t needle)
+{
+    while(*haystack) {
+        if(*haystack == needle) return (wchar_t*) haystack;
+        haystack++;
+    }
+    return NULL;
+}
+
+int wcscmp( const wchar_t * s1, const wchar_t * s2 )
+{
+    while ( ( *s1 ) && ( *s1 == *s2 ) )
+    {
+        ++s1;
+        ++s2;
+    }
+    return ( *(wchar_t *)s1 - *(wchar_t *)s2 );
+}
+
+int wcscoll( const wchar_t * s1, const wchar_t * s2 )
+{
+    return wcscmp(s1, s2);
+}
+
+wchar_t *wcscpy( wchar_t * XBOXRT_RESTRICT dest, const wchar_t * XBOXRT_RESTRICT src)
+{
+    wchar_t * rv = dest;
+    while( (*dest++ = *src++) );
+    return rv;
+}
+
+size_t wcscspn( const wchar_t * s1, const wchar_t * s2 )
+{
+    size_t len = 0;
+    const wchar_t * p;
+    while ( s1[len] )
+    {
+        p = s2;
+        while ( *p )
+        {
+            if ( s1[len] == *p++ )
+            {
+                return len;
+            }
+        }
+        ++len;
+    }
+    return len;
+}
+
+size_t wcslen( const wchar_t * str )
+{
+    size_t n = 0;
+    while(*(str++)) n++;
+    return n;
+}
+
+wchar_t * wcsncat( wchar_t * XBOXRT_RESTRICT s1, const wchar_t * XBOXRT_RESTRICT s2, size_t n )
+{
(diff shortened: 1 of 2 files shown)


Fisherman166:
The style in my PR is the style used in PDCLib. I made as few modifications to the code as possible (none were needed other than naming changes for XLIBC). Long term, style should be enforced through ClangFormat or a similar tool.

Fisherman166:
I can move this code into the xboxrt and delete my other pull request as this is better code than what I had before. 

Do you have a write up somewhere of what you want the xlibc format to look like for future reference? All I did with this pull request was copy collapse all of the PDClib functions into a single .c file, and then made as few changes as possible to get it compile. Restricted pointers. for example, are used in the functions inside of PDClib.

thrimbor:
> I can move this code into the xboxrt and delete my other pull request as this is better code than what I had before.

I would prefer that. The maintainers have the final word, though.
> Do you have a write up somewhere of what you want the xlibc format to look like for future reference? All I did with this pull request was copy collapse all of the PDClib functions into a single .c file, and then made as few changes as possible to get it compile.

I haven't written it down anywhere, but for xlibc, I tried to stick to how PDCLib handles things. That means a separate file per function (unless its a required static function or similar) and sticking as close to the C standard as possible (it has strict rules on what a header file is allowed to include).

> Restricted pointers. for example, are used in the functions inside of PDClib.

Yes, but the restrict-keyword is also important in the declaration. Otherwise, the compiler won't be able to warn you in case you violate the restriction with pointer aliasing.

Fisherman166:
Bumping this again. Any feedback on what I can do to get this in?

JayFoxRox:
> This has kind of stalled out.

Yes, very sorry! Not sure what happened. The other PRs on this repository were also forgotten apparently.

We need more (active) maintainers.
XboxDev has too many side-projects and active tasks.

If a small PR stalls this badly, please contact us directly. You can use @XboxDev, create an issue at https://github.com/XboxDev/XboxDev/issues or, ideally, [contact us directly via chat](https://github.com/XboxDev/XboxDev).

> @thrimbor makes it sound like it isn't his call whether or not this goes in.

Thrimbor is the one who set up the directory structure and worked on all existing xboxrt / xlibc code (see history / blame) Trust his advice on this one, even if his say isn't final - this is open-source, so everyone has an equal say during review. Only the final merge is done by maintainers.

> If I move this into xboxrt and add the restrict keyword to the function definitions, would that get this pull request accepted?

Yes. I'd think so.

I'll have to trust @thrimbor: this looked like this was meant for xlibc, but if he says xboxrt is correct, then that' it.

- Asking @Fisherman166 to move files, as suggested by @thrimbor 
- Asking @thrimbor to review *again* for final comments.
- Asking any other users to test if they can.
- Asking all other maintainers to review and decide to merge, especially in a couple of days (if @thrimbor has responded, or doesn't respond).

thrimbor:
My suggested changes for this PR (also with regards to compatibility with #85 or another upstream libc) would be:

1. Rebase onto current master (just to make sure Travis CI builds the PR at least once).
2. Move your code from the xlibc namespace to xboxrt (includes directory, include guards and struct names).
3. Add a macro to wchar.h to provide an alias for `restrict`/`__restrict`, similar to [this](https://github.com/XboxDev/nxdk/blob/master/lib/xlibc/include/_restrict.h) (xlibc does provide _restrict.h atm, but that might not be here to stay, so I suggest this for forward compatibility)
4. Replace all occurrences of `XLIBC_RESTRICT` in `wchar.c` with your own alias.
5. Make sure the declarations in `wchar.h` match the definitions in `wchar.c`, adding your restrict-alias where necessary.

That should fix the issues I see and make it easy to keep or replace the code in the future.

.ci_build_samples.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -e
+cd samples
+for dir in */
+do
+    cd "$dir"
+    make
+    cd ..
+done
```
JayFoxRox: Why are these commits part of this PR? Please re-base on latest master.

Also squash your 2 commits: no need for anybody to see that wchar was in "xlibc" first.
When squashing, keep the original commit title (which doesn't overflow, unlike the title of the last commit).

Fisherman166: Sorry about that. Been a while since I've had to do this and messed up. Should be better now.

JayFoxRox:
I have merged after a quick review, because nobody else seems to be reviewing / testing.

I did not test the code in question as no sample was supplied. If there's anything wrong with it, we'll have to iterate. But it's really bad if we have a PR sitting open for months, in a state like this (as there's projects which depend on these kind of changes).

Sorry again for the long "review" (or lack thereof) @Fisherman166; I hope we'll handle things better in the future.
Thanks for being so responsive, even after weeks of delays; also thanks to @thrimbor for review and recommendations!

PR #84, opened 2018-12-05
CI: Add Travis CI support for Linux and macOS

thrimbor:
Basically the same as #72, but with Travis instead of Appveyor and additional macOS support and without storing artifacts. macOS builds are rather slow compared to Linux builds (~8m vs ~1m) due to homebrew. It seems caching can be used to improve that, but messing with homebrew in that way is complicated and error-prone, so I didn't bother.

Travis doesn't seem to support msys ootb, therefore I didn't try adding Windows support.

Also addresses #36.

### DIFF
--- a/.ci_build_samples.sh
+++ b/.ci_build_samples.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+set -e
+cd samples
+for dir in */
+do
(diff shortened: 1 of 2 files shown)


mborgerson:
Thanks!

Issue #41, opened 2018-04-30
'triangle' sample gives GPU Error at current state


jthecodebuilder:
I don't know anything about programming the Xbox's GPU at the moment, but here is the error I am getting when running the output default.xbe:
(Keep in mind I had to hand copy this by hand as I do not have the proper means for debug log copy and paste here.)
```
GPU Error : object state invalid error!
Error binary flags : 08x
Channel ID : 2048 (0=3D)
Channel class : 0
Push buffer inner register target : 04x
Push buffer data (lo) or instance : 08x
Push buffer data (hi) or instance : 08x
Multi-purpose register A [0x1D8C] : 08x
Multi-purpose register A [0x1D90] : 08x

Report is accurate only if pb_trace_mode=1 (slower)
System halted
```
Not sure how to enable pb_trace_mode=1 though...

jthecodebuilder:
On a real Xbox, revision v1.0.

thrimbor:
I tried to dig into this a bit in the last few hours.

First of all, this sample seems to run fine on a debug BIOS loaded via PBL, but it crashes on the retail BIOS of my 1.0 PAL Xbox.

When using #85 to improve debug output and `#define DBG` to set `pb_trace_mode=1`, this comes out:
![img](https://user-images.githubusercontent.com/1339483/52969317-30f2f400-33b0-11e9-965e-7afc0821a040.jpg)

The error is triggered by `draw_arrays()`:
```
/* Send draw commands for the triangles */
static void draw_arrays(unsigned int mode, int start, int count)
{
    uint32_t *p = pb_begin();
    pb_push1(p, NV097_SET_BEGIN_END, mode); p += 2;

    pb_push(p++,0x40000000|NV097_DRAW_ARRAYS,1); //bit 30 means all params go to same register 0x1810
    *(p++) = MASK(NV097_DRAW_ARRAYS_COUNT, (count-1)) | MASK(NV097_DRAW_ARRAYS_START_INDEX, start);

    pb_push1(p,NV097_SET_BEGIN_END, NV097_SET_BEGIN_END_OP_END); p += 2;
    pb_end(p);
}
```

The register that the error message mentions (0x1810) is NV097_DRAW_ARRAYS, the following `pb_push1()` seems to trigger the error. p = 2213917156 and pb_PushNext = 2213917148 when the error message is generated.

[...]

PR #85, opened 2019-01-13
Replace custom C standard library with PDCLib

### LINKED ISSUE #13: nxdk has no proper libc

### LINKED ISSUE #81: xboxrt: Legacy wcsncpy and strncpy functions do not fill exactly n bytes
Legacy wcsncpy (and strncpy) functions in xboxrt do not match the definition of wcsncpy. Specifically, we do not fill exactly n bytes of data. A similar issue exists with strncpy. See the man page for more information:

Exactly n wide characters are written at dest. If the length wcslen(src) is smaller than n, the remaining wide characters in the array pointed to by dest are filled with null wide characters.

From @JayFoxRox 
We currently don't fill exactly n bytes (and neither could we, because our users don't respect that - see wcscpy which calls with SIZE_MAX)

*Edited by JayFoxRox: `wcspy` should not be affected by most of this (except that it's using `wcsncpy` poorly); However, `strncpy` is also affected by the exact same issue (in a different part of the code).*

thrimbor:
**I consider this to be ready for merging now.**

**The following issues are influenced by this:**
Closes #81 
Closes #13
Should help debug #41 by improving printf format specifiers

This replaces most of our custom C standard library code (meaning xboxrt, which afaik was based on lib43, and xlibc, which mostly based on PDCLib already) with an nxdk-port of PDCLib, which includes a slightly improved version of #74 and my own (incomplete) math.h. I chose PDCLib over newlib due to the permissive licensing as well as the simplicity of including #74.

While PDCLib does have a testsuite, it's currently not possible to run it on the Xbox, so help in testing the Xbox-specific parts is very much appreciated. I already ran all the samples built against this branch, and they work fine. The threading code is based on #74 ~~so it should be fine, but I will re-test soon to be sure~~, it has been re-tested and seems to work fine. ~~The main thing left to test thoroughly would be file I/O~~ File I/O also had some testcode run.

The respective submodule branch is [here](https://github.com/XboxDev/nxdk-pdclib/tree/nxdk).

List of the tests I ran so far:
- [x] All of the included nxdk samples
- [x] The tests written for #74
  - [x] Threads
      - [x] [thrd_create](https://gist.github.com/thrimbor/3b34f1a6bd1d9c874ad9cc499c9f10b9)
      - [x] [thrd_current](https://gist.github.com/thrimbor/8054503d3894ddab36a1fa548a12898f)
      - [x] thrd_detach (Not really testable, but should work)
      - [x] [thrd_sleep](https://gist.github.com/thrimbor/23285c4f9ccc0900cf9d41f87b3589a2)
      - [x] thrd_yield (Not really testable, but simply calls NtYieldExecution)
      - [x] [thrd_equal](https://gist.github.com/thrimbor/8054503d3894ddab36a1fa548a12898f)
      - [x] [thrd_join](https://gist.github.com/thrimbor/3b34f1a6bd1d9c874ad9cc499c9f10b9)
      - [x] [thrd_exit](https://gist.github.com/thrimbor/ec6a8b0e7b1f7d6846fe4c72fb2f6186)
  - [x] [thread_local](https://gist.github.com/thrimbor/f45184e6fade24fdf0f121d04c1cdd8e) (static TLS)
  - [x] [call_once](https://gist.github.com/thrimbor/e133d15a2be1ce60fd62592d3b63aa28)
  - [x] Mutexes
      - [x] mtx_destroy (not really testable, but should work)
      - [x] [mtx_init](https://gist.github.com/thrimbor/84bb003cc9b2b3f5f9a5b7d37e0682f9)
      - [x] [mtx_lock](https://gist.github.com/thrimbor/84bb003cc9b2b3f5f9a5b7d37e0682f9)
      - [x] [mtx_timedlock](https://gist.github.com/thrimbor/7247bd1d57a503507412c81f2fdcb6d2)
      - [x] [mtx_trylock](https://gist.github.com/thrimbor/8bc90eabd0968ee6858d36c14a101cd1)
      - [x] [mtx_unlock](https://gist.github.com/thrimbor/647fd2febedcda19da809f80c6afbeb3)
  - [x] TSS
      - [x] [tss_create](https://gist.github.com/thrimbor/7d4654271c5e632d4f73d1dc0d072cc7)
      - [x] [tss_delete](https://gist.github.com/thrimbor/f60d5cf39e6e3ea7778bd96b44a94d7e)
      - [x] [tss_get](https://gist.github.com/thrimbor/c41c174dee1abf41b915eb2f4b90b58a)
      - [x] [tss_set](https://gist.github.com/thrimbor/c41c174dee1abf41b915eb2f4b90b58a)
  - [x] Condition variables
      - [x] [cnd_broadcast](https://gist.github.com/thrimbor/74a6d1a26eea30a52e964fbd0d72d2b2)
      - [x] cnd_destroy (Not really testable, but should work)
      - [x] [cnd_init](https://gist.github.com/thrimbor/59ff937d3997d04c829c1eb874b79985)
      - [x] [cnd_signal](https://gist.github.com/thrimbor/d0e28760a43eb979ca97dc1c8768a963)
      - [x] [cnd_timedwait](https://gist.github.com/thrimbor/59ff937d3997d04c829c1eb874b79985)
      - [x] [cnd_wait](https://gist.github.com/thrimbor/74a6d1a26eea30a52e964fbd0d72d2b2) (Forwards to cnd_timedwait internally)
  - [x] Time
      - [x] [timespec_get](https://gist.github.com/thrimbor/7247bd1d57a503507412c81f2fdcb6d2)
- [x] [Some file I/O tests](https://gist.github.com/thrimbor/d2b1babe8b582e48304773cf5c2efcf7)


samples/0ldskoo1/graphics.h
```diff
@@ -1,9 +1,11 @@
 #ifndef _GRAPHICS_H_
 #define _GRAPHICS_H_
 
+#include <stdint.h>
+
 #define USE_V_BATCHES //first speed optimization, enqueue a vertex batch order (like dma-tag 'ref' on ps2)
 
-//#define USE_I_BATCHES //technic simulation, enqueue index batch order (saves a lots of memory, just simulated here) 
+//#define USE_I_BATCHES //technic simulation, enqueue index batch order (saves a lots of memory, just simulated here)
```
JayFoxRox: Is this an accidental change?

thrimbor:
I updated the above post with a list of all the testcode I ran and included fixes for the problems I encountered. I'm now waiting for feedback on the PR and if/when/how the submodule should be moved to XboxDev.

JayFoxRox:
I'm still not sure how to go forward with this:

> which includes a slightly improved version of #74 and my own (incomplete) math.h

Noteworthy that @thrimbor said on XboxDev Discord, that the math.h should only have to be good enough for now, but is not necessarily expected to stay for very long, as we might have a more complete one.

Fine with me, but the math.h situation (current state / future plans) should be documented.

> I chose PDCLib over newlib due to the permissive licensing

I'm still a bit torn on this, because newlib is more complete and used by similar projects (for Nintendo and Sony consoles) + Android.

However, PDCLib is fine with me, and *this PR is much better* than the hacky PDCLib and custom libc we already have in master.

> I chose PDCLib over newlib due to [...] the simplicity of including #74.

Apparently we didn't document it publicly, but on XboxDev Discord, there was some debate over the usefulness of #74, because C11 threads has got little support by most groups, so it's weird to push it in nxdk (without an alternative like Win32 or pthreads, which have acceptable C11 thread wrappers).

However, as @thrimbor is the main stakeholder and the only one who actually offers solutions, I'm also fine with this.

---

If anybody doesn't like some components here (math.h, PDCLib, C11 threads), they can propose a different solution and we can always replace and switch components later.

However, @thrimbor also has to prepare that *if* a better alternative arrives, we might drop some of his components, because we can't easily maintain it all, and it's unclean design to offer 3 libc variants or competing threading libs. The code would have to be moved to projects using nxdk then.

---

I did not test or review this PR, but the submodule has so few changes compared to upstream now, that it's very acceptable. I skimmed over the changes and it looks sane. I also trust @thrimbor to accept the submodule without extensive review or testing.

JayFoxRox:
I've noticed that you moved the submodule: please rename the repository to `nxdk-pdclib` to match with the other `nxdk-` prefixed repo(s) (such as `nxdk-sdl`)

.gitmodules
```diff
@@ -8,3 +8,6 @@
 	path = lib/sdl/SDL2
 	url = https://github.com/XboxDev/nxdk-sdl.git
 	branch = nxdk
+[submodule "lib/pdclib"]
+	path = lib/pdclib
+	url = https://github.com/XboxDev/pdclib-nxdk.git
```
JayFoxRox: please move to https://github.com/XboxDev/nxdk-pdclib.git to match the other(s).

thrimbor:
I updated the PR to the new URL of the pdclib-repo (sorry for forgetting to leave a comment when I did the actual move).

.gitmodules
JayFoxRox: I've ended up moving it, after @thrimbor contacted me that he didn't have permissions to access the "Settings" panel on GitHub.

.gitmodules
```diff
@@ -8,3 +8,6 @@
 	path = lib/sdl/SDL2
 	url = https://github.com/XboxDev/nxdk-sdl.git
 	branch = nxdk
+[submodule "lib/pdclib"]
+	path = lib/pdclib
+	url = https://github.com/XboxDev/nxdk-pdclib.git
```
JayFoxRox: Wouldn't this check-out the wrong branch?

thrimbor: git keeps a reference to the commit checked out in the submodule, so it still works. Specifying the branch seems to be helpful when updating the submodule, so I'll add it later tonight.

JayFoxRox:
Before I start my review in the next days, can you skim over our issues and make sure to add a "Closes #<issue>" for each affected issue in the first post. Also leave notes about affected issues which don't get closed (either by mentioning them here, or by commenting on them).

JayFoxRox:
I compiled "hello" and "0ldskoo1" as a test and both still compile fine.
I've also skimmed over the code in this repository. I trust the changes in the nxdk-pdclib repository, too, although I did not review it again.

---

An issue I've noticed is that the compilation takes much longer. Due to some problems with the build system, I have a couple of projects which quickly fail when not using `make -B`; so we need to improve that in the future.

The math.h situation should be documented, so please create an issue about the state of math.h and future plans for it @thrimbor.

This has marinated long enough. Sorry for all the delays.

PR #86, opened 2019-01-18
Properly set XBE stack size

thrimbor:
Currently, cxbe always sets dwPeStackCommit to 64KiB (which is the default value of Microsoft's XDK), while our intermediate PE files have m_sizeof_stack_commit set to 4KiB and m_sizeof_stack_reserve to 1MiB. This didn't cause us any problems yet afaik, but this commit fixes it anyway.

I already briefly talked about this on Discord last October:

> I think I figured it out
> PE has SizeOfStackCommit and SizeOfStackReserve. SizeOfStackReserve is the maximum stack size, while SizeOfStackCommit is the initial stack size, meaning only SizeOfStackCommit bytes get reserved in the beginning, and the rest is only mapped when required, at least on Windows.
> But the Xbox doesn't allow growing the stack dynamically, so it only has SizeOfStackCommit. So for the Xbox, it's XBE.SizeOfStackCommit = PE.SizeOfStackReserve
> The comment implies they tried to do XBE.SizeOfStackCommit = PE.SizeOfStackCommit, which would lead to an incorrectly small stack and cause problems
> So their solution was to just bump up the stack size to 64KiB to get rid of the problems
> Btw the XBE documentation by caustik makes the same mistake

I verified that the stack size now properly carries over to the XBE and ran the samples on my Xbox to make sure nothing broke.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,10 @@ ifeq ($(OUTPUT_DIR),)
 OUTPUT_DIR = bin
 endif
 
+ifeq ($(NXDK_STACKSIZE),)
+NXDK_STACKSIZE = 65536
+endif
+
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
 LD           = lld -flavor link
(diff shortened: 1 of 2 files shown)


JayFoxRox:
Shouldn't this have a companion change to reduce the default nxdk stacksize?

Even as-is, most of my larger programs ran out of memory.

I worry that clang will eventually have ~8MB of stack and so we constantly waste 8/64MB for the stack.
We could also have an issue to document this problem instead.

Other than that I'm fine with this change.

thrimbor:
Right, especially since #85 spawns a new thread to execute main() to be able to properly set up TLS support, so even simple code reserves 2MiB already. I'll come up with something that sets a lower default size but allows overriding (probably later today). Do you want it to go into this PR?

JayFoxRox:
Yes, I think that'd make most sense.

thrimbor:
I added a commit which sets the default stack size to 64KiB, which is what cxbe had hardcoded before and which is afaik also used by the XDK. It also allows overriding the default value by using e.g. `NXDK_STACKSIZE = 4096` in the Makefile of a sample (I just quickly checked that it actually works by using `objdump -x`).

PR #87, opened 2019-04-05
Add (rudimentary!) C++ support

thrimbor:
This adds basic C++ support by adding rules to the Makefile, adding `extern "C"` where necessary to get the name mangling right and renaming struct fields in the xboxkrnl-header.

Global constructors are already handled by my PDCLib-additions, so when this PR is combined with https://github.com/XboxDev/nxdk-pdclib/pull/2, simple code like [this](https://gist.github.com/thrimbor/62a550b36cb5ab65a7b6ceacc4fbd084) already works.

Note that CI on macOS currently fails due to the problem described in https://github.com/XboxDev/nxdk-pdclib/pull/1, this affects Arch Linux now, too. It works on the Ubuntu CI, though.

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -895,12 +895,12 @@ typedef struct _KQUEUE
     LIST_ENTRY ThreadListHead;
 } KQUEUE, *PKQUEUE, *RESTRICTED_POINTER PRKQUEUE;
 
-struct KTHREAD;
+struct _KTHREAD;
 
 typedef struct _KWAIT_BLOCK
 {
     LIST_ENTRY WaitListEntry;
-    struct KTHREAD *Thread;
+    struct _KTHREAD *Thread;
     PVOID Object;
     struct _KWAIT_BLOCK *NextWaitBlock;
     SHORT WaitKey;
@@ -912,7 +912,7 @@ typedef struct _KAPC
     SHORT Type;
     CHAR ApcMode;
     UCHAR Inserted;
-    struct KTHREAD *Thread;
+    struct _KTHREAD *Thread;
     LIST_ENTRY ApcListEntry;
     PVOID KernelRoutine;
     PVOID RundownRoutine;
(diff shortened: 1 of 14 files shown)


JayFoxRox:
Thanks! I only skimmed over the code and it looks fine. 

---

Could you please add a hello++ sample or something? I don't think anybody else will use it, if we don't provide one.

A sample should contain at least one object with constructor / destructor to show that that's working. Possibly a global object so we can show initializers / constructors being called by CRT.

*The sample could be added as a follow-up (but ideally it would be here from the beginning).*

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -895,12 +895,12 @@ typedef struct _KQUEUE
     LIST_ENTRY ThreadListHead;
 } KQUEUE, *PKQUEUE, *RESTRICTED_POINTER PRKQUEUE;
 
-struct KTHREAD;
+struct _KTHREAD;
```
mborgerson: Why is this rename required?

dracc: Struct above and below follow the same naming scheme. :+1:

JayFoxRox: In C++ `struct KTHREAD` declares a type called `KTHREAD`; whereas in C you'd still have to use `struct KTHREAD` to refer to the same thing (not sure about terminology here, but I believe you'd "prototype" the struct instead of "declaring" it?).

Hence you'd typically prefix the tag in C and use a typedef like `typedef struct _KTHREAD { ... } KTHREAD;`, so both, C and C++ can just use `KTHREAD` when referring to the same type (instead of C++ using `KTHREAD` but C using `struct KTHREAD`).

mborgerson: Ah yes. Struct identifiers should have underscore prefix. Type definitions via typedef should not. I was asking if there was a functional reason for this change, not just convention. IIRC C++ should allow the names to be the same.

> whereas in C you'd still have to use struct KTHREAD to refer to the same thing
You should be able to use the typedef in both C in C++ ofc.

mborgerson:
This looks good. Since we've had multiple reviewers approve, I'll merge it in. Thanks!

PR #88, opened 2019-04-05
Add hello++ cpp sample

dracc:
A small sample showing off a working class structure with a custom constructor.

### DIFF
--- a/samples/hello++/Makefile
+++ b/samples/hello++/Makefile
@@ -1,0 +1,5 @@
+XBE_TITLE=hello++
+GEN_XISO = $(XBE_TITLE).iso
+SRCS = $(CURDIR)/main.cpp
+NXDK_DIR = $(CURDIR)/../..
+include $(NXDK_DIR)/Makefile
--- a/samples/hello++/main.cpp
+++ b/samples/hello++/main.cpp
@@ -1,0 +1,35 @@
+#include <xboxrt/debug.h>
+#include <pbkit/pbkit.h>
+#include <hal/xbox.h>
+
+class NumberClass{
+private:
+  int hiddenValue = 0;
+public:
+  NumberClass() {hiddenValue = 5;};
+  NumberClass(int v) : hiddenValue(v) {};
+  int getValue() { return hiddenValue; };
+};
+
+int main() {
+  int ret = pb_init();
+  if (ret != 0) {
+    XSleep(2000);
+    return -1;
+  }
+
+  pb_show_debug_screen();
+
+  NumberClass defaultFive;
+  NumberClass customSix(6);
+  
+  while (true) {
+    debugPrint("Value of defaultFive: %i\n", defaultFive.getValue());
+    debugPrint("Value of customSix  : %i\n", customSix.getValue());
+    XSleep(2000);
+  }
+
+  pb_kill();
+
+  return 0;
+}


samples/hello++/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+  default:
+    XSleep(2000);
+    XReboot();
+    return -1;
+  }
+
+  pb_show_debug_screen();
+
+  hello h(6);
+  hello i;
```
JayFoxRox: I suggest calling them `defaultFive` and `customSix` or something. The classname could be `NumberClass`

samples/hello++/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+int main() {
+  switch(pb_init())
+  {
+  case 0: break;
+  default:
+    XSleep(2000);
+    XReboot();
+    return -1;
+  }
```
JayFoxRox: I always cringe with this init sequence; I suggest:

```
  int ret = pb_init();
  if (ret != 0) {
    XSleep(2000);
    XReboot();
    return -1;
  }
```

*And possibly a PR to change it in our other code, too*

samples/hello++/main.cpp
```diff
@@ -0,0 +1,40 @@
...
+  hello i;
+  
+  while(1) {
+    debugPrint("Value of h: %i\n", h.getValue());
+    debugPrint("Value of i: %i\n", i.getValue());
+    XSleep(2000);
+  }
+
+  pb_kill();
+  XReboot();
```
JayFoxRox: I believe nxdk reboots automatically when returning from `main`. I consider this the equivalent of `exit(0); return 0;`.

We should fix this in all of our samples.

samples/hello++/Makefile
```diff
@@ -0,0 +1,4 @@
+XBE_TITLE=hello++
+SRCS = $(wildcard $(CURDIR)/*.cpp)
```
JayFoxRox: Consider listing source files explicitly; we should do this in all our samples.

Wildcards are error prone.

samples/hello++/main.cpp
dracc: I tested it in XQEMU, returning simply restarts the program, `XReboot()` crashes with something nv2a related. I'd argue they're not interchangeable.
Alas, I can't test this on real hardware currently.

samples/hello++/main.cpp
dracc: Taken care of in this instance.
I'll file an issue about the other samples.

samples/hello++/main.cpp
JayFoxRox: We should fix this then. If a program returns from main, that should behave like an "exit" (including `atexit()` callbacks / destructors) and I'd argue that the strategy on Xbox to "exit" is to:

- Reboot to dashboard
- Reboot to the next program (Dashboard, or DVD if one is inserted)

I'm not sure which of these XReboot does, but it's safe to assume it should be one of them.
Eitherway, we shouldn't need XReboot in our sample code.

samples/hello++/main.cpp
```diff
@@ -0,0 +1,35 @@
...
+    XSleep(2000);
+    return -1;
+  }
+
+  pb_show_debug_screen();
+
+  NumberClass defaultFive;
+  NumberClass customSix(6);
+  
+  while(1) {
```
JayFoxRox: nit: `while(true)`

samples/hello++/Makefile
Ernegien: It also may be worth including `GEN_XISO = $(XBE_TITLE).iso` in this sample as well as some of the others that are missing it.

samples/hello++/main.cpp
Ernegien: Agreed, this should be looked into further, but since it affects all samples perhaps this would be better served as a separate issue?

samples/hello++/main.cpp
```diff
@@ -0,0 +1,35 @@
...
+    XSleep(2000);
+    return -1;
+  }
+
+  pb_show_debug_screen();
+
+  NumberClass defaultFive;
+  NumberClass customSix(6);
+  
+  while (true) {
```
JayFoxRox: This requires `#include <stdbool.h>` (or `#include <cstdbool>` eventually). We should fix it when doing a cleanup pass on all samples.

It currently depends on one of the other headers including it.

PR #89, opened 2019-04-05
SDL_ttf support

dracc:
The real work was made by @mborgerson and @thrimbor, I merely created the sample.

Font is CC0 from http://dotcolon.net/font/vegur/

Edit: I forgot to mention that xboxdev/nxdk-sdl#1 is required for this to work.

screenshot


samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
+/*
+  Copyright (C) 1997-2016 Sam Lantinga <slouken@libsdl.org>
```
JayFoxRox: Is this valid? There's at least some changes for nxdk (would possiby need someone elses copyright head) and I'm not sure how much is actually taken from SDL (maybe we can remove the entire copyright head).

I don't know how the copyright head looks for other samples, but I feel like there shouldn't be one, or it should be CC0 or something to signal intend (so people don't copy copyright headers from some sample into every program).

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+#include <hal/video.h>
+#include "stdio.h"
+#include "string.h"
+#include <SDL.h>
+#include <SDL_ttf.h>
+
+int _exit(int x)
+{
+    debugPrint("_exit called\n");
+    debugPrint( " SDL_Error: %s\n", SDL_GetError() );
```
JayFoxRox: This feels more like an `_fatal_error` than `_exit`.

Also, why is it `_` prefixed?

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+const extern int SCREEN_WIDTH;
+const extern int SCREEN_HEIGHT;
+#define NUM_SPRITES 10
+#define MAX_SPEED   5
+
+static SDL_Rect *positions, *velocities;
+
+/* Call this instead of exit(), so we can clean up SDL: atexit() is evil. */
+static void
+quit(int rc)
```
JayFoxRox: `quit` and `exit` sound too much alike in my opinion.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+    driver = SDL_GetCurrentVideoDriver();
+
+    window = SDL_CreateWindow("Demo",
+        SDL_WINDOWPOS_UNDEFINED,
+        SDL_WINDOWPOS_UNDEFINED,
+        SCREEN_WIDTH, SCREEN_HEIGHT,
+        SDL_WINDOW_SHOWN);
+    if(window == NULL)
+    {
+        debugPrint( "Window could not be created! SDL_Error: %s\n", SDL_GetError() );
```
JayFoxRox: Won't exit handle printing this error?

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+        _exit(1);
+    }
+
+    /* Create the renderer */
+    renderer = SDL_CreateRenderer(window, -1, 0);
+    if (!renderer) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't create renderer: %s\n", SDL_GetError());
```
JayFoxRox: Same as above; `_exit` seems to handle printing SDL errors.

So it should probably be extended to accept another string parameter to describe which function caused the error (if it's not already part of `SDL_GetError`)

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+#include <xboxrt/debug.h>
+#include <pbkit/pbkit.h>
+#include <hal/xbox.h>
+#include <hal/video.h>
+#include "stdio.h"
+#include "string.h"
+#include <SDL.h>
+#include <SDL_ttf.h>
+
+int _exit(int x)
```
JayFoxRox: Why is there an unused paramter here?

samples/sdl_ttf/main.c
JayFoxRox: This `rc` parameter remains unused.

Whenever this or `_exit` are being called, there are random numbers in the code - which is confusing to new users.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+/* Call this instead of exit(), so we can clean up SDL: atexit() is evil. */
+static void
+quit(int rc)
+{
+    SDL_VideoQuit();
+    _exit(rc);
+}
+
+SDL_Texture *
+LoadSprite(SDL_Renderer *renderer)
```
JayFoxRox: This sounds more like "CreateTextureFromFont" (there's no "SDL_Sprite" involved, and it doesn't "Load" anything from disk except a font - but that load is implied. The task this does, is creation of new content using parameters)

For reusability, it shoud probably be more like: `CreateTextureFromFont(const char* font_path, unsigned int font_size, SDL_Color font_color, const char* text, ...);`

samples/sdl_ttf/main.c
dracc: It probably needs some modification, sample is just a slightly modified version of the already available sdl one.
I'll be the first to admit that I don't feel comfortable when it comes to licenses, I'm not sure what modification should be done for this to be valid.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+{
+    SDL_Surface *temp;
+    SDL_Texture *sprite;
```
JayFoxRox: `sprite` &rarr; `texture`

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+    /* Create textures from the image */
+    sprite = SDL_CreateTextureFromSurface(renderer, temp);
+    if (sprite == NULL) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't create texture: %s\n", SDL_GetError());
+        SDL_FreeSurface(temp);
+        return NULL;
+    }
+    SDL_FreeSurface(temp);
+
+    /* We're ready to roll. :) */
```
JayFoxRox: Comment doesn't add anything

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+        position->x += velocity->x;
+        if ((position->x < 0) || (position->x >= (viewport.w - sprite_w))) {
+            velocity->x = -velocity->x;
+            position->x += velocity->x;
+        }
+        position->y += velocity->y;
+        if ((position->y < 0) || (position->y >= (viewport.h - sprite_h))) {
+            velocity->y = -velocity->y;
+            position->y += velocity->y;
+        }
```
JayFoxRox: All of this logic just masquerades what this sample does.

If I look at an TTF sample, all I really care about how I can integrate text into my program as quickly as possible.
The code we supply should be re-usable so people can just open a sample and copy/paste everything they need (so we should reduce amount of garbage between useful parts), and strip it down.

I specifically don't care about a physics simulation.

- This should be removed.

If we do anything else, other than trivial text rendering, we should probably show various variations: italic, outlines, auto-linebreaks, alignments in a textbox, ... (although it's not critical for now).

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+    /* Enable standard application logging */
+    SDL_LogSetPriority(SDL_LOG_CATEGORY_APPLICATION, SDL_LOG_PRIORITY_INFO);
+
+    if (SDL_VideoInit(NULL) < 0) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't initialize SDL video: %s\n",
+                SDL_GetError());
+        _exit(1);
+    }
+
+    if (TTF_Init() < 0) {
```
JayFoxRox: There should be a comment here or it should be part of a block which deals with font-texture creation, so people don't forget to copy it into their projects.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+    /* Create the renderer */
+    renderer = SDL_CreateRenderer(window, -1, 0);
+    if (!renderer) {
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Couldn't create renderer: %s\n", SDL_GetError());
+        _exit(5);
+    }
+
+    /* Clear the window, load the sprite and go! */
+    SDL_SetRenderDrawColor(renderer, 0xA0, 0xA0, 0xA0, 0xFF);
+    SDL_RenderClear(renderer);
```
JayFoxRox: This block of code should probably be removed?
It's done in the renderer loop anyay; and when this is called, the program is still loading resources and doesn't even draw (I think). The code currently reads at `init init init *render init init loop: render`; so this `*render` block feels out of place.

So by the time it renders, this will have repeated anyway.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+        return NULL;
+    }
+    SDL_FreeSurface(temp);
+
+    /* We're ready to roll. :) */
+    return sprite;
+}
+
+void
+MoveSprites(SDL_Renderer * renderer, SDL_Texture * sprite)
```
JayFoxRox: This function doesn't just move sprites (also see my other comments about removing "sprites" altogether): It handles rendering and displaying the entire frame.

If the physics are removed, I feel like that this could just be inlined into the mainloop.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+    done = 0;
+    while (!done) {
+        /* Check for events */
+        while (SDL_PollEvent(&event)) {
+            switch (event.type) {
+            case SDL_WINDOWEVENT:
+                switch (event.window.event) {
+                case SDL_WINDOWEVENT_EXPOSED:
+                    SDL_SetRenderDrawColor(renderer, 0xA0, 0xA0, 0xA0, 0xFF);
+                    SDL_RenderClear(renderer);
```
JayFoxRox: This feels weird `MoveSprites` will clear on every frame anyway (with another color).

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+        }
+        MoveSprites(renderer, sprite);
+    }
+
+    quit(0);
+}
+
+void main(void)
+{
+    if (pb_init() != 0)
```
JayFoxRox: Calling in an `if` is a bit frowned upon (because if you disable the check, you might accidentally break code). Hence my original version with a temporary `ret` variable.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+    {
+        XSleep(2000);
+        XReboot();
+        return;
+    }
+
+    pb_show_debug_screen();
+    demo();
```
JayFoxRox: Not sure where this name comes from, but I'd probably cal it `sample()` - because that's what it is.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+    pb_kill();
+    XReboot();
```
JayFoxRox: We had discussed this reboot on Discord and other issues: It shouldn't be necessary.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+#include "string.h"
+#include <SDL.h>
+#include <SDL_ttf.h>
+
+int _exit(int x)
+{
+    debugPrint("_exit called\n");
+    debugPrint( " SDL_Error: %s\n", SDL_GetError() );
+    XSleep(5000);
+    XReboot();
```
JayFoxRox: This function never calls `pb_kill` etc.

Consider a `cleanup()` function.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+#define NUM_SPRITES 10
+#define MAX_SPEED   5
+
+static SDL_Rect *positions, *velocities;
+
+/* Call this instead of exit(), so we can clean up SDL: atexit() is evil. */
+static void
+quit(int rc)
+{
+    SDL_VideoQuit();
```
JayFoxRox: Could this function modify `SDL_GetError()`?

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,241 @@
...
+        positions[i].y = rand()*(i+1) % (window_h - sprite_h);
+        positions[i].w = sprite_w;
+        positions[i].h = sprite_h;
+        velocities[i].x = 0;
+        velocities[i].y = 0;
+        while (!velocities[i].x && !velocities[i].y) {
+            velocities[i].x = (rand()*(i+1) % (MAX_SPEED * 2 + 1)) - MAX_SPEED;
+            velocities[i].y = (rand()*(i+1) % (MAX_SPEED * 2 + 1)) - MAX_SPEED;
+        }
+    }
```
JayFoxRox: See other comments: Too much non-TTF code. A single sprite with static position will suffice.

samples/sdl_ttf/main.c
dracc: SDL seems to be the culprit here. When `_exit` was removed a compilation error occured, "symbol __exit not found" or something to that effect.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,157 @@
...
+    XSleep(2000);
+    SDL_DestroyRenderer(renderer);
+    SDL_DestroyWindow(window);
+    cleanup(init_level);
+    return;
+  }
+
+  SDL_Rect outputPosition;
+  SDL_QueryTexture(texture, NULL, NULL, &outputPosition.w, &outputPosition.h);
+  SDL_SetRenderDrawColor(renderer, 0x00, 0x00, 0x00, 0xFF);
```
JayFoxRox: I'd move this into the renderer loop; if someone adds other draw calls this would otherwise break.

It's also better to put related code together. I'd also assume all of these calls are very cheap (performance wise).

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,157 @@
...
+  }
+  ++init_level;
+
+  SDL_Window   *window;
+  SDL_Renderer *renderer;
+  SDL_Surface  *surface;
+  SDL_Texture  *texture;
+  TTF_Font     *font;
+
+  SDL_Color     textcolor = {0x7F, 0xFF, 0x7F, 0xFF};
```
JayFoxRox: I'd move this to where it's actually used.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,157 @@
...
+  }
+  if (lvl >= 1) {
+    pb_kill();
+  }
+}
+
+
+void main() {
+  const extern int SCREEN_WIDTH;
+  const extern int SCREEN_HEIGHT;
```
JayFoxRox: `extern` variables should be on global scope in my opinion.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,157 @@
...
+  done = 0;
+  while (!done) {
+    /* Check for events */
+    while (SDL_PollEvent(&event)) {
+      switch (event.type) {
+      case SDL_WINDOWEVENT:
+        switch (event.window.event) {
+        case SDL_WINDOWEVENT_EXPOSED:
+          SDL_SetRenderDrawColor(renderer, 0xA0, 0xA0, 0xA0, 0xFF);
+          SDL_RenderClear(renderer);
```
JayFoxRox: Still shouldn't be necessary, it's done a couple lines below.

samples/sdl_ttf/main.c
JayFoxRox: This sounds like a bug in our SDL then?

Someone should look into this and send a separate PR. We shouldn't have to sacrifice our sample code quality for bugs elsewhere.

dracc: Quite possibly.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,145 @@
...
+    }
+    SDL_SetRenderDrawColor(renderer, 0x00, 0x00, 0x00, 0xFF);
+    SDL_RenderClear(renderer);
+    SDL_RenderCopy(renderer, texture, NULL, &outputPosition);
+    SDL_RenderPresent(renderer);
+  }
+
+  SDL_DestroyTexture(texture);
+cleanup:
+  SDL_DestroyRenderer(renderer);
```
JayFoxRox: `renderer` might be uninitialized here.
You should probably declare all 3 of these with NULL initialization, and then check for non-NULL here.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,145 @@
...
+  window = SDL_CreateWindow("NXDK SDL_ttf sample",
+                            SDL_WINDOWPOS_UNDEFINED,
+                            SDL_WINDOWPOS_UNDEFINED,
+                            SCREEN_WIDTH, SCREEN_HEIGHT,
+                            SDL_WINDOW_SHOWN);
+  if (window == NULL) {
+    debugPrint("Window creation failed: %s", SDL_GetError());
+    XSleep(2000);
+    goto cleanup;
+    return;
```
JayFoxRox: These returns are no longer necessary. The `goto` is enough.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,145 @@
...
+
+const extern int SCREEN_WIDTH;
+const extern int SCREEN_HEIGHT;
+
+// SDL requires this symbol to exist
+void _exit() {
+}
+
+void main() {
+  int init_ret;
```
JayFoxRox: POSIX style would just be `int ret` or `int r`.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,145 @@
...
+  initialized_SDL = true;
+
+  init_ret = TTF_Init();
+  if (init_ret != 0) {
+    debugPrint("TTF_Init failed: %s", TTF_GetError());
+    XSleep(2000);
+    goto cleanup;
+    return;
+  }
+  initialized_TTF = true;
```
JayFoxRox: I just realized you could just use `ttf_initialization = TTF_Init()` and check for 0 in cleanup.
You'd declare it as `int ttf_initialization = -1;` or something to signal that it wasn't initialized yet.

dracc: That's a very good idea!

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  SDL_Rect outputPosition;
+  int window_height, window_width;
+  SDL_GetWindowSize(window, &window_width, &window_height);
+  SDL_QueryTexture(texture, NULL, NULL, &outputPosition.w, &outputPosition.h);
+  outputPosition.x = window_width/2 - outputPosition.w / 2;
+  outputPosition.y = window_height/2 - outputPosition.h / 2;
+
+  int done;
```
JayFoxRox: Make this `bool running = true;` or something and move it closer to loop.

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,145 @@
...
+  SDL_Event event;
```
JayFoxRox: Can be moved into the loop where it's actually used (near event check - so the reader sees which types are used).

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,145 @@
...
+    goto cleanup;
+    return;
+  }
+
+  SDL_Rect outputPosition;
+  int window_height, window_width;
+  SDL_GetWindowSize(window, &window_width, &window_height);
+  SDL_QueryTexture(texture, NULL, NULL, &outputPosition.w, &outputPosition.h);
+  outputPosition.x = window_width/2 - outputPosition.w / 2;
+  outputPosition.y = window_height/2 - outputPosition.h / 2;
```
JayFoxRox: nit: `x/2` &rarr; `x / 2`

samples/sdl_ttf/main.c
```diff
@@ -0,0 +1,134 @@
...
+
+const extern int SCREEN_WIDTH;
+const extern int SCREEN_HEIGHT;
+
+// SDL requires this symbol to exist
+void _exit() {
+}
+
+void main() {
+  int ret;
```
JayFoxRox: No longer used.

samples/sdl_ttf/main.c
dracc: I'd prefer not to re-initialize it on every loop though?

JayFoxRox: It's not re-initialized (a variable declaration without initialization will mostly be a compile-time operation). The code will likely remain unchanged (variable will have reserved room on stack) - it's just easier to find for readers of the code (and it might be overwritten by other variables if it's out-of-scope; so there's less memory use).

samples/sdl_ttf/main.c
dracc: According to spec, SDL_DestroyRenderer and SDL_DestroyWindow only sets SDL_GetError() when an invalid pointer has been given - should be OK as we don't really care for a return value.

JayFoxRox: To me, it still feels unclean to pass random pointers to functions. If someone was to use this code elsewhere, the uninitialized variable might point at another valid SDL renderer by sheer luck - and it would then destroy an object that was never meant to be destroyed. It's a very theoretical case, but I hope you see my point.

I'd also move the `cleanup:` one line up, so `DestroyTexture` is labeled as cleanup.

dracc: I guess I'm narrow-minded. You make a great point. Will fix!

JayFoxRox:
- Why does this update the extract-xiso submodule? If you do this, it should be at least a separate commit
- This now has to update the nxdk-sdl submodule, since https://github.com/XboxDev/nxdk-sdl/pull/1 was merged

samples/sdl_ttf/Makefile
```diff
@@ -0,0 +1,12 @@
+DEBUG = y
+XBE_TITLE = sdl
+SRCS += $(wildcard $(CURDIR)/*.c)
```
JayFoxRox: Avoid wildcards

samples/sdl_ttf/Makefile
```diff
@@ -0,0 +1,12 @@
...
+NXDK_DIR = $(CURDIR)/../..
+NXDK_SDL = y
```
JayFoxRox: Makefile should do what #95 did, for consistency

samples/sdl_ttf/Makefile
Ernegien: This might be the cause of the following build error I receive in msys.

```[ LD       ] main.exe
make: execvp: lld-link: Argument list too long
make: *** [/home/Mike/projects/nxdk/samples/sdl_ttf/../../Makefile:114: main.exe] Error 127```

JayFoxRox: No, it's (almost certainly) because we list all source files, instead of moving larger groups of them into libraries.
SDL_ttf added a couple of more, and now you are hitting limits (which are very low on Windows).

I'm actually surprised this worked for as long as it did.

You can temporarily work around it by using shorter folder-names.

I suggest you create a separate issue for this problem, as it's not specific to this PR.

dracc:
@JayFoxRox: The commit by @mborgerson used to contain an `extract-xiso` revision change and upon trying to remove that change I incidentially chose the wrong revision myself. Should be fixed by now.

SDL_ttf submodule revision update was made part of the initial commit as it's needed for this changeset to work.

mborgerson:
Nice! Thanks

PR #93, opened 2019-04-10
hal: Fix buffer overflow error in XFindFirstFile/XFindNextFile

thrimbor:
**/edit: Ignore this for now, I plan to submit a PR which replaces all this code with a WinAPI-style implementation.**

This is the issue @dracc encountered and described in Discord.

The issue was that these functions directly used `FILE_DIRECTORY_INFORMATION` and passed its size to `NtQueryDirectoryFile`, which is problematic because `FILE_DIRECTORY_INFORMATION` contains a field of variable length called `FileName` at its end.
It seems that OpenXDK used to hack around this by giving this field a fixed length of 0x100.

I instead solved this by reserving a buffer with a length of `sizeof(FILE_DIRECTORY_INFORMATION) + 1024` and casting it to a `PFILE_DIRECTORY_INFORMATION`. I checked afterwards how ReactOS does this, and they utilize a similar approach but with a larger buffer. Considering the fact that this buffer is placed on the stack, I decided to go with the smaller buffer size.

While being a bit ugly, this solution allows us to keep our Kernel-header close to MS documentation / ReactOS.

### DIFF
--- a/lib/hal/fileio.c
+++ b/lib/hal/fileio.c
@@ -6,6 +6,9 @@
 #include <xboxrt/debug.h>
 #include <stdbool.h>
 
+// Defines the allowed size for a FILE_DIRECTORY_INFORMATION entry
+#define FIND_DATA_SIZE (sizeof(FILE_DIRECTORY_INFORMATION) + 1024)
+
 // #define DEBUG
 
 static char *currentDirString = NULL;
@@ -733,7 +736,8 @@ unsigned int XFindFirstFile(
 #endif
 
 	IO_STATUS_BLOCK IoStatusBlock;
-	FILE_DIRECTORY_INFORMATION FileInformation;
+	BYTE FileInfoRaw[FIND_DATA_SIZE];
+	PFILE_DIRECTORY_INFORMATION FileInformation = (PFILE_DIRECTORY_INFORMATION)FileInfoRaw;
 	HANDLE handle = NULL;
 	ANSI_STRING FileMask;
 	ANSI_STRING FileName;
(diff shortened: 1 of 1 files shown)


lib/hal/fileio.c
```diff
@@ -6,6 +6,9 @@
 #include <xboxrt/debug.h>
 #include <stdbool.h>
 
+// Defines the allowed size for a FILE_DIRECTORY_INFORMATION entry
+#define FIND_DATA_SIZE (sizeof(FILE_DIRECTORY_INFORMATION) + 1024)
```
JayFoxRox: Do we have a `MAX_PATH`? Should we have one?

lib/hal/fileio.c
```diff
@@ -802,34 +806,35 @@ int XFindNextFile(
 #endif
 
 	IO_STATUS_BLOCK IoStatusBlock;
-	FILE_DIRECTORY_INFORMATION FileInformation;
+	BYTE FileInfoRaw[FIND_DATA_SIZE];
+	PFILE_DIRECTORY_INFORMATION FileInformation = (PFILE_DIRECTORY_INFORMATION)FileInfoRaw;
```
JayFoxRox: nit: The standard way of handling this would be a union I think.

JayFoxRox:
This has been sitting idle for a while, and it was supposed to be replaced by a WinAPI option.

Should this PR be closed / rejected; or should it be merged as a temporary measure?

From what I can tell, it's "better than master". I did not test it yet, but did review a while ago.

mborgerson:
Closing for now as it seems like this PR is no longer necessary with #112

PR #99, opened 2019-04-28
hal: Allow xbox paths in XConvertDOSFilenameToXBOX

dracc:
This helper function is useful but when not needed it gets in the way.
By checking if the input filename already is in Xbox format, we might return early with a STATUS_SUCCESS.

This code was originally written by @JayFoxRox for JayFoxRox/nxdk#7

### DIFF
--- a/lib/hal/fileio.c
+++ b/lib/hal/fileio.c
@@ -89,6 +89,12 @@ static char *getCurrentDirString()
  */
 int XConvertDOSFilenameToXBOX(const char *dosFilename, char *xboxFilename)
 {
+	// Allow Xbox filenames directly
+	if (!memcmp(dosFilename, "\\Device\\", 8)) {
+		strcpy(xboxFilename, dosFilename);
+		return STATUS_SUCCESS;
+	}
+
 	// path contains the qualified pathname from the root
 	// directory without the leading slash.  eg. "foo\bar.txt"
 	const char *path;
--- a/lib/hal/fileio.c
+++ b/lib/hal/fileio.c
@@ -75,7 +75,9 @@ static char *getCurrentDirString()
 
 /**
  * Converts a DOS-style path (eg. "c:/foo/bar.txt") to a XBOX-style
- * path (eg. "\Device\Harddisk0\Partition2\foo\bar.txt")
+ * path (eg. "\Device\Harddisk0\Partition2\foo\bar.txt").
+ * Returns early if the path is alread in XBOX-style - will use the
+ * input path as-is.
  *
  * We handle the following scenarios (back/forward slashes are handled
  * in all cases):


lib/hal/fileio.c
```diff
@@ -89,6 +89,12 @@ static char *getCurrentDirString()
  */
 int XConvertDOSFilenameToXBOX(const char *dosFilename, char *xboxFilename)
 {
+	// Allow Xbox filenames directly
+	if (!memcmp(dosFilename, "\\Device\\", 8)) {
+		strcpy(xboxFilename, dosFilename);
+		return STATUS_SUCCESS;
+	}
```
JayFoxRox: This is a pretty dirty hack to be honest: `\\Device\\` probably isn't a valid `DOSFilename` (which the function name implies).

I think it's fine, if we intend to remove `XConvertDOSFilenameToXBOX` in the future, or it might be fine if we wrap `XConvertDOSFilenameToXBOX` into a better named function.

Generally, this function is used for `fopen` etc. and I'm not sure if that should even accept the `\\Device\\` style filenames in the first place.

lib/hal/fileio.c
```diff
@@ -89,6 +89,12 @@ static char *getCurrentDirString()
  */
 int XConvertDOSFilenameToXBOX(const char *dosFilename, char *xboxFilename)
 {
+	// Allow Xbox filenames directly
+	if (!memcmp(dosFilename, "\\Device\\", 8)) {
```
JayFoxRox: The documentation at the function header has to be adapted.

lib/hal/fileio.c
JayFoxRox: @XboxDev thoughts?

Should we reject this altogether? If not, it should be merged.

lib/hal/fileio.c
```diff
@@ -75,7 +75,8 @@ static char *getCurrentDirString()
 
 /**
  * Converts a DOS-style path (eg. "c:/foo/bar.txt") to a XBOX-style
- * path (eg. "\Device\Harddisk0\Partition2\foo\bar.txt")
+ * path (eg. "\Device\Harddisk0\Partition2\foo\bar.txt").
+ * Returns early if the path is alread in XBOX-style.
```
mborgerson: nit: This function doesn't just return early, it will simply use the input string as-is when copying to xboxFilename.

lib/hal/fileio.c
mborgerson: Note: invalid paths may make it through this check (e.g. if I just pass in "\\Device\\"). Though better error checking should happen down the road so I think it's fine for now.

lib/hal/fileio.c
dracc: Fixed, thanks!

lib/hal/fileio.c
mborgerson: I'm also not sure what the use case is for passing in non DOS-style paths. Is there an issue that motivated this PR? Either way, I don't think it's a bad thing to be able to handle both styles.

dracc: Current file IO API assumes DOS style paths and always calls this function leading to errors when Xbox style paths are supplied.

mborgerson: Yes, but my question was: what are the cases when this function is called with Xbox-style paths?

dracc: Only current case I know of is JayFoxRox/xbe-loader but it could probably be modified to use the NT functions directly if we don't like the changes from this PR. :+1:

PR #100, opened 2019-04-28
pktdrv: PktdrvInterrupt and MmioBase as extern instead of macros

dracc:
Get rid of a magic number and give these values proper types instead of just defining them in macros.

Code originally written by @JayFoxRox for JayFoxRox/nxdk#7.

### DIFF
--- a/lib/net/pktdrv/pktdrv.c
+++ b/lib/net/pktdrv/pktdrv.c
@@ -345,9 +345,9 @@ enum {
 
 
 //Register access macros for XBOX
-#define REG(x)	(*((DWORD *)((ULONG_PTR)PktdrvMmioBase+(x))))
-#define REGW(x)	(*((WORD *)((ULONG_PTR)PktdrvMmioBase+(x))))
-#define REGB(x)	(*((BYTE *)((ULONG_PTR)PktdrvMmioBase+(x))))
+#define REG(x)	(*((DWORD volatile *)((ULONG_PTR)PktdrvMmioBase+(x))))
+#define REGW(x)	(*((WORD volatile *)((ULONG_PTR)PktdrvMmioBase+(x))))
+#define REGB(x)	(*((BYTE volatile *)((ULONG_PTR)PktdrvMmioBase+(x))))
 
 
 
--- a/lib/net/pktdrv/pktdrv.c
+++ b/lib/net/pktdrv/pktdrv.c
@@ -42,6 +42,9 @@
 extern unsigned long times(void *);
 
 //temporary dirty interface
+void* PktdrvMmioBase = (void*)0xFEF00000;
+unsigned int PktdrvInterrupt = 4;
+
 extern int something_to_send;
 extern unsigned char *packet_to_send;
 extern unsigned int size_of_packet_to_send;
@@ -342,10 +345,9 @@ enum {
 
 
 //Register access macros for XBOX
-#define	BASE	0xFEF00000
-#define REG(x)	(*((DWORD *)(BASE+(x))))
-#define REGW(x)	(*((WORD *)(BASE+(x))))
-#define REGB(x)	(*((BYTE *)(BASE+(x))))
+#define REG(x)	(*((DWORD *)((ULONG_PTR)PktdrvMmioBase+(x))))
+#define REGW(x)	(*((WORD *)((ULONG_PTR)PktdrvMmioBase+(x))))
+#define REGB(x)	(*((BYTE *)((ULONG_PTR)PktdrvMmioBase+(x))))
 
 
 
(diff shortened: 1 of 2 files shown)


lib/net/pktdrv/pktdrv.h
```diff
@@ -1,6 +1,9 @@
 #ifndef _Pktdrv_
 #define _Pktdrv_
 
+extern void* PktdrvMmioBase;
+extern unsigned int PktdrvInterrupt;
```
JayFoxRox: This is a hack, so xbe-loader can relocate the NIC (which implies changing the MMIO addresses).

If this was only to get rid of the macros, these variables wouldn't have to be exposed outside (they'd be `static const`). I'm a bit torn about this implementation as there's probably better ways to handle this (such as passing an address to `Pktdrv_Init` or something - much like it's being done for kernel function `AvSendTVEncoderOption`, which expects the GPU MMIO address as first argument).

lib/net/pktdrv/pktdrv.c
```diff
@@ -342,10 +345,9 @@ enum {
 
 
 //Register access macros for XBOX
-#define	BASE	0xFEF00000
-#define REG(x)	(*((DWORD *)(BASE+(x))))
-#define REGW(x)	(*((WORD *)(BASE+(x))))
-#define REGB(x)	(*((BYTE *)(BASE+(x))))
+#define REG(x)	(*((DWORD *)((ULONG_PTR)PktdrvMmioBase+(x))))
```
mborgerson: Can we create inline functions instead of these macros?

JayFoxRox: Then we'll need to modify it in the caller as we'd have to separate reads and writes.

I think this would be an unrelated change, which should be a separate PR.

mborgerson: I'm not a big fan of the way this code is written, I guess we can clean it up in a follow-up PR. I thought *this* PR was supposed to do some cleaning, but that's OK.

Another issue I see here is that we need to use the `volatile` keyword when writing to MMIO to prevent the compiler from optimizing this stuff out.

JayFoxRox: > I'm not a big fan of the way this code is written, I guess we can clean it up in a follow-up PR. I thought this PR was supposed to do some cleaning, but that's OK.

This wasn't really a clean-up, but just a hack so xbe-loader can relocate the NIC. See https://github.com/XboxDev/nxdk/pull/100#discussion_r279218102

That's really the only purpose that I can remember.

> Another issue I see here is that we need to use the volatile keyword when writing to MMIO to prevent the compiler from optimizing this stuff out.

Yes. Should possibly happen in another PR *before* this one, just to avoid any regressions due to differences (as compiler can't assume a constant address anymore if we use a variable).

---

I've also raised some other concerns on Discord for a quick-discussion, but I'll move them here:

Moving away from macros would possibly make some operations non-atomic: `REG(NvRegDuplexMode) &= NVREG_DUPLEX_MODE_FDMASK;` which is read/modify/write would have to split this into 3 separate steps (which might affect behaviour).

Also readable code like `REG(NvRegMIIStatus)=REG(NvRegMIIStatus);` would become something like `write_reg(NvRegMIIStatus, read_reg(NvRegMIIStatus));`

I'm proposing structs for MMIO instead: `regs.NvRegMIIStatus = regs.NvRegMIIStatus;`.

However, these sort of changes should happen in a follow-up.

lib/net/pktdrv/pktdrv.c
```diff
@@ -42,6 +42,9 @@
 extern unsigned long times(void *);
 
 //temporary dirty interface
+void* PktdrvMmioBase = (void*)0xFEF00000;
```
mborgerson: Let's make this a `const` so the compiler can avoid doing reads to get the base address.

lib/net/pktdrv/pktdrv.c
```diff
@@ -342,10 +345,9 @@ enum {
 
 
 //Register access macros for XBOX
-#define	BASE	0xFEF00000
-#define REG(x)	(*((DWORD *)(BASE+(x))))
-#define REGW(x)	(*((WORD *)(BASE+(x))))
-#define REGB(x)	(*((BYTE *)(BASE+(x))))
+#define REG(x)	(*((DWORD volatile *)((ULONG_PTR)PktdrvMmioBase+(x))))
+#define REGW(x)	(*((WORD volatile *)((ULONG_PTR)PktdrvMmioBase+(x))))
+#define REGB(x)	(*((BYTE volatile *)((ULONG_PTR)PktdrvMmioBase+(x))))
```
JayFoxRox: nit: I prefer `volatile TYPE`, but I'm fine eitherway

JayFoxRox:
This has been sitting here for a week.

Someone has to do a final review, test and merge it.

lib/net/pktdrv/pktdrv.h
mborgerson: I would prefer we had a function that wrapped these variables for setting, but that can happen in a follow up.

mborgerson:
Thanks

lib/net/pktdrv/pktdrv.h
JayFoxRox: For the record: A similar issue exists for USB.

PR #101, opened 2019-05-03
sdl: Update submodule

thrimbor:
This updates our SDL submodule to the current HEAD (which merges upstream master). Makefile slightly adjusted to work with the newer version.


mborgerson:
Thanks

PR #102, opened 2019-05-06
WinAPI error and memory functions

thrimbor:
This adds some simple parts of the WinAPI, namely error handling and virtual memory handling.

This originally was part of my work on the file finding API (mentioned in Discord), but I decided to split it out as I required it for my current work on integrating dlmalloc (which recently got added to PDCLib to replace the old primitive malloc) - since dlmalloc already supports Win32, this makes it much easier to build for nxdk.

The code has received testing as part of my work on my other branches.

### DIFF
--- a/lib/winapi/error.c
+++ b/lib/winapi/error.c
@@ -1,0 +1,14 @@
+#include <winbase.h>
+#include <threads.h>
+
+static thread_local DWORD lastError = 0;
+
+DWORD GetLastError (void)
+{
+    return lastError;
+}
+
+void SetLastError (DWORD error)
+{
+    lastError = error;
+}
--- a/lib/winapi/windef.h
+++ b/lib/winapi/windef.h
@@ -1,0 +1,6 @@
+#ifndef __WINDEF_H__
+#define __WINDEF_H__
+
+#include <xboxkrnl/xboxdef.h>
+
+#endif
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,7 @@ NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
                -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt \
                -isystem $(NXDK_DIR)/lib/pdclib/include \
                -I$(NXDK_DIR)/lib/pdclib/platform/xbox/include \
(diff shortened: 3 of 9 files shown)


lib/winapi/error.c
```diff
@@ -0,0 +1,14 @@
+#include <winbase.h>
+#include <threads.h>
+
+thread_local DWORD lastError = 0;
```
JayFoxRox: Shouldn't this be static?

lib/winapi/windef.h
```diff
@@ -0,0 +1,21 @@
...
+typedef unsigned short WORD, WCHAR, *PWSTR;
+typedef unsigned long ULONG, *PULONG;
+typedef unsigned int DWORD, *PDWORD, *LPDWORD;
+
+typedef unsigned int SIZE_T;
+
+typedef int BOOL, *PBOOL;
+typedef const char *PCSZ, *PCSTR, *LPCSTR;
+
+#endif
```
JayFoxRox: I believe these `typedef` also exist in xboxkrnl.h? Shouldn't they be removed from there?

I'm surprised you didn't get a warning or even errors.

lib/winapi/error.c
thrimbor: It should and it is now.

lib/winapi/windef.h
thrimbor: Redundant typedefs seem to be fine in C (at least in clang), as long as they're all identical.
My intention for duplicating them was to keep xboxkrnl.h independent from my WinAPI work, especially since I didn't know what else might eventually end up in windef.h (and therefore pollute the namespace) - I added the types on an as-needed basis.
If de-duplicating them is considered cleaner, I can add that change though.

mborgerson: I think the cleanest approach (and my personal preference) is to have these types defined in one place.

thrimbor: I added another commit that splits out the typdefs into `xboxdef.h`, which gets included from both `xboxkrnl.h` and `windef.h` (I didn't name the new header `windef.h` as suggested on Discord to avoid include guard conflicts etc.).

mborgerson:
Thanks

PR #103, opened 2019-05-06
Sdl: update submodule

dracc:
Updated the SDL2 submodule.
The sdl_ttf sample received an update too in order to reflect these changes.
The main SDL sample did not change however as its `_exit` differs in functionality from the PDClib one.


mborgerson:
Thanks

PR #104, opened 2019-05-07
sdlsample: Remove local _exit() declaration

dracc:
When #103 has been merged, a local `_exit()` declaration become extraneous.

### DIFF
--- a/samples/sdl/main.c
+++ b/samples/sdl/main.c
@@ -18,13 +18,13 @@
 #include "string.h"
 #include <SDL.h>
 
-int _exit(int x)
+static void printSDLErrorAndReboot(void)
 {
-    debugPrint("_exit called\n");
-    debugPrint( " SDL_Error: %s\n", SDL_GetError() );
+    debugPrint("SDL_Error: %s\n", SDL_GetError());
+    debugPrint("Rebooting in 5 seconds.\n");
     XSleep(5000);
+    pb_kill();
     XReboot();
-    return 0;
 }
 
 static const struct {
@@ -182,14 +182,6 @@ const extern int SCREEN_HEIGHT;
 
 static SDL_Rect *positions, *velocities;
 
-/* Call this instead of exit(), so we can clean up SDL: atexit() is evil. */
-static void
-quit(int rc)
-{
-    SDL_VideoQuit();
-    _exit(rc);
-}
-
 SDL_Texture *
 LoadSprite(SDL_Renderer *renderer, char *file)
 {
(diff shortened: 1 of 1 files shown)


samples/sdl/main.c
```diff
@@ -18,10 +18,10 @@
 #include "string.h"
 #include <SDL.h>
 
-int _exit(int x)
+int Reboot(int x)
```
JayFoxRox: nit: `reboot` (or make `quit` &rarr; `Quit`).

JayFoxRox: You can also improve the function signature now: `static void Reboot()` or possibly `static void Reboot(unsigned int step)` (where `step` should be printed / have a unique number).

samples/sdl/main.c
```diff
@@ -187,7 +187,7 @@ static void
 quit(int rc)
 {
     SDL_VideoQuit();
-    _exit(rc);
+    Reboot(rc);
```
JayFoxRox: There's only 1 single instance where SDL video isn't initialized yet.
Consider merging `quit` and `Reboot` into one function - they do almost exactly the same.

You could just move `SDL_VideoQuit` into the caller for example.

samples/sdl/main.c
JayFoxRox: The name of this function is misleading. It's more of a `PrintErrorAndThenQuitAfter5Seconds`.

*I can't think of a good name myself.*

samples/sdl/main.c
JayFoxRox: I still think this argument is valid; especially as we use arbitrary return codes (1 and 5; whatever those mean).

Eitherway, this is "better than master" and it should be merged soon.

samples/sdl/main.c
```diff
@@ -187,7 +186,7 @@ static void
 quit(int rc)
```
thrimbor: nit: I don't consider `quit` to be a good name (the name sounds a bit like it's standard library functionality to me). Can't think of a better name right now, though.

thrimbor: The parameter is unused now.

samples/sdl/main.c
```diff
@@ -18,13 +18,12 @@
 #include "string.h"
 #include <SDL.h>
 
-int _exit(int x)
+static void printSDLErrorAndReboot(void)
 {
-    debugPrint("_exit called\n");
-    debugPrint( " SDL_Error: %s\n", SDL_GetError() );
+    debugPrint("Reboot called\n");
```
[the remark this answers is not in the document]
JayFoxRox: Good solution is probably to move this print down, and have it say "Rebooting in 5 seconds", so the user is aware what's about to happen.

samples/sdl/main.c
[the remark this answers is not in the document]
JayFoxRox: Also see existing discussion at https://github.com/XboxDev/nxdk/pull/104#discussion_r281714165

JayFoxRox: See existing discussion at https://github.com/XboxDev/nxdk/pull/104#discussion_r281713666 (which used to be resolved by @dracc )

JayFoxRox:
Code looks good to me now. I did not try running it.

thrimbor:
Tested and merged. Thanks!

PR #106, opened 2019-05-07
Add SDL input samples

dracc:
Samples which requires XboxDev/nxdk-sdl#5.

### DIFF
--- a/samples/sdl_gamecontroller/main.c
+++ b/samples/sdl_gamecontroller/main.c
@@ -1,0 +1,105 @@
+#include <debug.h>
+#include <pbkit/pbkit.h>
+#include <hal/video.h>
+#include <hal/xbox.h>
+#include <assert.h>
+#include <SDL.h>
+#include <stdbool.h>
+
+int main() {
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+    int ret = pb_init();
+    if (ret != 0) {
+        XSleep(2000);
+        return -1;
+    }
+
+    pb_show_debug_screen();
+
+    SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER);
+
+    int gameControllerIndex = 0;
+    SDL_GameController *gameController = SDL_GameControllerOpen(gameControllerIndex);
+    if (gameController == NULL) {
+        debugPrint("Could not open GameController %i, please plug in a\n", gameControllerIndex);
+        debugPrint("compatible controller into port 1. Exiting...\n");
+        SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+        XSleep(2000);
+        pb_kill();
+        return -1;
+    }
+
+    while (true) {
+        // Fetch current GameController state
+        SDL_GameControllerUpdate();
+        debugClearScreen();
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_A),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_A));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_B),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_B));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_X),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_X));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_Y),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_Y));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_BACK),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_BACK));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_START),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_START));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_LEFTSTICK),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_LEFTSTICK));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_RIGHTSTICK),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_RIGHTSTICK));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_LEFTSHOULDER),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_LEFTSHOULDER));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_RIGHTSHOULDER),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_RIGHTSHOULDER));
+        debugPrint("\n");
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_DPAD_UP),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_DPAD_UP));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_DPAD_DOWN),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_DPAD_DOWN));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_DPAD_LEFT),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_DPAD_LEFT));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_DPAD_RIGHT),
+                   SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_DPAD_RIGHT));
+        debugPrint("\n");
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_LEFTX),
+                   SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_LEFTX));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_LEFTY),
+                   SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_LEFTY));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_RIGHTX),
+                   SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_RIGHTX));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_RIGHTY),
+                   SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_RIGHTY));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_TRIGGERLEFT),
+                   SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_TRIGGERLEFT));
+        debugPrint("%s:%i\n",
+                   SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_TRIGGERRIGHT),
+                   SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_TRIGGERRIGHT));
+        XSleep(100);
+    }
+
+    SDL_GameControllerClose(gameController);
+    SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+    pb_kill();
+    return 0;
+}
--- a/samples/sdl_gamecontroller_event/main.c
+++ b/samples/sdl_gamecontroller_event/main.c
@@ -1,0 +1,64 @@
+#include <debug.h>
+#include <pbkit/pbkit.h>
+#include <hal/video.h>
+#include <hal/xbox.h>
+#include <assert.h>
+#include <SDL.h>
+#include <stdbool.h>
+
+int main() {
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+    int ret = pb_init();
+    if (ret != 0) {
+        XSleep(2000);
+        return -1;
+    }
+
+    pb_show_debug_screen();
+
+    SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER);
+
+    int gameControllerIndex = 0;
+
+    SDL_GameController *gameController = SDL_GameControllerOpen(gameControllerIndex);
+    if (gameController == NULL) {
+        debugPrint("Could not open GameController %i, please plug in a\n", gameControllerIndex);
+        debugPrint("compatible controller into port 1. Exiting...\n");
+        SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+        XSleep(2000);
+        pb_kill();
+        return -1;
+    }
+    
+    debugPrint("GameController name: %s\n", SDL_GameControllerName(gameController));
+    debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(gameController));
+    debugPrint("GameControllerEvent: %i\n", SDL_GameControllerEventState(SDL_QUERY));
+    debugPrint("\n");
+
+    while (true) {
+        SDL_Event event;
+        while (SDL_PollEvent(&event)) {
+            if (event.type == SDL_CONTROLLERDEVICEADDED) {
+                debugPrint("Controller %i added. (NOT opened yet)\n", event.cdevice.which);
+            } else if (event.type == SDL_CONTROLLERDEVICEREMOVED) {
+                debugPrint("Controller %i removed, exiting...\n", event.cdevice.which);
+                SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+                XSleep(2000);
+                pb_kill();
+                return -1;
+            } else if (event.type == SDL_CONTROLLERBUTTONDOWN) {
+                debugPrint("Button \"%s\" pressed\n",
+                           SDL_GameControllerGetStringForButton((SDL_GameControllerButton)event.cbutton.button));
+            } else if (event.type == SDL_CONTROLLERAXISMOTION) {
+                debugPrint("Axis \"%s\" moved, new value: %i\n",
+                           SDL_GameControllerGetStringForAxis((SDL_GameControllerAxis)event.caxis.axis),
+                           event.caxis.value);
+            }
+        }
+    }
+
+    SDL_GameControllerClose(gameController);
+    SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+    pb_kill();
+    return 0;
+}
--- a/samples/sdl_joystick_event/main.c
+++ b/samples/sdl_joystick_event/main.c
@@ -1,0 +1,65 @@
+#include <debug.h>
+#include <pbkit/pbkit.h>
+#include <hal/video.h>
+#include <hal/xbox.h>
+#include <assert.h>
+#include <SDL.h>
+#include <stdbool.h>
+
+#define ARRAY_SIZE(x) (sizeof(x)/sizeof(x[0]))
+
+int main() {
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+    int ret = pb_init();
+    if (ret != 0) {
+        XSleep(2000);
+        return -1;
+    }
+
+    pb_show_debug_screen();
+
+    SDL_InitSubSystem(SDL_INIT_JOYSTICK);
+
+    int joystickIndex = 0;
+    SDL_Joystick *joy = SDL_JoystickOpen(joystickIndex);
+    if (joy == NULL) {
+        debugPrint("Could not open Joystick %i, exiting...", joystickIndex);
+        SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+        XSleep(2000);
+        pb_kill();
+        return -1;
+    }
+
+    debugPrint("Number of axes   : %2i\n", SDL_JoystickNumAxes(joy));
+    debugPrint("Number of buttons: %2i\n", SDL_JoystickNumButtons(joy));
+    debugPrint("Number of balls  : %2i\n", SDL_JoystickNumBalls(joy));
+    debugPrint("Number of hats   : %2i\n", SDL_JoystickNumHats(joy));
+    debugPrint("\n");
+
+    while (true) {
+        SDL_Event event;
+        while (SDL_PollEvent(&event)) {
+            if (event.type == SDL_JOYBUTTONDOWN) {
+                debugPrint("Button %i pressed\n", event.jbutton.button);
+            } else if (event.type == SDL_JOYAXISMOTION) {
+                debugPrint("Axis %i moved, new value: %i\n", event.jaxis.axis, event.jaxis.value);
+            } else if (event.type == SDL_JOYHATMOTION) {
+                debugPrint("Hat %i moved, new value: %i\n", event.jhat.hat, event.jhat.value);
+            } else if (event.type == SDL_JOYDEVICEREMOVED) {
+                if (event.jdevice.which == joystickIndex) {
+                    debugPrint("Joystick %i removed, exiting...", event.jdevice.which);
+                    SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+                    XSleep(2000);
+                    pb_kill();
+                    break;
+                }
+            }
+        }
+    }
+
+    SDL_JoystickClose(joy);
+    SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+    pb_kill();
+
+    return 0;
+}
--- a/samples/sdl_gamecontroller/Makefile
+++ b/samples/sdl_gamecontroller/Makefile
@@ -1,0 +1,7 @@
+XBE_TITLE = sdl_gamecontroller
+GEN_XISO = $(XBE_TITLE).iso
+SRCS += $(CURDIR)/main.c
+NXDK_DIR = $(CURDIR)/../..
+NXDK_SDL = y
+
+include $(NXDK_DIR)/Makefile
--- a/samples/sdl_gamecontroller_event/Makefile
+++ b/samples/sdl_gamecontroller_event/Makefile
@@ -1,0 +1,7 @@
+XBE_TITLE = sdl_gamecontroller_event
+GEN_XISO = $(XBE_TITLE).iso
+SRCS += $(CURDIR)/main.c
+NXDK_DIR = $(CURDIR)/../..
+NXDK_SDL = y
+
+include $(NXDK_DIR)/Makefile
--- a/samples/sdl_joystick_event/Makefile
+++ b/samples/sdl_joystick_event/Makefile
@@ -1,0 +1,7 @@
+XBE_TITLE = sdl_joystick_event
+GEN_XISO = $(XBE_TITLE).iso
+SRCS += $(CURDIR)/main.c
+NXDK_DIR = $(CURDIR)/../..
+NXDK_SDL = y
+
+include $(NXDK_DIR)/Makefile


thrimbor:
I recommend delaying this until the SDL work has landed (and this commit was rebased). I'd rather have a non-bisectable commit in a submodule than in nxdk.

samples/sdl_joystick_gamepad/main.cpp
```diff
@@ -0,0 +1,50 @@
+#include <xboxrt/debug.h>
+#include <pbkit/pbkit.h>
+#include <hal/xbox.h>
+#include <assert.h>
+#include <SDL.h>
+
+#define ARRAY_SIZE(x) (sizeof(x)/sizeof(x[0]))
```
JayFoxRox: Unused macro

samples/sdl_joystick_gamepad/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+  SDL_Event event;
+
+  debugPrint("Num joysticks      : %02i\n", SDL_NumJoysticks());
+  debugPrint("GameController name: %s\n", SDL_GameControllerName(joy));
+  debugPrint("Is GameController  : %02i\n", SDL_IsGameController(0));
+  debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(joy));
+  //debugPrint("JoystickEvent      : %02i\n", SDL_JoystickEventState(SDL_QUERY));
+  //debugPrint("GameControllerEvent: %02i\n", SDL_GameControllerEventState(SDL_DISABLE));
+  //debugPrint("JoystickEvent      : %02i\n", SDL_JoystickEventState(SDL_DISABLE));
```
JayFoxRox: Dead code should be removed

samples/sdl_joystick_gamepad/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+  debugPrint("GameControllerEvent: %02i\n", SDL_GameControllerEventState(SDL_ENABLE));
```
JayFoxRox: This seems to be a setter, that's being used like a getter.

`debugPrint` shouldn't be called with sideeffects, because it might be removed in real applications (so it shouldn't have sideeffects).

samples/sdl_joystick_gamepad/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+  debugPrint("\n");
+  assert(SDL_IsGameController(0));
+    
+  while (true) {
+    while (SDL_PollEvent(&event)) {
+      if (event.type == SDL_CONTROLLERBUTTONDOWN) {
+        debugPrint("Button \"%s\" pressed\n", SDL_GameControllerGetStringForButton(static_cast<SDL_GameControllerButton>(event.cbutton.button)));
```
JayFoxRox: I think there's no benefit in using C++ in this sample as it's demo-ing a C API.
Using C would remove complexity in casts.

samples/sdl_joystick_gamepad/Makefile
```diff
@@ -0,0 +1,10 @@
+DEBUG = y
+XBE_TITLE = sdl_joystick_gamepad
```
JayFoxRox: This sample should be `sdl_joystick_gamecontroller` to reflect what it actually does.
It could also just be `sdl_gamecontroller` (recommended) as that's the API prefix.

samples/sdl_joystick_gamepad/Makefile
```diff
@@ -0,0 +1,10 @@
...
+GEN_XISO = $(XBE_TITLE).iso
+SRCS += $(CURDIR)/main.cpp
+NXDK_DIR = $(CURDIR)/../..
+NXDK_SDL = y
+
+all_local: all
```
JayFoxRox: nit: This shouldn't be necessary for simple examples like this one? Do the other samples do it?

samples/sdl_joystick_gamepad/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+int main() {
+  int ret = pb_init();
+  if (ret != 0) {
+    XSleep(2000);
+    return -1;
+  }
+
+  pb_show_debug_screen();
+
+  SDL_GameController *joy;
```
JayFoxRox: If you keep declarations in code, it should be moved to where it's actually used (next block, with `SDL_GameControllerOpen`)

samples/sdl_joystick_gamepad/main.cpp
```diff
@@ -0,0 +1,50 @@
...
+  SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER);
+
+  joy = SDL_GameControllerOpen(0);
+
+  SDL_Event event;
```
JayFoxRox: Again: should be moved to where it's actually used or moved near the top.

It can be part of the `while` loop (recommended / personal preference).

samples/sdl_joystick_gamepad/Makefile
```diff
@@ -0,0 +1,10 @@
+DEBUG = y
```
JayFoxRox: Please make commit titles consistent: one of the sample commit uses "sdl:" the other one uses "samples:" as prefix, although they seem to do almost exactly the same.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,46 @@
...
+  debugPrint("Num joysticks      : %02i\n", SDL_NumJoysticks());
+  debugPrint("GameController name: %s\n", SDL_GameControllerName(gameController));
+  debugPrint("Is GameController  : %02i\n", SDL_IsGameController(0));
+  debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(gameController));
+  debugPrint("GameControllerEvent: %02i\n", SDL_GameControllerEventState(SDL_QUERY));
+
+  debugPrint("\n");
+  assert(SDL_IsGameController(0));
+
+  SDL_Event event;
```
JayFoxRox: Can be moved into the loop. There's no reason this outer scope needs it.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,46 @@
...
+    XSleep(2000);
+    return -1;
+  }
+
+  pb_show_debug_screen();
+
+  SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER);
+
+  SDL_GameController *gameController;
+  gameController = SDL_GameControllerOpen(0);
```
JayFoxRox: nit: can be `SDL_GameController *gameController = SDL_GameControllerOpen(0);`

*Existing code is already like this: `int ret = pb_init();`*

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,46 @@
...
+    while (SDL_PollEvent(&event)) {
+      if (event.type == SDL_CONTROLLERBUTTONDOWN) {
+        debugPrint("Button \"%s\" pressed\n", SDL_GameControllerGetStringForButton((SDL_GameControllerButton)event.cbutton.button));
+      } else if (event.type == SDL_CONTROLLERAXISMOTION) {
+        debugPrint("Axis %s moved, new value: %i\n", SDL_GameControllerGetStringForAxis((SDL_GameControllerAxis)event.caxis.axis), event.caxis.value);
+      }
+    }
+  }
+
+  pb_kill();
```
JayFoxRox: If we are showing destruction of objects, there's probably stuff to do for SDL?

dracc: Thanks, taken care of. :+1:

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,48 @@
...
+  if (ret != 0) {
+    XSleep(2000);
+    return -1;
+  }
+
+  pb_show_debug_screen();
+
+  SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER);
+
+  SDL_GameController *gameController = SDL_GameControllerOpen(0);
```
JayFoxRox: What if the controller isn't opened successfully? Probably needs error handling to avoid crashes or strange behaviour.

JayFoxRox: One would normally search for a valid joystick index; a comment or self-describing code might be useful (`int joystick_index = 0;` as helper to show the meaning of constant 0)

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,48 @@
...
+  debugPrint("\n");
+  assert(SDL_IsGameController(0));
+
+  while (true) {
+    SDL_Event event;
+    while (SDL_PollEvent(&event)) {
+      if (event.type == SDL_CONTROLLERBUTTONDOWN) {
+        debugPrint("Button \"%s\" pressed\n", SDL_GameControllerGetStringForButton((SDL_GameControllerButton)event.cbutton.button));
+      } else if (event.type == SDL_CONTROLLERAXISMOTION) {
+        debugPrint("Axis %s moved, new value: %i\n", SDL_GameControllerGetStringForAxis((SDL_GameControllerAxis)event.caxis.axis), event.caxis.value);
```
JayFoxRox: nit: Button uses `"\"%s\""`, this just uses `"%s"`

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,48 @@
...
+  debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(gameController));
+  debugPrint("GameControllerEvent: %02i\n", SDL_GameControllerEventState(SDL_QUERY));
+
+  debugPrint("\n");
+  assert(SDL_IsGameController(0));
+
+  while (true) {
+    SDL_Event event;
+    while (SDL_PollEvent(&event)) {
+      if (event.type == SDL_CONTROLLERBUTTONDOWN) {
```
JayFoxRox: nit: I assume there's also a `BUTTONUP`? That would be equally interesting.

Overall I still think that the state-based API is *much, much, much* more important than the event based API. I have never heard of anyone using the event based APIs for SDL2 in practical use-cases.

dracc: I agree, however the spaminess without some kind of frame limiter would be immense and I don't think we should extend the samples that wide.

samples/sdl_gamecontroller_event/main.c
```diff
@@ -0,0 +1,57 @@
...
+  int gameControllerIndex = 0;
+  SDL_GameController *gameController = SDL_GameControllerOpen(gameControllerIndex);
+  if (gameController == NULL) {
+    debugPrint("Could not open GameController %i, exiting...", gameControllerIndex);
+    SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+    XSleep(2000);
+    return -1;
+  }
+
+  debugPrint("Num joysticks      : %02i\n", SDL_NumJoysticks());
```
JayFoxRox: Why "02" alignment?

samples/sdl_gamecontroller_event/main.c
```diff
@@ -0,0 +1,57 @@
...
+  debugPrint("GameController name: %s\n",   SDL_GameControllerName(gameController));
+  debugPrint("Is GameController  : %02i\n", SDL_IsGameController(gameControllerIndex));
```
JayFoxRox: Why "02" alignment?

How about: `"%s", SDL_IsGameController(gameControllerIndex) ? "Yes" : "No"`

JayFoxRox: Shouldn't this be checked *before* opening it as a `gamecontroller`?
Once you aquire the `gamecontroller`, you shouldn't have to use `gameControllerIndex` anymore.

samples/sdl_gamecontroller_event/main.c
JayFoxRox: 1. Joystick is not interesting for a gamecontroller sample
2. This message suggests that multiple gamecontrollers will react (they don't)

Unless the following code is ran for each gamecontroller, this message should be replaced with one that suggests that the gamecontroller in "Port 1" is being used.

samples/sdl_gamecontroller_event/main.c
```diff
@@ -0,0 +1,57 @@
...
+  debugPrint("GameController Map : %s\n",   SDL_GameControllerMapping(gameController));
+  debugPrint("GameControllerEvent: %02i\n", SDL_GameControllerEventState(SDL_QUERY));
+
+  debugPrint("\n");
+  assert(SDL_IsGameController(gameControllerIndex));
```
JayFoxRox: This check is way too late. If a joystick isn't a gamecontroller, it won't open correctly using the gamecontroller API.

samples/sdl_gamecontroller_event/main.c
dracc: No real reason, just thought it looked nice.

samples/sdl_gamecontroller_event/main.c
dracc: Good idea. :+1:

samples/sdl_gamecontroller/main.c
JayFoxRox: You should at least add a comment then to make people aware of this option.

Also buttons can't be pressed that fast, so it won't spam a lot.
Meanwhile, any axis might jitter will probably cause > 100 of messages per second (which will drain any other message).

dracc: Problem is that the non-event kind doesn't do "This has recently been pressed" but "is it currently pressed?". But your suggestion of adding an `XSleep()` could fix that problem.

samples/sdl_gamecontroller_event/main.c
dracc: Removed the zero padding - update soon to be pushed.

samples/sdl_gamecontroller/main.c
JayFoxRox: I'm not sure what you mean.

> Problem is that the non-event kind doesn't do "This has recently been pressed" but "is it currently pressed?".

This is the event-sample; anything related to the non-event sample is irrelevant for now.

*I don't see the issue of not knowing "This has recently been pressed" either; games are written as interactive applications, which draw on their own. There typically isn't any lockstep either as it would complicate the game logic. It's typically compensated by running multiple iterations of the physics simulation, which can benefit from updated inputs. You'd only need to know about higher sub-frame precision if your framerate is very low. But you wouldn't optimize your game for that as the player would be controlling "blindly".*

> But your suggestion of adding an XSleep() could fix that problem.

I did not recommend this, and it likely won't prevent spam as the events will still queue up probably.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,67 @@
...
+  SDL_GameController *gameController = SDL_GameControllerOpen(gameControllerIndex);
+  if (gameController == NULL) {
+    debugPrint("Could not open GameController %i, exiting...", gameControllerIndex);
+    SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+    XSleep(2000);
+    return -1;
+  }
+
+  debugPrint("GameController name: %s\n", SDL_GameControllerName(gameController));
+  debugPrint("Is GameController  : %s\n", SDL_IsGameController(gameControllerIndex)?"Yes":"No");
```
JayFoxRox: See comment from earlier: This should happen *before* opening it.

This can only ever be "Yes"; as `gameController` is `NULL` otherwise *and* the assertion would have failed.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,67 @@
...
+    debugPrint("%s:%i\n",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_Y),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_Y));
+    debugPrint("%s:%i                 ",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_BACK),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_BACK));
+    debugPrint("%s:%i\n",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_START),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_START));
+    debugPrint("%s:%i       ",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_LEFTSTICK),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_LEFTSTICK));
+    debugPrint("%s:%i\n",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_RIGHTSTICK),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_RIGHTSTICK));
+    debugPrint("%s:%i ",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_LEFTSHOULDER),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_LEFTSHOULDER));
```
JayFoxRox: Are these using the correct "Black" / "White" names?

These should map as they do for backwards compatibility on 360 (official MS mapping).

There should probably be a comment in the source code which explains this detail.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,67 @@
...
+    debugPrint("%s:%i\n\n",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_RIGHTSHOULDER),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_RIGHTSHOULDER));
```
JayFoxRox: These "\n\n" should probably be move into separate `debugPrint` to be more visible.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,67 @@
...
+  debugPrint("GameController name: %s\n", SDL_GameControllerName(gameController));
+  debugPrint("Is GameController  : %s\n", SDL_IsGameController(gameControllerIndex)?"Yes":"No");
+  debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(gameController));
+  debugPrint("\n");
+
+  while (true) {
+    // Fetch current GameController state
+    SDL_GameControllerUpdate();
+    debugClearScreen();
+    debugPrint("%s:%i      ",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_A),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_A));
```
JayFoxRox: Consider grouping the `debugPrint` per line, or even combining them into a single print.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,67 @@
...
+    debugPrint("%s:%i      ",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_B),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_B));
+    debugPrint("%s:%i      ",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_X),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_X));
+    debugPrint("%s:%i\n",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_Y),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_Y));
+    debugPrint("%s:%i                 ",SDL_GameControllerGetStringForButton(SDL_CONTROLLER_BUTTON_BACK),SDL_GameControllerGetButton(gameController, SDL_CONTROLLER_BUTTON_BACK));
```
JayFoxRox: Why does this one have so many spaces? Wouldn't the amount of padding depend on the length of the string returned by `SDL_GameControllerGetStringForButton`?

samples/sdl_gamecontroller/main.c
JayFoxRox: Is it even necessary to print multiple outputs on the same line?

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,67 @@
...
+    debugPrint("%s:%6i               ",SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_LEFTX),SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_LEFTX));
+    debugPrint("%s:%6i\n",SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_RIGHTX),SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_RIGHTX));
+    debugPrint("%s:%6i               ",SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_LEFTY),SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_LEFTY));
+    debugPrint("%s:%6i\n",SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_RIGHTY),SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_RIGHTY));
+    debugPrint("%s:%6i   ",SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_TRIGGERLEFT),SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_TRIGGERLEFT));
+    debugPrint("%s:%6i",SDL_GameControllerGetStringForAxis(SDL_CONTROLLER_AXIS_TRIGGERRIGHT),SDL_GameControllerGetAxis(gameController, SDL_CONTROLLER_AXIS_TRIGGERRIGHT));
+    XSleep(100);
+  }
+
+  if (gameController != NULL) {
```
JayFoxRox: This couldn't be reached if `gameController` wasn't `NULL`

samples/sdl_gamecontroller_event/main.c
```diff
@@ -0,0 +1,56 @@
...
+                   SDL_GameControllerGetStringForButton((SDL_GameControllerButton)event.cbutton.button));
+      } else if (event.type == SDL_CONTROLLERAXISMOTION) {
+        debugPrint("Axis \"%s\" moved, new value: %i\n",
+                   SDL_GameControllerGetStringForAxis((SDL_GameControllerAxis)event.caxis.axis),
+                   event.caxis.value);
+      }
+    }
+  }
+
+  if (gameController != NULL) {
```
JayFoxRox: This sample can still continue, even if gamecontroller wasn't successfully opened. There just won't be any events.

But it makes the assumption that it will only continue if the gamecontroller was opened successfully. So this check is probably useless, or the exit above is useless.

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,63 @@
...
+    SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+    Xsleep(2000);
+    return -1;
+  }
+
+  const char* buttonNames[10] = {"A", "B", "X", "Y", "White", "Black", "Back", "Start", "LeftThumb", "RightThumb"};
+  const char* axisNames[6] = {"LX", "LY", "LTrigg", "RX", "RY", "RTrigg"};
+
+  debugPrint("Num joysticks    : %02i\n", SDL_NumJoysticks());
+  debugPrint("Joystick name    : %s\n", SDL_JoystickNameForIndex(joystickIndex));
```
JayFoxRox: There's an API to get name by index, and another API to get it by opened handle.
At this point you already own a handle.

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,63 @@
...
+    debugPrint("Could not open Joystick %i, exiting...", joystickIndex);
+    SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+    Xsleep(2000);
+    return -1;
+  }
+
+  const char* buttonNames[10] = {"A", "B", "X", "Y", "White", "Black", "Back", "Start", "LeftThumb", "RightThumb"};
+  const char* axisNames[6] = {"LX", "LY", "LTrigg", "RX", "RY", "RTrigg"};
+
+  debugPrint("Num joysticks    : %02i\n", SDL_NumJoysticks());
```
JayFoxRox: (Same issues as previously pointed out in gamecontroller sample)

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,63 @@
...
+
+  while (true) {
+    SDL_Event event;
+    while (SDL_PollEvent(&event)) {
+      if (event.type == SDL_JOYBUTTONDOWN) {
+        debugPrint("Button \"%s\" pressed\n", buttonNames[event.jbutton.button]);
+      } else if (event.type == SDL_JOYAXISMOTION) {
+        debugPrint("Axis %s moved, new value: %i\n", axisNames[event.jaxis.axis], event.jaxis.value);
+      } else if (event.type == SDL_JOYHATMOTION) {
+        debugPrint("Hat moved, new value: %i\n", event.jhat.value);
```
JayFoxRox: Isn't there an index for the hat?

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,63 @@
...
+  SDL_Joystick *joy = SDL_JoystickOpen(joystickIndex);
+  if (joy == NULL) {
+    debugPrint("Could not open Joystick %i, exiting...", joystickIndex);
+    SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+    Xsleep(2000);
+    return -1;
+  }
+
+  const char* buttonNames[10] = {"A", "B", "X", "Y", "White", "Black", "Back", "Start", "LeftThumb", "RightThumb"};
+  const char* axisNames[6] = {"LX", "LY", "LTrigg", "RX", "RY", "RTrigg"};
```
JayFoxRox: I wonder if we should even keep this. It's mainly used to emulate the SDL gamecontroller API, when the only real benefit of the joystick API is that you can use non-Xbox style gamepads (where this would fail).

samples/sdl_gamecontroller/main.c
dracc: Yes, that is true. Padding just made the formatting nice for the current output - so far I have not seen any reason why names should change between controller types.

samples/sdl_gamecontroller/main.c
dracc: They output the x360 mapping name (LeftTrigger and RightTrigger [or something like that]).
Mapping is 1:1 what was used by XInput in SDL_gamecontrollerdb.h, I'm not sure how to map it better.

samples/sdl_gamecontroller/main.c
dracc: Consider it done. :+1:

samples/sdl_gamecontroller/main.c
dracc: All could probably be listed, one per line on a single screen. I found the grouping more appealing though.

samples/sdl_gamecontroller/main.c
JayFoxRox: These names must be "White" and "Black" for original Xbox controllers, whereas they would probably be something like "LB" (Left-Button) and "RB" (Right-Button) [or the other way around] on 360 controllers.

Our gamecontroller mapping has to reflect that.

samples/sdl_gamecontroller/main.c
JayFoxRox: - What if you connect a 360 gamepad to the original Xbox?
- What about PC gamepads?
- What about Playstation gamepads?
- What about steering wheels where these might be named differently ("Shift paddle left" / "Shift paddle right") for usability?

I think one of the benefits of the nxdk is that we can support such devices.
So we shouldn't hardcode it.

If we really want to assume that these are constant names, then we don't have to read the names from `SDL_GameControllerGetStringForButton` either. But that defeats the point of using these APIs.

*If* we hardcode the padding, we should at least add a comment. But I'd still consider it bad code, given we have cross-file dependencies about these strings.

Putting each field on a different line, or also padding the names isn't too hard either.

samples/sdl_gamecontroller/main.c
dracc: > Overall I still think that the state-based API is much, much, much more important than the event based API. I have never heard of anyone using the event based APIs for SDL2 in practical use-cases.

This was the part I was responding to (re spaminess), sorry for being vague.

samples/sdl_joystick_event/main.c
dracc: Might be a good idea to remove it, yes.

samples/sdl_gamecontroller/main.c
dracc: Fixed, thanks. :+1:

samples/sdl_gamecontroller/main.c
dracc: I have not seen any indication that button names can be remapped freely.
I believe the whole idea of the GameController API is to make sure that the button SDL_GAMECONTROLLER_BUTTON_A always is the one located closest to where A is located on an x360 controller and that the string returned is always "a" for that particular `enum` value? I might be wrong though, the documentation of the string getter is not very clear.

JayFoxRox: Looks like labels are not implemented yet for gamecontroller mappings.

[This code in SDL2](https://github.com/spurious/SDL-mirror/blob/74e23455ab2abbced7d1ef6c610449129582788f/src/joystick/SDL_gamecontroller.c#L492) even suggests that these names are lowercase unique identifiers (key-like).

In that case the code can remain as-is. I was under the impression gamecontroller-labels had existed in SDL2 already.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,94 @@
...
+
+  debugPrint("GameController name: %s\n", SDL_GameControllerName(gameController));
+  debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(gameController));
+  debugPrint("\n");
+
+  while (true) {
+    // Fetch current GameController state
+    SDL_GameControllerUpdate();
+    debugClearScreen();
+    debugPrint("%s:%i\n%s:%i\n%s:%i\n%s:%i\n",
```
JayFoxRox: If you are printing in newlines anyway, you can also put a separate `debugPrint` for each line.
Only exception would be if this causes performance issues (screen-tearing).

samples/sdl_gamecontroller/main.c
JayFoxRox: As these names are still constant (SDL2 still lacks labels for gamecontrollers), this isn't too critical after all. The current solution is still better though. Thanks!

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,56 @@
...
+  pb_show_debug_screen();
+
+  SDL_InitSubSystem(SDL_INIT_JOYSTICK);
+
+  int joystickIndex = 0;
+  SDL_Joystick *joy = SDL_JoystickOpen(joystickIndex);
+  if (joy == NULL) {
+    debugPrint("Could not open Joystick %i, exiting...", joystickIndex);
+    SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+    Xsleep(2000);
```
JayFoxRox: `Xsleep` &rarr; `XSleep`

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  debugPrint("GameController name: %s\n", SDL_GameControllerName(gameController));
+  debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(gameController));
+  debugPrint("\n");
+
+  while (true) {
+    // Fetch current GameController state
+    SDL_GameControllerUpdate();
+    debugClearScreen();
```
JayFoxRox: This hides any information printed before this loop; could there be a better way, where you only clear one half of the screen or re-print other messages?

The sample starts, prints stuff, immediately deletes it. It's okay, but then it shouldn't print stuff before the loop.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,108 @@
...
+    SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+    XSleep(2000);
+    return -1;
+  }
+
+  debugPrint("GameController name: %s\n", SDL_GameControllerName(gameController));
+  debugPrint("GameController Map : %s\n", SDL_GameControllerMapping(gameController));
+  debugPrint("\n");
+
+  while (true) {
```
JayFoxRox: If there's an option to check if the controller is still connected, you should use it (even if it might still not work) to drop out of this loop / inform the user that the input no longer works and the app will exit in X seconds.

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,56 @@
...
+
+  debugPrint("Number of axes   : %2i\n", SDL_JoystickNumAxes(joy));
+  debugPrint("Number of buttons: %2i\n", SDL_JoystickNumButtons(joy));
+  debugPrint("Number of balls  : %2i\n", SDL_JoystickNumBalls(joy));
+  debugPrint("Number of hats   : %2i\n", SDL_JoystickNumHats(joy));
+  debugPrint("\n");
+
+  while (true) {
+    SDL_Event event;
+    while (SDL_PollEvent(&event)) {
```
JayFoxRox: Check if the joystick was disconnected, if it was, leave the loop, inform user / exit.

samples/sdl_gamecontroller/main.c
dracc: I can't find such a function. `SDL_IsGameController()` is probably the closest I can find - good enough?

samples/sdl_gamecontroller/main.c
dracc: I couldn't find such a cleaning procedure, went the "remove unreadable messages" route instead.

samples/sdl_joystick_event/main.c
dracc: Done. :+1:

RadWolfie: @dracc, you might want to update the other samples too. They're stuck in the loop even if the controller is disconnected.

samples/sdl_gamecontroller_event/main.c
```diff
@@ -0,0 +1,56 @@
...
+  pb_show_debug_screen();
+
+  SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER);
+
+  int gameControllerIndex = 0;
+
+  SDL_GameController *gameController = SDL_GameControllerOpen(gameControllerIndex);
+  if (gameController == NULL) {
+    debugPrint("Could not open GameController %i, please plug in a\n", gameControllerIndex);
+    debugPrint("compatible controller into port 1...\n");
```
RadWolfie: Where's the termination here? Otherwise the sample will be stuck in the while statement endless loop.

Should be similar to your sdl_gamecontroller sample.
https://github.com/XboxDev/nxdk/blob/a0f35be81b6856b953fbdc22e557c4097405f335/samples/sdl_gamecontroller/main.c#L24-L29

dracc: Good catch, fixed now!

samples/sdl_joystick_event/main.c
dracc: The GameController API doesn't provide such a check as far as I know, but the `sdl_gamecontroller_event` now uses it. :+1:

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,105 @@
+#include <xboxrt/debug.h>
+#include <pbkit/pbkit.h>
+#include <hal/video.h>
+#include <hal/xbox.h>
+#include <assert.h>
+#include <SDL.h>
+#include <stdbool.h>
+
+int main() {
+  XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
```
thrimbor: nit: It's less of an issue here since the whole sample is formatted equally, but I still recommend to stick to 4 spaces like most of nxdk.

(we should come up with a way to avoid having to deal with indentation, line-endings etc. manually - [EditorConfig](https://editorconfig.org/) maybe?)

samples/sdl_gamecontroller/main.c
dracc: I fail to see how SDL_GameControllerEventState could be used in a helpful manner here. Would you please enlighten me on the details?

thrimbor: It can't, ignore what I said, my fingers were faster than my eyes.

samples/sdl_gamecontroller_event/main.c
JayFoxRox: New solution isn't ideal either (will close even if controller from unused port is disconnected?). Anyhow, can be a separate discussion / review comment.

samples/sdl_gamecontroller/main.c
JayFoxRox: Please create an issue about this, if none exists yet @thrimbor ; then mark as resolved.

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,65 @@
...
+                debugPrint("Axis %i moved, new value: %i\n", event.jaxis.axis, event.jaxis.value);
+            } else if (event.type == SDL_JOYHATMOTION) {
+                debugPrint("Hat %i moved, new value: %i\n", event.jhat.hat, event.jhat.value);
+            } else if (event.type == SDL_JOYDEVICEREMOVED) {
+                if (event.jdevice.which == joystickIndex) {
+                    debugPrint("Joystick %i removed, exiting...", event.jdevice.which);
+                    SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
+                    XSleep(2000);
+                    pb_kill();
+                    break;
```
JayFoxRox: If this breaks, it will run the same deinit code again outside of the main-loop.
If you don't break, you probably still want to close the `joy` joystick.

samples/sdl_joystick_event/main.c
```diff
@@ -0,0 +1,65 @@
...
+#include <stdbool.h>
+
+#define ARRAY_SIZE(x) (sizeof(x)/sizeof(x[0]))
+
+int main() {
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+    int ret = pb_init();
+    if (ret != 0) {
+        XSleep(2000);
+        return -1;
```
JayFoxRox: Negative numbers returned from main are very uncommon.

samples/sdl_gamecontroller_event/main.c
```diff
@@ -0,0 +1,64 @@
...
+        SDL_Event event;
+        while (SDL_PollEvent(&event)) {
+            if (event.type == SDL_CONTROLLERDEVICEADDED) {
+                debugPrint("Controller %i added. (NOT opened yet)\n", event.cdevice.which);
+            } else if (event.type == SDL_CONTROLLERDEVICEREMOVED) {
+                debugPrint("Controller %i removed, exiting...\n", event.cdevice.which);
+                SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+                XSleep(2000);
+                pb_kill();
+                return -1;
```
JayFoxRox: See same comment in joystick sample about de-init after main-loop and closing of `gameController`.

samples/sdl_gamecontroller/main.c
```diff
@@ -0,0 +1,105 @@
...
+        SDL_QuitSubSystem(SDL_INIT_GAMECONTROLLER);
+        XSleep(2000);
+        pb_kill();
+        return -1;
+    }
+
+    while (true) {
+        // Fetch current GameController state
+        SDL_GameControllerUpdate();
+        debugClearScreen();
```
JayFoxRox: The existing review comments, about checking wether `gameController` is still connected, seem to have been ignored / missed.

Consider https://wiki.libsdl.org/SDL_GameControllerGetAttached?highlight=%28%5CbCategoryGameController%5Cb%29%7C%28CategoryEnum%29

samples/sdl_gamecontroller/main.c
JayFoxRox: I have proposed `SDL_GameControllerGetAttached` in another (non-outdated) discussion.

samples/sdl_gamecontroller/main.c
JayFoxRox: Resolving as unrelated. No issue has been created yet.

JayFoxRox:
Shall we:

- Merge this as-is, despite being subpar quality (for being a reference-sample, for an already well-documented SDL API), with fixes in the future? The risk is non-ideal application quality based on these samples and additional code to maintain with future changes.

*..or..*

- Should we close this; potentially spending more time on review on resubmission?

dracc:
I'll pull the trigger on closing unless someone objects.

dracc:
Closed as the design was bad and my implementation of this bad design was even worse. :+1:

PR #107, opened 2019-05-13
hal: Add LED API

dracc:
This was requested on Discord back in September of 2018.

I have no real hardware to test on ATM but it compiles without any warnings.

### DIFF
--- a/lib/hal/led.c
+++ b/lib/hal/led.c
@@ -1,0 +1,30 @@
+/*
+  Copyright and related rights waived by Lucas Eriksson via CC0
+  https://creativecommons.org/publicdomain/zero/1.0/
+*/
+
+#include <hal/led.h>
+#include <xboxkrnl/xboxkrnl.h>
+#include <assert.h>
+
+#define LED_MODE_AUTO   0x00
+#define LED_MODE_MANUAL 0x01
+
+#define SMC_REG_LEDMODE 0x07
+#define SMC_REG_LEDSEQ  0x08
+
+void XResetLED(void) {
+    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, LED_MODE_AUTO);
+}
+
+void XSetCustomLED(XLEDColor t1, XLEDColor t2, XLEDColor t3, XLEDColor t4) {
+    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, LED_MODE_MANUAL);
+
+    ULONG colorScheme =
+        ((t1 & 0x11) << 3) |
+        ((t2 & 0x11) << 2) |
+        ((t3 & 0x11) << 1) |
+        ((t4 & 0x11) << 0);
+
+    HalWriteSMBusValue(0x20, SMC_REG_LEDSEQ, FALSE, colorScheme);
+}
--- a/lib/hal/led.h
+++ b/lib/hal/led.h
@@ -1,0 +1,24 @@
+#ifndef HAL_LED_H
+#define HAL_LED_H
+
+typedef enum _XLEDColor {
+    XLED_OFF = 0x00,
+    XLED_GREEN = 0x01,
+    XLED_RED = 0x10,
+    XLED_ORANGE = 0x11
+} XLEDColor;
+
+/*
+  Restore system LED control.
+*/
+void XResetLED(void);
+
+/*
+  Set the LED color manually
+  t1-t4 are used to set the color for each segment
+  in a four step loop. For more details about front LED characteristics,
+  please see https://xboxdevwiki.net/LED
+*/
+void XSetCustomLED(XLEDColor t1, XLEDColor t2, XLEDColor t3, XLEDColor t4);
+
+#endif
--- a/samples/led/main.c
+++ b/samples/led/main.c
@@ -1,19 +1,12 @@
+#include <hal/led.h>
 #include <hal/xbox.h>
-#include <xboxkrnl/xboxkrnl.h>
-
-#define SMC_REG_LEDMODE             0x07
-#define SMC_REG_LEDSEQ              0x08
 
 void main(void)
 {
-    // Set LED to be user controlled
-    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+    // Set front LEDs to shine green, red and then turned off in a repeating pattern.
+    XSetCustomLED(XLED_GREEN, XLED_RED, XLED_OFF, XLED_OFF);
 
-    // LED pattern is:
-    //   Red:   0xC = 1100 (ON,  ON,  OFF, OFF;   ON,  ON,  OFF, OFF;   ...)
-    //   Green: 0x0 = 0000 (OFF, OFF, OFF, OFF;   OFF, OFF, OFF, OFF;   ...)
-    // This means the red LED will blink slowly and the green LED is off
-    HalWriteSMBusValue(0x20, SMC_REG_LEDSEQ, FALSE, 0xC0);
+    // XResetLED() can be used to restore system control of the front LEDs
 
     // If we quit, the LED might get reset, so we loop forever
     while(1) {


lib/hal/led.c
```diff
@@ -0,0 +1,15 @@
...
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x00);
+}
+
+void XSetManualLED(int t1, int t2, int t3, int t4) {
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+
+  int colorScheme = t1<<1;
+  colorScheme = (colorScheme|t2)<<1;
+  colorScheme = (colorScheme|t3)<<1;
+  colorScheme = (colorScheme|t4) & 0xFF;
```
JayFoxRox: This algorithm looks very broken.

Each LED color is 2 bits, not 1 bit. The & 0xFF is also strange.

It should probably be something like:

```c
//FIXME: make sure t1 - t4 are unsigned
assert(t1 <= 0x3);
assert(t2 <= 0x3);
assert(t3 <= 0x3);
assert(t4 <= 0x3);
uint8_t colorScheme = (t1 << 6) | (t2 << 4) | (t3 << 2) | (t4 << 0);
```

lib/hal/led.h
```diff
@@ -0,0 +1,18 @@
+#ifndef HAL_LED_H
+#define HAL_LED_H
+
+#include <hal/xbox.h>
+#include <xboxkrnl/xboxkrnl.h>
+
+#define LED_OFF    0x00
+#define LED_GREEN  0x01
+#define LED_RED    0x10
+#define LED_AMBER  0x11
```
JayFoxRox: Shouldn't these be binary?

```c
#define LED_OFF   0x0
#define LED_GREEN 0x1
#define LED_RED   0x2
#define LED_AMBER 0x3 //FIXME: Use ORANGE?
```

lib/hal/led.h
```diff
@@ -0,0 +1,18 @@
+#ifndef HAL_LED_H
+#define HAL_LED_H
+
+#include <hal/xbox.h>
+#include <xboxkrnl/xboxkrnl.h>
```
JayFoxRox: These don't seem to be used here

lib/hal/led.h
```diff
@@ -0,0 +1,18 @@
...
+
+#define LED_OFF    0x00
+#define LED_GREEN  0x01
+#define LED_RED    0x10
+#define LED_AMBER  0x11
+
+#define SMC_REG_LEDMODE 0x07
+#define SMC_REG_LEDSEQ  0x08
```
JayFoxRox: These should be in a different file - for now, I'd probably move them into the C file. We should expose them more generally from xboxkrnl.h or similar in the future (in a more complete set).

lib/hal/led.c
dracc: https://github.com/XboxDev/nxdk/blob/a647c6d072beba61ac41fcb25a4a26de0dbd74c9/samples/led/main.c#L12 would suggest that each color is 1 bit per time value but constantly repeating?

lib/hal/led.h
dracc: I opted for HEX for easy OR'ing.

lib/hal/led.h
dracc: `<hal/xbox.h>` is probably not used, `<xboxkrnl/xboxkrnl.h>` is used for `HalWriteSMBusValue()`.

lib/hal/led.h
dracc: Agreed. :+1:

lib/hal/led.h
JayFoxRox: Your old code was right, the LEDs are split into two different nibbles (one nibble for RED, one for GREEN).

lib/hal/led.c
JayFoxRox: Your old code is right, it's lower and upper nibble. The code you originally submitted should work.
It should probably `assert((t1 & ~0x11) == 0)` to make sure that only the useful bits are being used.

lib/hal/led.h
JayFoxRox: But not in the header. It should only be present in the source file.

*Otherwise you increase compilation time and clutter the namespace with typenames and functions which possibly conflict with others.*

dracc: Oh, TIL. :+1:

lib/hal/led.c
```diff
@@ -0,0 +1,19 @@
...
+}
+
+void XSetManualLED(unsigned int t1, unsigned int t2, unsigned int t3, unsigned int t4) {
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+
+  unsigned int colorScheme =
+    ((t1&LED_ORANGE)<<3) |
+    ((t2&LED_ORANGE)<<2) |
+    ((t3&LED_ORANGE)<<1) |
+    (t4&LED_ORANGE);
```
JayFoxRox: nit: you probably want to use `0x11` here, as `LED_ORANGE` is a bit misleading

JayFoxRox: nit: `t4` can be aligned (with spaces, or a `<< 0`)

lib/hal/led.h
```diff
@@ -0,0 +1,15 @@
+#ifndef HAL_LED_H
+#define HAL_LED_H
+
+#include <xboxkrnl/xboxkrnl.h>
+#include <assert.h>
```
JayFoxRox: `assert` is not used in the header. Only include what you use, and include it so it has the least side-effects possible.

lib/hal/led.c
```diff
@@ -0,0 +1,27 @@
+#include <hal/led.h>
```
JayFoxRox: As most of the existing HAL code is OpenXDK licensed, it might be a good idea to add a license header here (with a more suitable license).

*To avoid any license confusion.*

JayFoxRox:
Please update the existing LED sample with the new API.

lib/hal/led.c
```diff
@@ -0,0 +1,27 @@
...
+
+void XSetManualLED(unsigned int t1, unsigned int t2, unsigned int t3, unsigned int t4) {
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+
+  assert((t1 & ~0x11) == 0x00);
+  assert((t2 & ~0x11) == 0x00);
+  assert((t3 & ~0x11) == 0x00);
+  assert((t4 & ~0x11) == 0x00);
+
+  unsigned int colorScheme =
```
JayFoxRox: `HalWriteSMBusValue` takes a `ULONG`; so it's probably best to use ULONG here, to avoid type issues

lib/hal/led.c
```diff
@@ -0,0 +1,27 @@
...
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x00);
+}
+
+void XSetManualLED(unsigned int t1, unsigned int t2, unsigned int t3, unsigned int t4) {
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+
+  assert((t1 & ~0x11) == 0x00);
+  assert((t2 & ~0x11) == 0x00);
+  assert((t3 & ~0x11) == 0x00);
+  assert((t4 & ~0x11) == 0x00);
```
JayFoxRox: nit: I'd use `== 0` instead of `0x00` as the intention is specifically to check for 0 outside of the 0x11 pattern, so specifically `0x000000...` (of unknown size, hence: `0`)

dracc:
LED sample updated - became very short.

samples/led/main.c
```diff
@@ -1,19 +1,13 @@
...
 void main(void)
 {
-    // Set LED to be user controlled
-    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
-
     // LED pattern is:
-    //   Red:   0xC = 1100 (ON,  ON,  OFF, OFF;   ON,  ON,  OFF, OFF;   ...)
-    //   Green: 0x0 = 0000 (OFF, OFF, OFF, OFF;   OFF, OFF, OFF, OFF;   ...)
+    //   Red:   (ON,  ON,  OFF, OFF;   ON,  ON,  OFF, OFF;   ...)
+    //   Green: (OFF, OFF, OFF, OFF;   OFF, OFF, OFF, OFF;   ...)
```
JayFoxRox: These comments should be removed.

They serve no purpose here.

The comments could be moved into the LED API code, but they are of such poor quality that they are probably more confusing than doing any good.

The new API is pretty descriptive; the only comment that's necessary is that each of the states, from left-to-right, will be shown for X ms and then it will repeat after 4*X ms.

samples/led/main.c
```diff
@@ -1,19 +1,13 @@
...
     // This means the red LED will blink slowly and the green LED is off
-    HalWriteSMBusValue(0x20, SMC_REG_LEDSEQ, FALSE, 0xC0);
+    XSetManualLED(LED_RED, LED_RED, LED_OFF, LED_OFF);
```
JayFoxRox: nit: There should be a comment that `XSetAutoLED` can reset the LED to the kernel-controlled mode.

samples/led/main.c
```diff
@@ -1,19 +1,11 @@
...
 {
-    // Set LED to be user controlled
-    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
-
-    // LED pattern is:
-    //   Red:   0xC = 1100 (ON,  ON,  OFF, OFF;   ON,  ON,  OFF, OFF;   ...)
-    //   Green: 0x0 = 0000 (OFF, OFF, OFF, OFF;   OFF, OFF, OFF, OFF;   ...)
-    // This means the red LED will blink slowly and the green LED is off
-    HalWriteSMBusValue(0x20, SMC_REG_LEDSEQ, FALSE, 0xC0);
+    // Set front LEDs to shine red for 2*X ms every 4*X ms
```
JayFoxRox: "X" should obviously be roughly documented. I just used "X" in a previous review as I didn't remember the frequency.

samples/led/main.c
```diff
@@ -1,19 +1,11 @@
...
+    XSetManualLED(LED_RED, LED_RED, LED_OFF, LED_OFF);
+    // Use XSetAutoLED() to restore kernel control of the front LEDs
```
JayFoxRox: nit: empty line before this line to highlight this section.

JayFoxRox: nit: "SetAutoLED() can be used..". The word "can" highlights that this is an option, and not an implicit "FIXME"/WIP-part for the sample.

lib/hal/led.c
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright and related rights waived via CC0
```
JayFoxRox: nit: You'd normally still say who waves the right. See https://creativecommons.org/choose/zero/

The text can be trimmed but I assume you won't get around using your name?

JayFoxRox:
I prototyped the original code on Discord a while back, so I expect someone else to do final review / test this. @XboxDev

lib/hal/led.h
```diff
@@ -0,0 +1,12 @@
+#ifndef HAL_LED_H
+#define HAL_LED_H
+
+#define LED_OFF    0x00
+#define LED_GREEN  0x01
+#define LED_RED    0x10
+#define LED_ORANGE 0x11
+
+void XSetAutoLED(void);
+void XSetManualLED(unsigned int t1, unsigned int t2, unsigned int t3, unsigned int t4);
```
mborgerson: Please add a comment indicating what these parameters do: that t1-t4 correspond to the four LED states which are cycled repeatedly.

lib/hal/led.c
```diff
@@ -0,0 +1,32 @@
...
+
+#define SMC_REG_LEDMODE 0x07
+#define SMC_REG_LEDSEQ  0x08
+
+void XSetAutoLED(void) {
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x00);
+}
+
+void XSetManualLED(unsigned int t1, unsigned int t2, unsigned int t3, unsigned int t4) {
+  HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
```
mborgerson: Please add constant definitions for LED mode and use them here. ([See here](https://github.com/rkalz/xbmc4xbox/blob/50f38fb2fce74f2d7679ffc14f3d5002b9c65b6a/xbmc/xbox/XKUtils.h#L134))

lib/hal/led.c
```diff
@@ -0,0 +1,49 @@
...
+  return ret;
+}
+
+unsigned int XGetLEDColorScheme(void) {
+  ULONG ret = 0;
+  if (XGetLEDMode() == LED_MODE_MANUAL) {
+    HalReadSMBusValue(0x20, SMC_REG_LEDSEQ, FALSE, &ret);
+  }
+  return ret;
+}
```
JayFoxRox: This makes the API asymmetric.

I'd rather have it be `XGetManualLED(unsigned int *t1, [...])` and a return value wether it was successful.

For the auto/manual mode it can be a `bool XIsAutoLED()` (and maybe a negating `bool XIsManualLED() { return !XIsAutoLED(); }`).

Maybe someone else has design thoughts?

*I also don't think this "getter" is really necessary, as the application is normally under full control of LED (so it knows the state), and it can always fall back to lower-level APIs if it needs niche features.*

JayFoxRox: Other idea for API: `void XResetLED()` / `void XSetCustomLED(unsigned int t1, [...])` and `bool XGetCustomLED(unsigned int* t1, [...])` which returns an error to indicate LED wasn't even in custom mode.

So instead of naming 2 modes, just have 1 default mode and 1 named mode (custom).

samples/led/main.c
```diff
@@ -1,19 +1,12 @@
...
-
-#define SMC_REG_LEDMODE             0x07
-#define SMC_REG_LEDSEQ              0x08
 
 void main(void)
 {
-    // Set LED to be user controlled
-    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+    // Set front LEDs to shine red for 2*X ms every 4*X ms
+    XSetManualLED(LED_RED, LED_RED, LED_OFF, LED_OFF);
```
JayFoxRox: Let's make this `LED_GREEN, LED_RED, LED_OFF, LED_OFF` so we can see the order (green happens before red)?

lib/hal/led.h
```diff
@@ -0,0 +1,40 @@
+#ifndef HAL_LED_H
+#define HAL_LED_H
+
+#define LED_OFF    0x00
+#define LED_GREEN  0x01
+#define LED_RED    0x10
+#define LED_ORANGE 0x11
```
JayFoxRox: nit: Can use an `enum`? Not sure what other maintainers think

mborgerson: Yes, enum please.

lib/hal/led.c
mborgerson: Agree. Don't see much of a use case for this getter.

lib/hal/led.c
```diff
@@ -0,0 +1,54 @@
...
+  ULONG colorScheme =
+    ((t1 & 0x11) << 3) |
+    ((t2 & 0x11) << 2) |
+    ((t3 & 0x11) << 1) |
+    ((t4 & 0x11) << 0);
+
+  HalWriteSMBusValue(0x20, SMC_REG_LEDSEQ, FALSE, colorScheme);
+}
+
+bool XGetLEDColorScheme(unsigned int* t1, unsigned int* t2, unsigned int* t3, unsigned int* t4) {
```
mborgerson: Is this necessary?

samples/led/main.c
```diff
@@ -1,19 +1,12 @@
...
-#include <xboxkrnl/xboxkrnl.h>
-
-#define SMC_REG_LEDMODE             0x07
-#define SMC_REG_LEDSEQ              0x08
 
 void main(void)
 {
-    // Set LED to be user controlled
-    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+    // Set front LEDs to shine green for ~250 ms then red for ~250 ms roughly every second
```
JayFoxRox: nit: Someone should confirm the timing; someone might use these exact values for handling serial communication

JayFoxRox: A bit of a LED-timing-analysis (with cameras, oscilloscopes and logic analyzers) party started on Discord. - No consensus on timing was reached.

We should probably have https://xboxdevwiki.net/LED (or something) and link to that, to stop the bikeshed and prevent decentralizing information. Having a wiki article also leaves more room to explain how the LED behaves when opening the tray and so on.

We'll continue to bikeshed over this stuff for days otherwise.

lib/hal/led.h
```diff
@@ -0,0 +1,24 @@
...
+/*
+  Restore kernel LED control.
+*/
+void XResetLED(void);
+
+/*
+  Set the LED color manually
+  t1-t4 are used to set the color for each segment
+  in a four step loop. For more details about front LED characteristics,
+  please see https://www.XboxDevWiki.net/LED
```
JayFoxRox: remove www (breaks link for me)

JayFoxRox: nit: use lowercase xboxdevwiki (more suitable for domain name)

samples/led/main.c
```diff
@@ -1,22 +1,17 @@
...
-#define SMC_REG_LEDMODE             0x07
-#define SMC_REG_LEDSEQ              0x08
 
 void main(void)
 {
-    // Set LED to be user controlled
-    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+  // Set front LEDs to shine green and then red.
+  // For information about timings and such, see
+  // https://www.XboxDevWiki.net/LED
```
JayFoxRox: Same as above; however, I wouldn't even bother documenting this as detailed here.

People using the API will probably look this up themselves.

samples/led/main.c
```diff
@@ -1,22 +1,17 @@
...
+  XSetCustomLED(LED_OFF, LED_OFF, LED_OFF, LED_OFF);
 
-    // LED pattern is:
-    //   Red:   0xC = 1100 (ON,  ON,  OFF, OFF;   ON,  ON,  OFF, OFF;   ...)
-    //   Green: 0x0 = 0000 (OFF, OFF, OFF, OFF;   OFF, OFF, OFF, OFF;   ...)
-    // This means the red LED will blink slowly and the green LED is off
-    HalWriteSMBusValue(0x20, SMC_REG_LEDSEQ, FALSE, 0xC0);
+  // XResetLED() can be used to restore kernel control of the front LEDs
```
JayFoxRox: The kernel is never really in control of it; I'd call it `system control` probably (the SMC handles it, but I think putting SMC here is too cryptic)

lib/hal/led.h
JayFoxRox: Force @Ernegien to document his findings there.

samples/led/main.c
```diff
@@ -1,22 +1,15 @@
...
-
-#define SMC_REG_LEDMODE             0x07
-#define SMC_REG_LEDSEQ              0x08
 
 void main(void)
 {
-    // Set LED to be user controlled
-    HalWriteSMBusValue(0x20, SMC_REG_LEDMODE, FALSE, 0x01);
+  // Set front LEDs to shine green and then red.
+  XSetCustomLED(LED_OFF, LED_OFF, LED_OFF, LED_OFF);
```
JayFoxRox: Why is this OFF, OFF, OFF, OFF; if it's meant to be `green and then red`. Also maybe make clear that this is a repeating pattern.

dracc: You're right, I forgot I had made a change in order to help @Ernegien troubleshoot. Fixed!

dracc:
The changes requested by @mborgerson Have been taken care of, but seeing as I've force pushed my changes those have not been automatically resolved.

Is this PR ready for merge or is anyone expecting any further changes?

lib/hal/led.h
```diff
@@ -0,0 +1,24 @@
+#ifndef HAL_LED_H
+#define HAL_LED_H
+
+typedef enum _LEDColor {
+  LED_OFF = 0x00,
+  LED_GREEN = 0x01,
+  LED_RED = 0x10,
+  LED_ORANGE = 0x11
+} LEDColor;
```
JayFoxRox: nit: I wonder if we should prefix these somehow and if the style is right - it's part of the API, and once set, we shouldn't keep changing it.

JayFoxRox:
I haven't tested it, but the code looks fine. Can be merged.
This bikeshed can fit a jumbo-jet.

lib/hal/led.h
dracc: `FRONT_LED_XXX`?

PR #108, opened 2019-05-14
Move SDL Makefile into submodule

dracc:
As requested on Discord, this change was extracted from #106 as that one should focus on one thing instead of many.


thrimbor:
Thanks!

PR #109, opened 2019-05-16
xpad: Match interface instead of VID/PID

dracc:
As kindly explained by JFR, this change ought to enable most (if not all) Xbox compatible controllers instead of just the few identified earlier.

### DIFF
--- a/lib/usb/sys/xpad.c
+++ b/lib/usb/sys/xpad.c
@@ -13,6 +13,8 @@
 #include "../usb_wrapper.h"
 #include "usb/include/config.h"
 
+#include <xboxkrnl/xboxkrnl.h>
+
 // history for the Rising - falling events
 unsigned char xpad_button_history[7];
 
@@ -72,8 +74,8 @@ static void xpad_irq(struct urb *urb, struct pt_regs *regs)
 
 static int GetPadNum(const char* szDevPath)
 {
-	int pos			= strlen(szDevPath) - 3;
-	int portchar	= szDevPath[pos];
+	int pos = (XboxHardwareInfo.Flags & XBOX_HW_FLAG_INTERNAL_USB_HUB) ? 2 : 0;
+	int portchar = szDevPath[pos];
 
 	switch(portchar)
 	{
--- a/lib/usb/sys/xpad.c
+++ b/lib/usb/sys/xpad.c
@@ -156,37 +156,7 @@ static void xpad_disconnect(struct usb_interface *intf)
 }
 /*------------------------------------------------------------------------*/ 
 static struct usb_device_id xpad_ids [] = {
-	{ USB_DEVICE(0x044f, 0x0f07) },//Thrustmaster, Inc. Controller
-	{ USB_DEVICE(0x045e, 0x0202) },//Microsoft Xbox Controller
-	{ USB_DEVICE(0x045e, 0x0285) },//Microsoft Xbox Controller S
-	{ USB_DEVICE(0x045e, 0x0287) },//Microsoft Xbox Controller S
-	{ USB_DEVICE(0x045e, 0x0289) },//Microsoft Xbox Controller S
-	{ USB_DEVICE(0x046d, 0xca84) },//Logitech Cordless Controller for Xbox
-	{ USB_DEVICE(0x046d, 0xca88) },//Logitech Compact Controller for Xbox
-	{ USB_DEVICE(0x05fd, 0x1007) },//???Mad Catz Controller???
-	{ USB_DEVICE(0x05fd, 0x107a) },//InterAct PowerPad Pro
-	{ USB_DEVICE(0x0738, 0x4516) },//Mad Catz Control Pad
-	{ USB_DEVICE(0x0738, 0x4522) },//Mad Catz LumiCON
-	{ USB_DEVICE(0x0738, 0x4526) },//Mad Catz Control Pad Pro
-	{ USB_DEVICE(0x0738, 0x4536) },//Mad Catz MicroCON
-	{ USB_DEVICE(0x0738, 0x4540) },//Mad Catz Beat Pad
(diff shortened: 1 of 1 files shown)


lib/usb/sys/xpad.c
```diff
@@ -156,37 +156,7 @@ static void xpad_disconnect(struct usb_interface *intf)
...
-	{ USB_DEVICE(0x0e6f, 0x0005) },//Eclipse wireless Controlle
-	{ USB_DEVICE(0x0e6f, 0x0006) },//Edge wireless Controller
-	{ USB_DEVICE(0x0e8f, 0x0201) },//Smartjoy Frag PS2/Xpad adaptor
-	{ USB_DEVICE(0x0f30, 0x0202) },//Joytech Advanced Controller 
-	{ USB_DEVICE(0x0f30, 0x8888) },//BigBen XBMiniPad Controller
-	{ USB_DEVICE(0x102c, 0xff0c) },//Joytech Wireless Advanced Controller
-	{ USB_DEVICE(0x12ab, 0x8809) },//Xbox DDR dancepad 
-	{ USB_DEVICE(0xffff, 0xffff) },//Chinese-made Xbox Controller 
-	{ USB_DEVICE(0x0000, 0x0000) }, // nothing detected - FAIL
+	{ USB_INTERFACE_INFO(88, 66, 0) }, /* Xbox Controller */
```
JayFoxRox: This isn't just `Xbox Controller`, but `Xbox Controller Interfaces`

JayFoxRox:
Did you test this code?

It would be great if as many people as possible tested this (with different controllers / device classes).

lib/usb/sys/xpad.c
dracc: Fixed, thanks!

dracc:
I have only tested with a keyboard through XQEMU, works as expected there.

JayFoxRox:
Should be tested with hot-plugging of real devices (on a real console), as I had this other issue with xpad where it would fail hotplugging (devicepath being short, padnum not being recognized). It's possible that some other code is used to recognize xpad if already connected at startup.

dracc:
This PR is "easily" tested by using my xpadprint branch. https://github.com/dracc/nxdk/tree/xpadprint

JayFoxRox:
I've tested this on my Xbox 1.1; I got the following `szDevPath` for `GetPadNum` (where X is the corresponding USB port digit):

- Original Controller S: Detected, Path "X.1"
- bigben XBOX JOYPAD: Detected, Path "X.1"
- Fanatec Speedser 3: Detected, Path "X.1"
- EMS TopGun II (Xbox mode): Detected, Path "X" **(Broken, index is -2)**
- EMS TopGun II (PC mode): Not detected (Expected)
- Xbox 360 controller: Not detected (Expected)

I'd be interested in more notes by others (other controllers / reports on Xbox 1.0) so we can figure out how to locate gamepads.

To me, it looks like we should be using `path[0]` (which is the root port).
However, on Xbox 1.0, with the USB hack, it seems to prepend "1." for the additional hub. So we have to manually detect Xbox 1.0, and skip over that internal hub (by doing `+= 2`on the char index): See `XBOX_HW_FLAG_INTERNAL_USB_HUB` in `XboxHardwareInfo.Flags` (kernel export).

*Specifically, we should not be doing the lookup backwards, from the end of the path (which would fail horribly when plugging hubs into hubs).*

Other than this location issue, the new method is a lot better than the old one. The old code did not accept some non-Microsoft gamepads for me (the licensed Fanatac Speedster 3 and the unlicensed EMS TopGun II). I also got no false positives yet with the new code, whereas catch-alls like 0xffff / 0xffff and 0x0000 / 0x0000 in the old code look scary.

---

The debug printing code in xpad.c was changed in my test, to avoid a crash before the first message:

```c
int pos = strlen(szDevPath) - 3;
debugPrint("Controller: szDevPath: %s, stringPos: %i\n", szDevPath, pos);

int portchar = szDevPath[pos]; // Might crash, if index is negative
debugPrint("... portchar: '%c'\n", portchar);
```

---

I had another (independent) issue, where the linked branch wouldn't work. The issue is that the USB interrupt is apparently not triggered when connecting a gamepad. I manually hacked this mainloop together for my tests:

```c
XInput_Init(); // Initialize USB with interrupts (broken)

while(1) {
  debugPrint("Alive!   "); // Show me that the Xbox didn't crash
  USBGetEvents(); // Find the new devices, because we get no interrupt for that
  XSleep(3000); // Give the Xbox 3 seconds to call interrupt on new device plug
}
```

dracc:
@GXTX has helped doing some testing, turns out the DVD dongle is detected with this change, yielding `szDevPath: 2` when plugged in port 4.

lib/usb/sys/xpad.c
```diff
@@ -72,8 +74,8 @@ static void xpad_irq(struct urb *urb, struct pt_regs *regs)
 
 static int GetPadNum(const char* szDevPath)
 {
-	int pos			= strlen(szDevPath) - 3;
-	int portchar	= szDevPath[pos];
+	int pos = (XboxHardwareInfo.Flags & XBOX_HW_FLAG_INTERNAL_USB_HUB) ? 2 : 0;
```
JayFoxRox: This should be in common USB code; either as a function, or as modification to all incoming device paths (which might prevent detection of non-connected USB ports in Xbox 1.0; a very niche use-case).

JayFoxRox:
We could hack the offset change into this:

https://github.com/XboxDev/nxdk/blob/77b5de45f0c64e70f2ff68248873448d5edccc71/lib/usb/core/hub.c#L975

However, I think keeping it as part of the gamepad is fine for now.

The IR dongle being detected is a bit annoying, but not too bad either.

---

**I think this is ready for merge. Waiting for someone to do final testing on Xbox 1.0 and non-1.0.**

PR #110, opened 2019-05-16
fp20compiler: Fix corrupted output on Windows

thrimbor:
Fix for corruption at the end of fp20compiler's output when running under Windows (as ranted about on Discord). The changes are:
1. Open input stream as binary to avoid line ending conversion issues.
2. Reserve an additional byte to make space for a null-terminator.
3. Add a null-terminator after reading the input.

Tested and working on msys2 on Windows 10.

### DIFF
--- a/tools/fp20compiler/main.cpp
+++ b/tools/fp20compiler/main.cpp
@@ -66,7 +66,7 @@ int main(int argc, char** argv) {
         exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


mborgerson:
Thanks!

Nit: this PR is not strictly a fix just for Windows, but rather a bug fix that corrects a failure to null-terminate the file contents string after reading.

PR #111, opened 2019-05-16
build: Fix debug option on Windows

thrimbor:
This simply changes the debug parameter from `/debug` to `-debug` which makes it work on Windows.

Tested and working on msys2 on Windows 10.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ NXDK_CXXFLAGS = $(NXDK_CFLAGS)
 ifeq ($(DEBUG),y)
 NXDK_CFLAGS += -g
 NXDK_CXXFLAGS += -g
-LDFLAGS += /debug
+LDFLAGS += -debug
 endif
 
 NXDK_CFLAGS += $(CFLAGS)


mborgerson:
> Which makes it work on Windows
Were other platforms tested?  Are the flags consistent?

thrimbor:
I tested on Arch afterwards and noticed no difference (same file size, same behavior). `/` and `-` behave the same on Linux (`/` is there for compatibility with Microsoft's link.exe), but `/` is problematic on msys.

JayFoxRox:
I'm surprised this is an issue in MSYS, do we know why that is?

I'm even surprised we don't use `/` everywhere.

thrimbor:
@JayFoxRox tbh I don't know enough about MSYS specialties to have an explanation for this. I can only say that this isn't the only weirdness I found in Windows ports of popular Linux/UNIX tools (like git-bash sometimes swallowing whitespaces unless they're escaped).

JayFoxRox:
Merged, but we should investigate this in the future probably.

Reaching MSVC (LINK.EXE) compatibility in the future might be beneficial (and I'm not sure if it supports dashes instead of slashes).

PR #112, opened 2019-05-16
Add WinAPI file finding

thrimbor:
This is meant to replace #93. Instead of fixing the X*-functions, I implemented the functionality from scratch using the WinAPI.

I intended to go for a more XDK-like behavior for drive mounting etc. (we had a discussion about this on Discord), but since that would take too long considering the current state is broken, this currently re-uses XConvertDOSFilenameToXBOX.

~~/edit: There is a rough sample demonstrating it here: https://github.com/thrimbor/nxdk/commit/c278467bd3cb0b3847b004d34e2fa906a59a07f6~~
Sample is now included.

/edit: Screenshot of the sample running in xqemu:
Screenshot_20190517_081357


thrimbor:
Sample added as requested by @JayFoxRox.

samples/winapi_filefind/main.c
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    pb_show_debug_screen();
+
+    debugPrint("Content of C:\\\n");
+
+    WIN32_FIND_DATA findFileData;
+    HANDLE hFind;
+
+    hFind = FindFirstFile("C:\\*.*", &findFileData);
```
JayFoxRox: Why is this `\\*.*` and not just `\\*`? Will this also find files without a dot in their name?

samples/winapi_filefind/main.c
```diff
@@ -0,0 +1,55 @@
+#include <stdio.h>
+#include <fileapi.h>
```
JayFoxRox: Is fileapi.h a standard Windows name?

We should attempt to reach more XDK / Windows code compatibility in the future.

*I know that the XDK often uses a catchall `xbox.h`; which would probably be recreated in include/xdk/xbox.h in the future.*

samples/winapi_filefind/main.c
thrimbor: `\\*.*` and `\\*` behave the same, both will find all files, independent of whether they contain a dot or not. My implementation of FindFirstFile() contains [code](https://github.com/XboxDev/nxdk/pull/112/commits/551a39ae4234d7cff5d937f14af17029f36ecda2#diff-71556cfa5b0777b50584eaf8d9d2223dR70) to explicitly handle the `\\*.*` special case.
My intention for using `\\*.*` was to demonstrate that it works just like it does on Windows (I recall seeing several samples for file finding using that).

samples/winapi_filefind/main.c
thrimbor: Yes, according to [Microsoft's online docs](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-findfirstfilea), fileapi.h is the correct header name. I tried to put everything exactly where it belongs to ensure maximum compatibility with Windows (using Microsoft and ReactOS/wine docs as a reference).

Microsoft does recommend to just include windows.h though, imho we should definitely add that header in the future.

samples/winapi_filefind/main.c
JayFoxRox: Interesting; You should add a comment here (about it not caring about the dot) and a comment that explains this "hack" in the linked mask-handler code.

- Did you verify this actually happens on Windows? `*.*` *really* listing files without dot?
- Are there other known differences to Windows / XDK FindFile you are aware of? What about `?` wildcard?

samples/winapi_filefind/main.c
JayFoxRox: Excellent! Thanks for confirming.

samples/winapi_filefind/main.c
thrimbor: I believe I did check it before, but I just ran a small sample in my Windows 10 VM to be sure, and I can confirm that `*.*` works as described and identical to `*`.
Tbh I wasn't aware of the `?` wildcard, but I'll try to write some more test code covering that later today.
I'll push some changes including the requested comments in a few minutes.

lib/winapi/findfile.c
```diff
@@ -0,0 +1,129 @@
...
+
+    if (dirPath.Length == 0 || mask.Length == 0) {
+        SetLastError(ERROR_INVALID_PARAMETER);
+        return INVALID_HANDLE_VALUE;
+    }
+
+    // Ensure that "*.*" lists all files, just like on Windows
+    if (mask.Length == 3 && strncmp(mask.Buffer, "*.*", 3)) {
+        mask.Length = 0;
+    }
```
mborgerson: In comments below there is discussion about `*.*` vs `*`, but I don't see the latter case handled here. Should this be handled as well?

JayFoxRox: The code implies that "*" as mask is handled the same way on Windows and Xbox.
This bit of code only compensates for the differences.

It should still be tested though. Also see https://github.com/XboxDev/nxdk/pull/112/files/f44a0316dcc43a4d43b282d440ea86ffe394c9fd#r284989562

lib/winapi/findfile.c
```diff
@@ -0,0 +1,129 @@
...
+    lpFindFileData->ftCreationTime.dwHighDateTime = dirInfo->CreationTime.HighPart;
+    lpFindFileData->ftLastAccessTime.dwLowDateTime  = dirInfo->LastAccessTime.LowPart;
+    lpFindFileData->ftLastAccessTime.dwHighDateTime = dirInfo->LastAccessTime.HighPart;
+    lpFindFileData->ftLastWriteTime.dwLowDateTime  = dirInfo->LastWriteTime.LowPart;
+    lpFindFileData->ftLastWriteTime.dwHighDateTime = dirInfo->LastWriteTime.HighPart;
+    lpFindFileData->nFileSizeHigh = dirInfo->EndOfFile.HighPart;
+    lpFindFileData->nFileSizeLow = dirInfo->EndOfFile.LowPart;
+    memcpy(lpFindFileData->cFileName, dirInfo->FileName, dirInfo->FileNameLength);
+    lpFindFileData->cFileName[dirInfo->FileNameLength] = '\0';
+    lpFindFileData->cAlternateFileName[0] = '\0';
```
mborgerson: Should this `cAlternateFileName` field be populated with the 8.3 name for compatibility?

JayFoxRox: No, not done in XAPI either.

lib/winapi/findfile.c
```diff
@@ -0,0 +1,129 @@
...
+    lpFindFileData->ftCreationTime.dwLowDateTime = dirInfo->CreationTime.LowPart;
+    lpFindFileData->ftCreationTime.dwHighDateTime = dirInfo->CreationTime.HighPart;
+    lpFindFileData->ftLastAccessTime.dwLowDateTime  = dirInfo->LastAccessTime.LowPart;
+    lpFindFileData->ftLastAccessTime.dwHighDateTime = dirInfo->LastAccessTime.HighPart;
+    lpFindFileData->ftLastWriteTime.dwLowDateTime  = dirInfo->LastWriteTime.LowPart;
+    lpFindFileData->ftLastWriteTime.dwHighDateTime = dirInfo->LastWriteTime.HighPart;
+    lpFindFileData->nFileSizeHigh = dirInfo->EndOfFile.HighPart;
+    lpFindFileData->nFileSizeLow = dirInfo->EndOfFile.LowPart;
+    memcpy(lpFindFileData->cFileName, dirInfo->FileName, dirInfo->FileNameLength);
+    lpFindFileData->cFileName[dirInfo->FileNameLength] = '\0';
```
mborgerson: Is it possible for `dirInfo->FileNameLength == MAX_PATH` potentially overrunning `cFileName` (buffer of size `MAX_PATH`) or will it always be strictly `< MAX_PATH`?

lib/winapi/findfile.c
mborgerson: How does the code compensate for the difference? It looks like if it matches `*.*` it removes the mask entirely, while `*` will make it through? Should `mask.Length` be set to 1, instead of 0, so it matches `*` instead?

JayFoxRox: This is how XAPI also seems to handle it.

I assume it could also be set to `mask.Length = 1`, but I think setting it to zero better highlights this special case.

samples/winapi_filefind/main.c
JayFoxRox: I checked XAPI disassembly earlier, and the check you do is the only one I could spot.
They do a `memcmp` instead of `strncmp` though (so it won't check for a `\0`; it checks the length for 3, then assumes the string is 3 symbols that match exactly).

lib/winapi/findfile.c
```diff
@@ -0,0 +1,129 @@
...
+{
+    NTSTATUS status;
+    ANSI_STRING dirPath;
+    ANSI_STRING mask;
+    IO_STATUS_BLOCK ioStatusBlock;
+    OBJECT_ATTRIBUTES attributes;
+    struct FileInfo fileInformation;
+    HANDLE handle;
+    size_t maskOffset;
+    char fullPath[MAX_PATH+29];
```
JayFoxRox: Where does this 29 come from?

lib/winapi/findfile.c
thrimbor: It was possible, but I just fixed it. The buffer provided by `struct FileInfo` is now `MAX_PATH-1`, which makes sure we don't overflow the output buffer.
I had a similar error in the assertion, which should check `< MAX_PATH` and not `<= MAX_PATH`.

lib/winapi/findfile.c
thrimbor: 29 is the maximum amount of characters added by `XConvertDOSFilenameToXBOX`. It's added here to make sure the buffer is always large enough.

samples/winapi_filefind/main.c
thrimbor: I swapped `strncmp` for `memcmp` and wrote a bit of test code to check `?` wildcard behavior, and from what I can tell, it seems to work correctly: https://gist.github.com/thrimbor/34f1914f9c62f5b584bc382605d854a6

lib/winapi/findfile.c
JayFoxRox: That's a bit strange (I'd rather add a safer value, like 64). But okay.

I assume `XConvertDOSFilenameToXBOX` will be gone soon enough anyway.

lib/winapi/findfile.c
```diff
@@ -0,0 +1,131 @@
...
+#include <stdbool.h>
+#include <winbase.h>
+#include <xboxkrnl/xboxkrnl.h>
+#include <hal/fileio.h>
+#include <hal/winerror.h>
+#include <fileapi.h>
+
+struct FileInfo {
+    FILE_DIRECTORY_INFORMATION dirInfo;
+    // Reserve path buffer minus the null-terminator and the first byte provided by dirInfo
```
JayFoxRox: Huh?! Why is it even `[1]` in dirinfo? is `[]` as trailing member an extension?

JayFoxRox: Nevermind, this is how MSDN defines it.. - but then I wonder how / why it should be used in the real world. Our current solution looks very hacky.

JayFoxRox: Are you sure that this doesn't have to include the zero terminator though? As it has a dedicated `dirInfo->FileNameLength` field the source string doesn't have to be zero-terminated. So this field probably needs room for the additional zero termination.

lib/winapi/findfile.c
JayFoxRox: I was very confused by this comment at first, but it all makes sense. Thanks :+1:

JayFoxRox:
Works for me - reviewed code a while ago, and also ran on my Xbox. Output of sample looks fine.

As this hasn't gotten an update (is it considered ready for merge?), I'll delay merge until the next time I skim over open PRs; so if anybody has a different opinion they should veto as soon as possible. - if someone also accepts this, feel free to merge yourself.

Issue #113, opened 2019-05-17
fp20compiler: Split by shader-magic ("!!") instead of comment


JayFoxRox:
fp20compiler currently splits 2 shaders (Texture Shaders + Register Combiners) in the same file by looking for a comment (that was placed by `cg`):

https://github.com/XboxDev/nxdk/blob/82e4860c2f5a358677f60000e40f8f00529e5e8e/tools/fp20compiler/main.cpp#L44

Instead, we should be searching for each `!!` in the file, and then ensure it's either at offset 0 in the string (`!!` in first line), or it's following a `\n` (`\n!!`, so `!!` at start of any other line).

We should iterate over each segment started by a magic, so we can translate each shader and report errors for unknown types.

That way, we'd make it easier for users to write their own shaders by hand, without having to remember cryptic comments.

*Note: I did not check if the OpenGL spec has a better suggestion for keeping different programs in the same string. That might impact this design.*

PR #115, opened 2019-05-18
fp20compiler: Refactor translation function

JayFoxRox:
Implements what I proposed in #113 

However, while working on this, I realized this might be overkill.
It's unlikely we'll have more shader types per file.

I did not test this for long, but the output from `cg` was still parsed correctly from what I can tell.

### DIFF
--- a/tools/fp20compiler/main.cpp
+++ b/tools/fp20compiler/main.cpp
@@ -34,27 +34,68 @@ nvparse_errors errors;
 int line_number;
 char * myin = 0;
 
+static char* find_at_line_start(char* haystack, char* cursor, const char* needle) {
+    cursor = strstr(cursor, needle);
+    while (cursor != NULL) {
+
+        // Accept if candidate is at beginning of file or line
+        if ((cursor == haystack) || (cursor[-1] == '\n')) {
+            return cursor;
+        }
+
+        cursor = strstr(cursor + 1, needle);
+    }
+    return NULL;
+}
+
 void translate(const char* s) {
     char* ns = strdup(s);
 
-    // printf("-- %s\b", s);
+    // Look for the first shader magic
+    char* shader_start = find_at_line_start(ns, ns, "!!");
 
-    char* ts_start = strstr(ns, "!!TS1.0");
-    if (ts_start) {
-        char* ts_end = strstr(ts_start, "// End of program");
-        assert(ts_end);
-        *ts_end = 0;
-
-        ts10_init(ts_start);
-        ts10_parse();
-
-        *ts_end = '/';
+    // Warn the user if we couldn't find any shader at all
+    if (shader_start == NULL) {
+        fprintf(stderr, "no shaders found\n");
     }
 
-    char* rc_start = strstr(ns, "!!RC1.0");
-    if (rc_start) {
-        rc10_init(rc_start);
-        rc10_parse();
+    // Loop until we can't find a shader anymore
+    while (shader_start != NULL) {
+
+        // Find end of shader magic line
+        char* line_end = strchr(shader_start, '\n');
+
+        // The next shader magic marks the end of the current shader; find it
+        char* next_shader_start;
+        if (line_end == NULL) {
+            next_shader_start = NULL;
+        } else {
+            next_shader_start = find_at_line_start(ns, line_end + 1, "!!");
+        }
+
+        // If another shader exists, modify it to end current shader string
+        if (next_shader_start) { *next_shader_start = '\0'; }
+
+        // Process shader
+        if (is_ts10(shader_start)) {
+            ts10_init(shader_start);
+            ts10_parse();
+        } else if (is_rc10(shader_start)) {
+            rc10_init(shader_start);
+            rc10_parse();
+        } else {
+            if (line_end) { *line_end = '\0'; }
+            fprintf(stderr, "unknown shader type \"%s\"\n", shader_start);
+            if (line_end) { *line_end = '\n'; }
+        }
+
+        // Recover next magic (was modified to mark end of current shader)
+        if (next_shader_start) { *next_shader_start = '!'; }
+
+
+        // Continue with next shader
+        shader_start = next_shader_start;
+
     }
 
     free(ns);


dracc:
Any suggestions on how to test this? :sweat_smile:

JayFoxRox:
See #113 for which use-case is possible now.

To test, just write shaders which contain a texture shader and a register combiner. Possibly also test the case where only one shader is present / no shader is present.
Make sure existing auto-generated shaders (from Cg compiler) still work.

This is mostly a refactor and cleanup, hence I consider it "overkill": there's simpler options to achieve the same goal (like modifying the existing search for "// End of program" to search for "\n!!"; then using `ts_end` as cursor).

tools/fp20compiler/main.cpp
```diff
@@ -37,25 +37,69 @@ char * myin = 0;
...
-    char* rc_start = strstr(ns, "!!RC1.0");
-    if (rc_start) {
-        rc10_init(rc_start);
-        rc10_parse();
-    }
+    // Move to start of file
+    char* cursor = ns;
+
+    // Remember where the current shader starts
+    char* shader_start = NULL;
```
JayFoxRox: This is misleading. It should say that we initially don't know where the current shader starts.

tools/fp20compiler/main.cpp
```diff
@@ -37,25 +37,69 @@ char * myin = 0;
...
+
+    // Loop over each shader (between two shader start tokens)
+    do {
+
+        // Try to find a shader start token candidate to mark the end
+        char* shader_end = strstr(cursor, "!!");
```
JayFoxRox: This is misleading. It doesn't search a `shader_end`, but `next_shader_start`.
It's used as `shader_start` or `shader_end` depending on the current state:
- If we don't have a start for the current shader, the next shader start marks the start
- If we already have a start for the current shader, the next shader start marks the end

tools/fp20compiler/main.cpp
```diff
@@ -37,25 +37,69 @@ char * myin = 0;
...
+        // Make sure we found a token candidate
+        if (shader_end != NULL) {
+
+            // Only accept token at beginning of file, or beginning of line
+            if ((shader_end != ns) && (shader_end[-1] != '\n')) {
+                cursor = shader_end + 1;
+                continue;
+            }
+
+        }
```
JayFoxRox: nit: This should probably have been a function: `shader_end = find_line_start(ns, cursor, "!!");`

tools/fp20compiler/main.cpp
```diff
@@ -37,25 +37,69 @@ char * myin = 0;
...
+            // If we only found one token, use it as start + find another one
+            shader_start = shader_end;
+            cursor = shader_start + 1;
+            continue;
+        }
+
+        // Mark end of string at next token (if it exists)
+        if (shader_end) { *shader_end = '\0'; }
+
+        // Process shader
```
JayFoxRox: This should add a comment so the user knows what generated the respective pbkit commands.

tools/fp20compiler/main.cpp
JayFoxRox: Fixed. I rewrote the algorithm differently.

tools/fp20compiler/main.cpp
JayFoxRox: I've done this now. The function name is slightly confusing, but still okay.

tools/fp20compiler/main.cpp
JayFoxRox: I've added a new commit which does this.

tools/fp20compiler/main.cpp
JayFoxRox: I've rewritten the function by moving the search for the first shader out of the loop. It has more redundancy, but is easier to understand.

JayFoxRox:
I've done a major refactor of this. I didn't test this very much, but it should still work.
The code should be easier to understand now.

I've also added line counting as I plan to use it when fixing #189.

tools/fp20compiler/main.cpp
```diff
@@ -34,27 +34,63 @@ nvparse_errors errors;
...
     char* ns = strdup(s);
 
-    // printf("-- %s\b", s);
+    // Look for the first shader magic
+    char* shader_start = find_line_start(ns, ns, "!!");
+
+    // If we couldn't find any shader at all, we have to give up
+    if (shader_start == NULL) {
+        fprintf(stderr, "no shaders found\n");
+        return;
```
JayFoxRox: This path doesn't free `ns`.

tools/fp20compiler/main.cpp
```diff
@@ -34,27 +34,80 @@ nvparse_errors errors;
 int line_number;
 char * myin = 0;
 
+static char* find_at_line_start(char* haystack, char* cursor, const char* needle) {
+    cursor = strstr(cursor, needle);
+    while(cursor != NULL) {
```
dracc: No space between `while` and `(`, not sure how I feel about that.

tools/fp20compiler/main.cpp
```diff
@@ -34,27 +34,80 @@ nvparse_errors errors;
...
-        rc10_parse();
+    // Loop until we can't find a shader anymore
+    while(shader_start != NULL) {
+
+        // The next shader magic marks the end of the current shader; find it
+        char* line_end = strchr(shader_start, '\n');
+        char* next_shader_start = find_at_line_start(ns, line_end + 1, "!!");
+
+        // Create a copy of the current shader magic
+        if (line_end) { *line_end = '\0'; }
```
dracc: I'm no fan of one-liner `if`s.

dracc:
Logic looks good, I'm not sure about the code style though.

tools/fp20compiler/main.cpp
JayFoxRox: Me neither, but this is quick back-and-forth patching for `strdup`, so I think it's fine.
Willing to change if someone else has a strong opinion on this.

*I think what's worse, is that this isn't described in a comment, when the same change is documented below, when modifying `*next_shader_start`.*

tools/fp20compiler/main.cpp
JayFoxRox: I've changed this to make it more consistent with the rest of main.cpp and most of the nvidia code.

tools/fp20compiler/main.cpp
thrimbor: Couldn't this patching be replaced with `strndup`?

tools/fp20compiler/main.cpp
```diff
@@ -66,6 +70,20 @@ void translate(const char* s) {
...
+        char* shader_magic = strdup(shader_start);
+        if (line_end) { *line_end = '\n'; }
+
+        // Count which line the current shader starts on
+        while (line_cursor < shader_start) {
+            line_cursor = strchr(line_cursor + 1, '\n');
+            line++;
+        }
```
thrimbor: Is `line` meant to specify the line number of the magic or the start of the actual code? It currently does the latter.

JayFoxRox: The shader magic is part of the shader code already.

`line` is meant to be the line which contains the shader magic.
I believe this worked fine for me?

tools/fp20compiler/main.cpp
JayFoxRox: Yes, sort-of.

I'd still have to differentiate the case where no line-ending exists (it implies end of line is at file ending instead). That makes the code slightly awkward, too.

However, more importantly, the old code already copies all input data for `ns`, to allow patching.
The same style of patching is also done for `next_shader_start` below, which can not be avoided without copying large chunks of input data again.
The patching for `line_end` is thereby consistent with the rest of the code.

However, I also agree that this patching isn't ideal. It might be worth to consider this for a future refactor though (remove `ns` and `strndup` each individual shader / magic to avoid patching).

tools/fp20compiler/main.cpp
thrimbor: It seems to be off by one - when testing with the triangle sample, I get [this cgc output](https://pastebin.com/3QrbdtRW), and [this](https://pastebin.com/6qFjpPgi) is the end result, which lists the number of the first line after the magic.

JayFoxRox: There was an off-by-one. I had to change the algorithm slightly and added an `assert` for safety.

JayFoxRox:
I added a new commit which disables `\r` to appear in the input file.

- The yacc parser has explicit support for this, but somehow it starts reading from stdin.
- My additions in this PR would mess up the shader magic comment if an `\r` is also in that line.

The easiest solution was to reject such input for now.
I'll create an issue about this not being supported.

tools/fp20compiler/main.cpp
```diff
@@ -34,27 +34,82 @@ nvparse_errors errors;
...
+        // Count which line the current shader starts on
+        while (line_cursor < shader_start) {
+            line_cursor = strchr(line_cursor, '\n');
+            assert(line_cursor != NULL);
+            line_cursor++;
+            line++;
+        }
+
+        // Add information about shader section to output
+        printf("/* %s (line %u) */\n", shader_magic, line);
```
JayFoxRox: nit: This can break if the input file has something like: `!!RC1.0 #*/ C-code /*`
I don't consider this a deal-breaker though, as it's unlikely to have such things in the first line.

I do not intend to fix / prevent this for now.

tools/fp20compiler/main.cpp
```diff
@@ -34,27 +34,82 @@ nvparse_errors errors;
...
-    char* rc_start = strstr(ns, "!!RC1.0");
-    if (rc_start) {
-        rc10_init(rc_start);
-        rc10_parse();
+    // Loop until we can't find a shader anymore
+    while (shader_start != NULL) {
+
+        // The next shader magic marks the end of the current shader; find it
+        char* line_end = strchr(shader_start, '\n');
+        char* next_shader_start = find_at_line_start(ns, line_end + 1, "!!");
```
JayFoxRox: `line_end` could be `NULL` at this point. `next_shader_start` must be `NULL` then.
A check has to be added.

Alternatively, we could report an error for this case.
However, I'd like to avoid that for now (as nvparse might also report this as an error - I'm not sure yet).

JayFoxRox: I've fixed this and tested with `echo -n '!!RC1.0' > test.ps.asm`

tools/fp20compiler/main.cpp
```diff
@@ -84,6 +146,11 @@ int main(int argc, char** argv) {
     fread(buffer, size, 1, fh);
     buffer[size] = '\0';
 
+    if (strchr(buffer, '\r') != NULL) {
+        fprintf(stderr, "found unsupported line-endings\n");
+        exit(1);
+    }
```
JayFoxRox: The commit "fp20compiler: Reject input with unsupported line-endings" triggers on Windows CI: https://travis-ci.org/XboxDev/nxdk/jobs/601952923

This implies that cgc on Windows introduces `\r\n` line-endings.

However, if I manually create a file with those line-endings on Linux, then fp20compiler will get stuck waiting for input via `stdin` (which should never happen). I assume that Windows would do the same (why doesn't it?).

Additionally, the code added in "fp20compiler: Annotate shader sections in output" does not support those line-endings.
As I have no meaningful way to test on Windows at the moment, I'll have to remove that feature from this PR.

I'll split this PR into 2 PRs so only "fp20compiler: Refactor translation function" remains for now (which should be able to handle `\r\n` just fine).
I'll separately PR the other commits, once I'm able to have them working on all platforms.

tools/fp20compiler/main.cpp
JayFoxRox: This commit was removed, see https://github.com/XboxDev/nxdk/pull/115#discussion_r338232789.

JayFoxRox:
I had to remove 2 commits.

The commit "fp20compiler: Reject input with unsupported line-endings" triggered on Windows CI: https://travis-ci.org/XboxDev/nxdk/jobs/601952923

This implies that cgc on Windows introduces `\r\n` line-endings.

However, if I manually create a file with those line-endings on Linux, then fp20compiler will get stuck waiting for input via `stdin` (which should never happen). I assume that Windows would do the same (why doesn't it?).

Additionally, the code added in "fp20compiler: Annotate shader sections in output" did not support those line-endings.
As I can't just reject those files and I have no meaningful way to test on Windows at the moment, I'll have to remove that feature from this PR.

I'll split this PR into 2 PRs so only "fp20compiler: Refactor translation function" remains for now (which should be able to handle `\r\n` just fine).
I'll separately PR the other commits, once I'm able to have them working on all platforms.

tools/fp20compiler/main.cpp
```diff
@@ -34,27 +34,68 @@ nvparse_errors errors;
...
+        // Process shader
+        if (is_ts10(shader_start)) {
+            ts10_init(shader_start);
+            ts10_parse();
+        } else if (is_rc10(shader_start)) {
+            rc10_init(shader_start);
+            rc10_parse();
+        } else {
+            if (line_end) { *line_end = '\0'; }
+            fprintf(stderr, "unknown shader type \"%s\"\n", shader_start);
```
JayFoxRox: This won't display a meaningful message if `\r` is included (such as windows-line endings).

I consider this a minor issue. We can create an issue after merge.
I finally want this change merged so I can continue with actual work in fp20compiler and XQEMU.

JayFoxRox:
I'm closing this to do a full refactor (which I have almost completed).
It will use `strndup` to copy individual shaders, and also handle `\r` and other niche cases correctly.

I'll probably send a new PR soon.

PR #120, opened 2019-05-19
Support OutputDebugString

JayFoxRox:
Very useful function for getting data to a debugger.

### DIFF
--- a/lib/winapi/debug.c
+++ b/lib/winapi/debug.c
@@ -1,0 +1,17 @@
+#include <winbase.h>
+#include <winnt.h>
+
+#include <string.h>
+
+void WINAPI OutputDebugStringA (LPCTSTR lpOutputString)
+{
+    ANSI_STRING s;
+
+    s.Buffer = (LPSTR)lpOutputString;
+    s.Length = (USHORT)strlen(s.Buffer);
+    s.MaximumLength = s.Length + 1;
+
+    __asm__ __volatile__("mov $1, %%eax\n" // $1 = BREAKPOINT_PRINT
+                         "int $0x2D\n"
+                         "int $3\n": :"c" (&s));
+}
--- a/lib/winapi/winbase.h
+++ b/lib/winapi/winbase.h
@@ -3,6 +3,7 @@
 
 #include <windef.h>
 #include <minwinbase.h>
+#include <winnt.h>
 
 #ifdef __cplusplus
 extern "C"
@@ -32,6 +33,14 @@ typedef struct _WIN32_FIND_DATAA {
 
 DWORD GetLastError (void);
 void SetLastError (DWORD error);
+
+void WINAPI OutputDebugStringA (LPCTSTR lpOutputString);
+
+#ifndef UNICODE
+#define OutputDebugString OutputDebugStringA
+#else
+#error nxdk does not support the Unicode API
+#endif
 
 #ifdef __cplusplus
 }
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -133,12 +133,7 @@ typedef enum _MODE {
 /**
  * This struct defines a counted string used for ANSI-strings
  */
-typedef struct _ANSI_STRING
-{
-    USHORT Length;        /**< Length in bytes of the string stored in the buffer */
-    USHORT MaximumLength; /**< Length in bytes of the buffer */
-    PSTR Buffer;          /**< Pointer to the buffer used for the character-string */
-} ANSI_STRING, *PANSI_STRING, STRING, *PSTRING, OBJECT_STRING, *POBJECT_STRING;
+typedef STRING OBJECT_STRING, *POBJECT_STRING;
 
 /**
  * This struct defines a counted string used for UNICODE-strings
--- a/lib/winapi/windef.h
+++ b/lib/winapi/windef.h
@@ -3,6 +3,8 @@
 
 #include <xboxkrnl/xboxdef.h>
 
+#define WINAPI __stdcall
+
 #define MAX_PATH 260
 
 #endif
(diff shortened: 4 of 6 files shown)


lib/xboxkrnl/xboxkrnl.h
```diff
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <xboxkrnl/xboxdef.h>
+#include <xboxkrnl/xboxnt.h>
```
JayFoxRox: Reminder: These will probably need this in the future:

```
/* stop clang from crying */
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wlanguage-extension-token"

/* MSVC-compatibility for structure-packing */
#pragma ms_struct on
```

*(Can be done after this PR, but it's something worth noting before we add even more)*

lib/winapi/winbase.h
```diff
@@ -2,6 +2,7 @@
 #define __WINBASE_H__
 
 #include <windef.h>
+#include <winnt.h>
```
JayFoxRox: I'm not sure if this should be included from here, or from windef or elsewhere.

I don't remember why I put this here.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -132,12 +133,12 @@ typedef enum _MODE {
 /**
  * This struct defines a counted string used for ANSI-strings
  */
-typedef struct _ANSI_STRING
+typedef struct
```
dracc: Anonymous structs is a GCC extension, do we want to use those or should we name structs when we have the opportunity?

JayFoxRox: This change also causes `warning: incompatible pointer types passing 'POBJECT_STRING'
      (aka 'STRING *') to parameter of type 'PANSI_STRING' (aka 'struct _STRING *') [-Wincompatible-pointer-types]`

So I'll have to look into fixing this in general.

JayFoxRox: Fixed by reusing the `STRING` type.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -132,12 +132,7 @@ typedef enum _MODE {
 /**
  * This struct defines a counted string used for ANSI-strings
  */
-typedef struct _ANSI_STRING
-{
-    USHORT Length;        /**< Length in bytes of the string stored in the buffer */
-    USHORT MaximumLength; /**< Length in bytes of the buffer */
-    PSTR Buffer;          /**< Pointer to the buffer used for the character-string */
-} ANSI_STRING, *PANSI_STRING, STRING, *PSTRING, OBJECT_STRING, *POBJECT_STRING;
+typedef STRING OBJECT_STRING, *POBJECT_STRING;
```
JayFoxRox: The comment for this is broken.

lib/winapi/debug.c
```diff
@@ -0,0 +1,17 @@
+#include <winbase.h>
+#include <winnt.h>
+
+#include <string.h>
+
+void WINAPI OutputDebugStringA (LPCTSTR lpOutputString)
```
thrimbor: MS docs describe that this function calls `DbgPrint()` internally. I couldn't spot any difference in behavior when using this functions vs. using DbgPrint (when called on Retail or Debug BIOSes), so is this something we should do, too?

lib/winapi/winbase.h
```diff
@@ -11,6 +12,14 @@ extern "C"
 DWORD GetLastError (void);
 void SetLastError (DWORD error);
 
+void WINAPI OutputDebugStringA (LPCTSTR lpOutputString);
```
thrimbor: [MS docs](https://docs.microsoft.com/en-us/windows/desktop/api/debugapi/nf-debugapi-outputdebugstringa) indicate that this should be located in a header called `debugapi.h`.

lib/xboxkrnl/xboxkrnl.h
thrimbor: Should that typedef get moved, too?

lib/winapi/debug.c
JayFoxRox: MS docs are either wrong (changes over time?), or `DbgPrint` was inlined (which I doubt).

My version is based on reverse engineering of games, and it does not call any function which does string formatting (which `DbgPrint` would do).

Another difference is that this one doesn't need a function call and can be LTCG'd; it also isn't imported from the kernel, which can be relevant in some cases.

lib/winapi/winbase.h
JayFoxRox: I believe there's 2 different versions of this function: for userspace, and for kernel drivers.

https://source.winehq.org/WineAPI/OutputDebugStringA.html (and other languages of the MSDN article) claim it's in winbase.h, and I believe I found the location from a copy of an old DDK myself.

It's also possible that they moved the function over time to make winbase.h / windows.h more modular.

lib/xboxkrnl/xboxkrnl.h
JayFoxRox: The object API is part of the kernel; I'm not sure if anything else uses it.

I think this is a custom name for Xbox anyway - I believe I checked Windows SDK and DDK and wasn't able to find this name? Current MSDN documents the users as `UNICODE_STRING` (which would be equivalent to a generic `ANSI_STRING`?) as far as I know.

Also I had no use for it as part of `OutputDebugString`, so I didn't want to touch it.

thrimbor:
Code appears to be fine and nobody voiced any concerns, so if you resolve the conflicts I'll merge later today.

thrimbor:
Turns out I can force push to PR-ed branches, so I took the liberty of rebasing this myself. Merging now.

PR #122, opened 2019-05-20
usb: Don't consume events at startup

JayFoxRox:
`BootStartUSB` consumes important events during 200ms of the startup; these events are probably meant to register connected devices in the previously loaded drivers.

However, our code currently doesn't remember already-connected devices. So if a device driver is not registered as part of `BootStartUSB` it will never see events which announce the device (as they have been consumed already).
So the only way to get these events / support devices is to re-plug them to get another announcement.

Therefore, this PR removes the initial packet consumption, which means you must call `USBGetEvents` yourself (after `BootStartUSB`), to make sure that the devices are actually found. This already happens as part of the `XInput_Init` function anyway, so I assume no changes are necessary for most programs.

With this change, our USB stack works for custom drivers (which are not part of nxdk), and it also avoids a 200ms delay at startup.

*(There's something similar in `BootStopUSB`; it's not as problematic, but will probably be removed in the future)*

---

I have barely tested this, and I did not try it on Xbox 1.0.

It's possible that the intention behind the 200ms delay is a different one, such as removing garbage from the event queue during startup or similar. If that's the case, a different solution will be necessary. However, that's unlikely as it would probably cause issues with existing drivers.

### DIFF
--- a/lib/usb/sys/BootUSB.c
+++ b/lib/usb/sys/BootUSB.c
@@ -60,11 +60,6 @@ void BootStartUSB(void)
 	XRemoteInit();	
 	UsbKeyBoardInit();
 	UsbMouseInit();
-
-	for(n=0;n<200;n++) {
-		USBGetEvents();
-		wait_ms(1);
-	}
 }
 /*------------------------------------------------------------------------*/ 
 void USBGetEvents(void)


mborgerson:
Waiting for someone to test this PR

dracc:
I've just tested NevolutionX with this PR; failed to open the `SDL_GameController()`.

dracc:
I've rebuilt and tested https://github.com/dracc/NevolutionX/commit/173631967 after rebasing this PR upon the current `master HEAD` (b1cd396).
Still doesn't work. Gives this error:
```Joystick Error: Couldn't find mapping for device (0)```

Same error encountered both in XQEMU and on my 1.6, works fine in both places without this change.

JayFoxRox:
It couldn't possibly work: only the polling variant of `XInput_Init` calls `USBGetEvents`.
I believe I had some more issues with the interrupt based USB code, too (hotplugging for example); so I'm not confident enough to fix this PR right now.

I'll close this, will eventually analyze the USB code in-depth, and come back to this (as a new PR).

PR #131, opened 2019-06-10
vp20compiler: Forward comments from input to output

JayFoxRox:
The final vs.inl currently doesn't contain the VS variable / constant locations that the Cg vertex program uses. That's an issue as it's necessary to set up a working program.

This adds a small loop, which collects all comments, and forwards them from input to output.

Unfortunately, this is not position aware, so it won't interleave encoded instructions and comments. But it's still good enough to read back the `#var` and `#const` lines.

It isn't really secure, and it's probably easy to write malicious code which outputs garbage C code (like introducing `/*` as part of the comment).
Ideally we'd parse only those important lines and generate a `float c[][4];` and code which uploads it; however, this will have to do for now.

This was barely tested, but should work fine.

---

As an alternative for this PR, we could also remove the deletion step for the intermediate shader files.

### DIFF
--- a/tools/vp20compiler/main.c
+++ b/tools/vp20compiler/main.c
@@ -296,6 +296,24 @@ void translate(const char* str)
 
 //     printf("num_instructions: %u\n", num_instructions);
 
+    const char* cur = str;
+    const char* end = &str[strlen(str) - 1];
+    while(cur < end) {
+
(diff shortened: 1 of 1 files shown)


thrimbor:
Tested and better than master. Merged.

PR #133, opened 2019-06-10
net: Fix memory flags

### LINKED ISSUE #128: Allocations in network code use wrong flags
The allocations in the network code claim that memory flags `4` is non-cached, non-ordered.

I'm not sure where these values come from. Normally these Xbox functions use permission bits (`READ` / `WRITE`), and then bits for additional flags. Uncached, in particular, seems to be 0x200.

I'm not sure if these wrong flags have any effect.

I discovered this while trying to make network interrupts work (thinking that bad data might still be in cache), but it seems to crash regardless of these flags.

JayFoxRox:
Closes #128 

Untested code, although I have been running similar variations without problems.

### DIFF
--- a/lib/net/nforceif/src/driver.c
+++ b/lib/net/nforceif/src/driver.c
@@ -122,19 +122,21 @@ low_level_init(struct netif *netif)
 
   /* Do whatever else is needed to initialize interface. */
 
+  // This is a temporary buffer, independent of hardware; cache is fine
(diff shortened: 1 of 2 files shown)


thrimbor:
Hmm, not a single requested reviewer responded. :(
Anyway, gave the code a quick review and it seems to be fine, also confirmed it works on my Xbox. Merging.

PR #135, opened 2019-06-11
build: Explicitly request embedded DWARF debug info

thrimbor:
LLVM's default behavior seems to have changed, making it necessary to explicitly request DWARF debug info for source level debugging as described in the wiki.
Tested with clang 8.0.

/edit:
This specifies DWARF 4, which is the most recent working version - attempting to use DWARF 5 crashes clang 8.0.


mborgerson:
Great, thanks!

PR #136, opened 2019-06-12
pbkit: Use surface offset instead of DMA base to select surface

### LINKED ISSUE #75: pbkit surface addresses controlled using base
It's not necessarily a problem, but I'd like to have on record that the framebuffer is configured using the surface offset in D3D (base 0), but pbkit uses the base to address the surface (offset 0).

This caused issues in nv2a-trace fo me (which I'll address in the future). I'd recommend to switch to D3D behaviour - there might be a reason why this was done.

JayFoxRox:
Closes #75 

The DMA object always includes all RAM (except instance memory), and the GPU just switches the offset of the surface within that DMA range.

By avoiding `NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT` we don't need any CPU involvement for setting the new backbuffer, so this should be faster.

However, as also mentioned in the discussion for #75:

> We could keep the current method for debug builds, because the GPU will warn if the surface area was miscalculated (which appears for 16 bit surfaces - although I'm not sure why / how yet.)

So after merge we should create a new issue to reintroduce the old code (albeit with cleaner code style).

I'm also not sure if this code is complete. It's possible we have to add some operations like the following, to make it safe:

```c
pb_push1(p,NV20_TCL_PRIMITIVE_3D_ASK_FOR_IDLE,0); p+=2; //ask for idle
pb_push1(p,NV20_TCL_PRIMITIVE_3D_NOP,0); p+=2; //wait for idle
pb_push1(p,NV20_TCL_PRIMITIVE_3D_WAIT_MAKESPACE,0); p+=2; //wait/makespace (obtains null status)
```

I've only tested this code with the triangle sample. I didn't notice any issues.

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -1839,101 +1839,16 @@ void pb_target_back_buffer(void)
     DWORD           height;
     DWORD           pitch;
     DWORD           pitch_depth_stencil;
-
-    DWORD           dma_flags;
-    DWORD           dma_addr;
-    DWORD           dma_limit;
-
-    int         flag;
-    int         depth_stencil;
     
     width=pb_FrameBuffersWidth;
     height=pb_FrameBuffersHeight;
     pitch=pb_FrameBuffersPitch;
     pitch_depth_stencil=pb_DepthStencilPitch;
 
-    //DMA channel 9 is used by GPU in order to render pixels
-    dma_addr=pb_FBAddr[pb_back_index]&0x03FFFFFF;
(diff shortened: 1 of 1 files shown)


JayFoxRox:
I'm still having some trouble with the old method and 16bpp surfaces. We should delay this until the issue is understood.
It's possible that pbkit miscalculates the surface sizes.

JayFoxRox:
The issue I mentioned can be prevented by #140 . I'm not sure why exactly, but I assume that the Zeta tile expects a certain size / pitch, so that compression and other optimizations work.

PR #137, opened 2019-06-21
samples: Add XVideoSetMode()

dracc:
This will be required when XboxDev/nxdk-pdclib#3 is merged and
`XVideoSetMode()` in `crt0.c` has been removed.

Not suitable for merge before then, do want to update the submodule in this same PR.

### DIFF
--- a/samples/httpd/main.c
+++ b/samples/httpd/main.c
@@ -8,6 +8,7 @@
 #include "netif/etharp.h"
 #include "pktdrv.h"
 #include <hal/input.h>
+#include <hal/video.h>
 #include <hal/xbox.h>
 #include <pbkit/pbkit.h>
 #include <xboxkrnl/xboxkrnl.h>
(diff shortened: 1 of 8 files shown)


thrimbor:
https://github.com/XboxDev/nxdk-pdclib/pull/3 has been merged, so you can continue to work on this. https://github.com/XboxDev/nxdk-pdclib/pull/3#issuecomment-506748349 is important for that.

samples/httpd/main.c
```diff
@@ -67,6 +68,8 @@ void main(void)
 	sys_sem_wait(&init_complete);
 	sys_sem_free(&init_complete);
 
+        XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
```
DiscoStarslayer: nit: Mix of tab + spaces here. Seems to be the only sample using tabs, but we should probably be consistent for the file and create a PR later to change the styling of this file down the road if changing this is important enough.

DiscoStarslayer:
https://github.com/XboxDev/nxdk/issues/149 Issue created

thrimbor:
Thanks, merging.

PR #140, opened 2019-06-26
pbkit: Only allocate a single depth buffer

JayFoxRox:
For some odd reason the existing pbkit code allocates one zeta buffer per framebuffer.

This is stupid because there's rarely any situation where you need to use the z-buffer of a framebuffer that isn't currently being drawn.
Using one shared buffer reduces memory usage.

Additionally the existing code used the color-buffer size instead of the z-buffer size.

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -3569,13 +3569,13 @@ int pb_init(void)
 
     pb_DSSize=Size;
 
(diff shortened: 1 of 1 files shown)


thrimbor:
Testing didn't reveal any issues. Checking pbkit and sample code, the other z-buffers seem to never get accessed, so I'm merging this.

PR #141, opened 2019-06-27
pbkit: Allow setting color surface format

JayFoxRox:
This allows users to control the pbkit framebuffer color format (prior to `pb_init`). To make this more useful, R5G5B5 output is added to our video code.
The `debugPrint` code also wasn't aware of this mode, so I added some code to handle it.

This is a temporary API, as we'll have to make many more changes to the pbkit surface API in the future (as it doesn't really allow non-framebuffer rendertargets except the flawed `extra_buffer` code). However, such refactors will be in future PRs.

A similar PR for the zeta buffer will follow in the near future.

### DIFF
--- a/lib/hal/debug.c
+++ b/lib/hal/debug.c
@@ -30,7 +30,7 @@ static const unsigned char systemFont[] =
 static void drawChar(unsigned char c, int x, int y, int fgColour, int bgColour)
 {
 	unsigned char *videoBuffer = XVideoGetFB();
-	videoBuffer += (y * SCREEN_WIDTH + x) * (SCREEN_BPP >> 3);
+	videoBuffer += (y * SCREEN_WIDTH + x) * ((SCREEN_BPP+7)/8);
 
 	unsigned char mask;
 	const unsigned char *font = systemFont + (c * FONT_WIDTH);
@@ -60,6 +60,7 @@ static void drawChar(unsigned char c, int x, int y, int fgColour, int bgColour)
 					videoBuffer += sizeof(int);
 					break;
 				case 16:
+				case 15:
 					*((short*)videoBuffer) = colourToDraw & 0xFFFF;
 					videoBuffer += sizeof(short);
 					break;
@@ -71,7 +72,7 @@ static void drawChar(unsigned char c, int x, int y, int fgColour, int bgColour)
 #endif
 		}
 		
-		videoBuffer += (SCREEN_WIDTH-FONT_WIDTH)  * (SCREEN_BPP >> 3);
+		videoBuffer += (SCREEN_WIDTH-FONT_WIDTH)  * ((SCREEN_BPP+7)/8);
 		font++;
 	}
 }
@@ -155,6 +156,10 @@ void debugPrint(const char *format, ...)
 	case 16:
 		fgColour = WHITE_16BPP;
 		bgColour = BLACK_16BPP;
+		break;
+	case 15:
+		fgColour = WHITE_15BPP;
+		bgColour = BLACK_15BPP;
 	}
 
 	unsigned char *s = (unsigned char*)	buffer;
@@ -188,7 +193,7 @@ void debugPrint(const char *format, ...)
 
 void advanceScreen( void )
 {
-	int pixelSize = SCREEN_BPP >> 3;
+	int pixelSize = (SCREEN_BPP+7)/8;
 	int screenSize  = SCREEN_WIDTH * (SCREEN_HEIGHT - MARGINS)  * pixelSize;
 	int lineSize    = SCREEN_WIDTH * (FONT_HEIGHT + 1) * pixelSize;
 	
@@ -205,7 +210,7 @@ void debugClearScreen( void )
 {
 	unsigned char* videoBuffer = XVideoGetFB();
 
-	memset( videoBuffer, 0, (SCREEN_BPP >> 3) * (SCREEN_WIDTH * SCREEN_HEIGHT) );
+	memset( videoBuffer, 0, ((SCREEN_BPP+7)/8) * (SCREEN_WIDTH * SCREEN_HEIGHT) );
 	nextRow = MARGIN;
 	nextCol = MARGIN; 
 }
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -2,6 +2,8 @@
 #include <hal/xbox.h>
 #include <hal/video.h>
 #include <string.h>
+#include <assert.h>
+#include <stdbool.h>
 
 
 // Registers
@@ -26,8 +28,9 @@
 #define VIDEO_REGION_PAL			0x00000300
 
 // Defines for pixel formats
-#define VIDEO_BPP_16				0x00000011
-#define VIDEO_BPP_32				0x00000012
+#define VIDEO_A1R5G5B5				0x00000010
+#define VIDEO_R5G6B5				0x00000011
+#define VIDEO_A8R8G8B8				0x00000012
 
 
 unsigned char*	_fb;
@@ -259,15 +262,32 @@ VIDEO_MODE XVideoGetMode(void)
 void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 {
 	ULONG Step = 0;
-	DWORD dwFormat = (bpp == 32) ? VIDEO_BPP_32 : VIDEO_BPP_16;
-	int screenSize = width * height * (bpp/8);
+	DWORD dwFormat = 0;
+	int bytesPerPixel = (bpp+7)/8;
+	int pitch = width * bytesPerPixel;
+	int screenSize = pitch * height;
+
+	switch(bpp) {
+	case 15:
+		dwFormat = VIDEO_A1R5G5B5;
+		break;
+	case 16:
+		dwFormat = VIDEO_R5G6B5;
+		break;
+	case 32:
+		dwFormat = VIDEO_A8R8G8B8;
+		break;
+	default:
+		assert(false);
+		break;
+	}
 
 	XVideoSetVideoEnable(FALSE);
 
 	do
 	{
 		Step = AvSetDisplayMode((PVOID)VIDEO_BASE, Step, 
-			dwMode, dwFormat, width*(bpp/8), VIDEO_FRAMEBUFFER);
+			dwMode, dwFormat, pitch, VIDEO_FRAMEBUFFER);
 	} while(Step);
 
 	XVideoSetFlickerFilter(5);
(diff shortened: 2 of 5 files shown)


lib/hal/video.c
```diff
@@ -2,6 +2,8 @@
 #include <hal/xbox.h>
 #include <hal/video.h>
 #include <string.h>
+#include <assert.h>
+#include <stdbool.h>
```
JayFoxRox: This adds newer C features to this file, is this okay?
Should our driver code even use libc?

I'd argue this is fine; but we should have a minimalistic `assert` that uses `RtlAssert` in the future (to avoid side-effects of driver code).

lib/hal/video.c
```diff
@@ -259,15 +262,32 @@ VIDEO_MODE XVideoGetMode(void)
...
-	DWORD dwFormat = (bpp == 32) ? VIDEO_BPP_32 : VIDEO_BPP_16;
-	int screenSize = width * height * (bpp/8);
+	DWORD dwFormat = 0;
+	int bytesPerPixel = (bpp+7)/8;
+	int pitch = width * bytesPerPixel;
+	int screenSize = pitch * height;
+
+	switch(bpp) {
+	case 15:
+		dwFormat = VIDEO_A1R5G5B5;
```
JayFoxRox: bpp=15 was the sanest solution I could find to integrate this. This is also how it's handled in QEMU

lib/pbkit/pbkit.c
```diff
@@ -20,14 +20,14 @@
 #include <hal/io.h>
 #include <xboxkrnl/xboxkrnl.h>
 #include <xboxrt/debug.h>
+#include <stdbool.h>
+#include <assert.h>
```
JayFoxRox: Same question as for video file: Should we use libc?

lib/pbkit/pbkit.c
```diff
@@ -79,6 +79,8 @@ struct s_CtxDma
     DWORD               isGr;
 };
 
+static unsigned int pb_ColorFmt = NV097_SET_SURFACE_FORMAT_COLOR_LE_A8R8G8B8;
+static unsigned int pb_DepthFmt = NV097_SET_SURFACE_FORMAT_ZETA_Z24S8;
```
JayFoxRox: Should this be `uint32_t` or something? These are pushbuffer method data fields (which are 32 bit unsigned integers)

lib/pbkit/pbkit.c
```diff
@@ -2544,9 +2546,22 @@ void pb_fill(int x, int y, int w, int h, DWORD color)
...
+        color=((color>>16)&0x8000)|((color>>7)&0x7C00)|((color>>5)&0x03F0)|((color>>3)&0x001F);
+        break;
+    case NV097_SET_SURFACE_FORMAT_COLOR_LE_R5G6B5:
+        color=((color>>8)&0xF800)|((color>>5)&0x07E0)|((color>>3)&0x001F);
+        break;
+    case NV097_SET_SURFACE_FORMAT_COLOR_LE_A8R8G8B8:
+        // Nothing to do, input is A8R8G8B8
+        break;
+    default:
+        assert(false);
```
JayFoxRox: This should also throw an error message which explains the problem.

More modes can be added in the future.

lib/pbkit/pbkit.c
```diff
@@ -2761,7 +2776,10 @@ void pb_kill(void)
 }
 
 
-
+void pb_set_color_format(unsigned int fmt, bool swizzled) {
+    pb_ColorFmt = fmt;
+    assert(swizzled == false);
```
JayFoxRox: Support for swizzled buffers should be added in the future.  I prepared the API to be able to handle this.

samples/triangle/main.c
```diff
@@ -52,6 +52,10 @@ void main(void)
     int       start, last, now;
     int       fps, frames, frames_total;
 
+    XVideoSetMode(640, 480, 16, REFRESH_DEFAULT);
+
+    pb_set_color_format(NV097_SET_SURFACE_FORMAT_COLOR_LE_R5G6B5, false);
+
```
JayFoxRox: This can be removed. It's only for testing.
But we might as well keep it to demo how to use this?

Alternative combinations:
```c
    XVideoSetMode(640, 480, 16, REFRESH_DEFAULT);
    pb_set_color_format(NV097_SET_SURFACE_FORMAT_COLOR_LE_X1R5G5B5_Z1R5G5B5, false);
```
```c
    XVideoSetMode(640, 480, 16, REFRESH_DEFAULT);
    pb_set_color_format(NV097_SET_SURFACE_FORMAT_COLOR_LE_X1R5G5B5_O1R5G5B5, false);
```
```c
    pb_set_color_format(NV097_SET_SURFACE_FORMAT_COLOR_LE_A8R8G8B8, false);
```

*(Last one is default if it's omitted)*

lib/pbkit/pbkit.c
```diff
@@ -79,6 +79,8 @@ struct s_CtxDma
     DWORD               isGr;
 };
 
+static unsigned int pb_ColorFmt = NV097_SET_SURFACE_FORMAT_COLOR_LE_A8R8G8B8;
```
JayFoxRox: This hardcodes the default mode to 32bpp.
Previously, `vm.bpp` was used to select 16bpp or 32bpp mode in `pb_init`.

However, we can keep 32bpp hardcoded, because pbkit didn't even work in 16pp mode before, so every pbkit program only handles 32bpp mode (at least it didn't before #140 - I'll have to check if it works now). If anyone had `vm.bpp` at 16, pbkit would have crashed.

lib/pbkit/pbkit.c
```diff
@@ -2544,9 +2546,22 @@ void pb_fill(int x, int y, int w, int h, DWORD color)
...
     x2=x+w;
     y2=y+h;
-    
-    //if you supply 32 bits color and res is 16 bits, apply function below
-    //color=((color>>8)&0xF800)|((color>>5)&0x07E0)|((color>>3)&0x001F);
+
+    switch(pb_ColorFmt) {
+    case NV097_SET_SURFACE_FORMAT_COLOR_LE_X1R5G5B5_Z1R5G5B5:
+    case NV097_SET_SURFACE_FORMAT_COLOR_LE_X1R5G5B5_O1R5G5B5:
+        color=((color>>16)&0x8000)|((color>>7)&0x7C00)|((color>>5)&0x03F0)|((color>>3)&0x001F);
```
JayFoxRox: Should be able to remove the alpha portion: it's always forced to 1 (O1R5G5B5) or 0 (Z1R5G5B5).

lib/hal/video.c
JayFoxRox: Adding vm.bpp = 15 might cause issues in existing code, which only expects vm.bpp to be 16 or 32.

The debug.c code is affected by this, but I'm not even sure if it works correctly in 16bpp mode. Should be checked, and possibly fixed as part of this PR?

JayFoxRox: Fixed.

lib/xboxrt/debug.h
```diff
@@ -19,6 +19,9 @@ extern "C"
 #define WHITE_16BPP   0xFFFF
 #define BLACK_16BPP  0x0000
 
+#define WHITE_15BPP   0xFFFF
```
JayFoxRox: *Note: 0x7FFF would be enough, but setting alpha bit might be useful if `debugPrint` buffer is used as A1R5G5B5 texure in the future.*

lib/hal/video.c
thrimbor: I think it's fine. The functionality it uses is rather minimal and easily stubbed should someone need to get rid of the libc dependency.

Using `RtlAssert` for `assert` is already possible by building PDCLib with `DEBUG_CONSOLE` defined. It's not the default to make it easier to use for users of a retail BIOS - we should probably have code decide on runtime which `assert` implementation is to be used.

lib/pbkit/pbkit.h
```diff
@@ -21,6 +21,7 @@ extern "C"
 #endif
 
 #include <xboxkrnl/xboxkrnl.h>
+#include <stdbool.h>
```
thrimbor: `stdbool.h` is not supposed to be used in C++, so this should get enclosed in a corresponding `#if`. Our CI trips over this problem atm as our version of PDCLib's `stdbool.h` is not C++ compatible (which it is allowed to be). I have reported this upstream nonetheless and a workaround has been added to make it more forgiving, but ofc we shouldn't rely on that.

lib/xboxrt/debug.h
JayFoxRox: Using 0x7FFF (still untested), to match the 32bpp variant.

lib/pbkit/pbkit.h
JayFoxRox: I've added a guard (`#if !defined(__cplusplus)`)

JayFoxRox:
Rebased (to avoid merge conflicts due to debug.h move)

lib/pbkit/pbkit.c
mborgerson: `0x3F0` is a 6 bit mask, not 5 bits as expected. Should probably be `0x3E0`

lib/hal/debug.c
```diff
@@ -63,6 +63,10 @@ static void drawChar(unsigned char c, int x, int y, int fgColour, int bgColour)
 					*((short*)videoBuffer) = colourToDraw & 0xFFFF;
 					videoBuffer += sizeof(short);
 					break;
+				case 15:
+					*((short*)videoBuffer) = colourToDraw & 0xFFFF;
```
mborgerson: Same exactly as case for 16, prefer merging and having two case labels if we handle them exactly the same

samples/triangle/main.c
mborgerson: Waiting for removal

lib/pbkit/pbkit.c
mborgerson: `assert` is ok for now I think

lib/hal/debug.c
JayFoxRox: Changed; Only kept the case label

lib/pbkit/pbkit.c
JayFoxRox: Fixed.

samples/triangle/main.c
JayFoxRox: Removed.

mborgerson:
Thanks!

PR #142, opened 2019-07-03
video: Respect EEPROM video settings

dracc:
When a video setting has been selected and stored to EEPROM it should be respected.

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -209,7 +209,11 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
 	
 	DWORD dwAdapter = dwEnc & 0x000000FF;
 	DWORD dwStandard = dwEnc & 0x0000FF00;
-	
+
+	bool allow_480p = dwEnc & VIDEO_MODE_480P;
+	bool allow_720p = dwEnc & VIDEO_MODE_720P;
+	bool allow_1080i = dwEnc & VIDEO_MODE_1080I;
+
 	VIDEO_MODE_SETTING *pVidMode;
 	if (bpp == 0)
 	{
@@ -238,6 +242,27 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
 		if(pVidMode->refresh != refresh)
 			continue;
 
+		if(dwAdapter == AV_PACK_HDTV)
+		{
+			bool is_hd = pVidMode->dwMode & 0x80000000;
+
+			/* If 480p support is enabled, reject 480i modes
+			   (forces a 480p mode to be picked) */
+			if(allow_480p && (pVidMode->height == 480 && !is_hd))
+				continue;
+
+			/* If 480p support is disabled, reject those modes */
+			if(!allow_480p && (pVidMode->height == 480 && is_hd))
+				continue;
+
+			/* If 720p support is disabled, reject those modes */
+			if(!allow_720p && (pVidMode->height == 720 && is_hd))
+				continue;
+
+			/* If 1080i support is disabled, reject those modes */
+			if(!allow_1080i && (pVidMode->height == 1080 && is_hd))
+				continue;
+		}
 		break;
 	}
 	
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -310,67 +310,27 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 
 BOOL XVideoSetMode(int width, int height, int bpp, int refresh)
 {
-	VIDEO_MODE_SETTING *pVidMode = NULL;
-	int vidRefresh = 0;
-	int i = 0;
+	VIDEO_MODE vm;
+	void *p = NULL;
 
-	DWORD dwEnc = XVideoGetEncoderSettings();
-	
-	DWORD dwAdapter		= dwEnc & 0x000000FF;
-	DWORD dwStandard	= dwEnc & 0x0000FF00;
-
-	if(bpp != 15 && bpp != 16 && bpp != 32)
-		bpp = 32;
-
-	if(refresh > 0)
-		vidRefresh = refresh;
-	else
+	while(XVideoListModes(&vm, bpp, refresh, &p))
 	{
-		if(dwEnc & 0x00400000) //60Hz refresh rate
-			vidRefresh = 60;
-		else
-			vidRefresh = 50;
+		if(vm.width != width || vm.height != height)
+			continue;
+		
+		DWORD dwMode = vidModes[(int)p-1].dwMode;
+
+		XVideoInit(dwMode, vm.width, vm.height, vm.bpp);
+
+		vmCurrent.width = vm.width;
+		vmCurrent.height = vm.height;
+		vmCurrent.bpp = vm.bpp;
+		vmCurrent.refresh = vm.refresh;
+		return TRUE;
 	}
 
-	for(i=0; i<iVidModes; i++)
-	{
-		pVidMode = &vidModes[i];
-
-		if((pVidMode->dwFlags & 0x000000FF) != dwAdapter)
-			continue;
-
-		if(pVidMode->dwStandard != dwStandard)
-			continue;
-
-		if(pVidMode->width != width || pVidMode->height != height)
-			continue;
-
-		if(pVidMode->refresh != vidRefresh)
-			continue;
-
-		break;
-	}
-
-	if(i >= iVidModes) // No compatible mode found
-	{
-		memset(&vmCurrent, 0x00, sizeof(VIDEO_MODE));
-		return FALSE;
-	}
-
-	XVideoInit(pVidMode->dwMode, pVidMode->width, pVidMode->height, bpp);
-
-//Will be registered at first XVideoWaitForVBlank() call anyway. No need to lock IRQ3 here.
-//	if (! IsrRegistered) {
-//		if (InstallVBLInterrupt())
-//			IsrRegistered = TRUE;
-//	}
-
-	vmCurrent.width = pVidMode->width;
-	vmCurrent.height = pVidMode->height;
-	vmCurrent.bpp = bpp;
-	vmCurrent.refresh = vidRefresh;
-
-	return TRUE;
+	memset(&vmCurrent, 0x00, sizeof(VIDEO_MODE));
+	return FALSE;
 }
 
 
--- a/lib/hal/video.h
+++ b/lib/hal/video.h
@@ -23,6 +23,11 @@ extern "C"
 #define VIDEO_ENC_FLICKERFILTER     11
 #define VIDEO_ENC_SOFTEN_FILTER     14
 
+// Defines for video resolution modes
+#define VIDEO_MODE_720P   0x20000
+#define VIDEO_MODE_1080I  0x40000
+#define VIDEO_MODE_480P   0x80000
+
 //Defines for refresh rates
 #define REFRESH_50HZ                50
 #define REFRESH_60HZ                60


lib/hal/video.c
```diff
@@ -227,6 +231,17 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
...
+		if(eepromVidSize & 0x000E0000)
+		{
+			if(pVidMode->height == 480 && !(eepromVidSize & 0x00080000))
+				continue;
+
+			if(pVidMode->height == 720 && !(eepromVidSize & 0x00020000))
+				continue;
+
+			if(pVidMode->height == 1080 && !(eepromVidSize & 0x00040000))
+				continue;
```
JayFoxRox: This code is very non-descriptive. Introduce constants for the masks or at least move result into temporaries: `allow_480p = [...]` / `allow_720p = [...]` / `allow_1080p = eepromVidSize & 0x00040000;` etc.

lib/hal/video.c
```diff
@@ -198,6 +198,10 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
 	
 	DWORD dwAdapter = dwEnc & 0x000000FF;
 	DWORD dwStandard = dwEnc & 0x0000FF00;
+
+	ULONG eepromVidSize, eepromValueType, eepromReturnLength;
+
+	ExQueryNonVolatileSetting(0x08, &eepromValueType, (void*)&eepromVidSize, 4, &eepromReturnLength);
```
JayFoxRox: We should be using named constants for the first parameter. If we don't have any documented, you should craft your own and keep it in this file, just above this line (to make code more descriptive).

lib/hal/video.c
```diff
@@ -31,7 +31,7 @@
 
 
 unsigned char*	_fb;
-DWORD			dwEncoderSettings 	= 0;
+DWORD			dwEncoderSettings	= 0;
```
JayFoxRox: Accidental change? This file uses bad tabs anyway, so it's doomed to fail, if this is for alignment.

JayFoxRox:
You can potentially add tags to the mode table, therefore explicitly marking which modes are considered 480p, 720p, 1080p, widescreen, ... - this is better as programs could eventually query that same table to get information about the current mode.

JayFoxRox:
If this wasn't a draft, it could potentially get merged with minor changes and we could improve this feature later.

However, there has been very little progress over the past 3 months, and I'm tempted to close this.

dracc:
I'll make sure to take care of the review comments tonight and hopefully reach a non-draft stage.

lib/hal/video.c
JayFoxRox: This call is not necessary, `VIDEO_ENC_GET_SETTINGS` already retrieves this field. See https://github.com/Cxbx-Reloaded/Cxbx-Reloaded/blob/a6c0353c5beaf14766313050bfc0c9fe222147ac/src/core/kernel/exports/EmuKrnlAv.cpp#L133

lib/hal/video.c
dracc: Better now?

lib/hal/video.c
dracc: Fixed, thanks!

lib/hal/video.c
dracc: Fixed. :+1:

lib/hal/video.c
```diff
@@ -198,7 +198,13 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
...
 	DWORD dwAdapter = dwEnc & 0x000000FF;
 	DWORD dwStandard = dwEnc & 0x0000FF00;
-	
+
+        bool is_480p;
+        bool allow_480p = dwEnc & VIDEO_MODE_480P;
+        bool allow_480i = !allow_480p;
+        bool allow_720p = dwEnc & VIDEO_MODE_720P;
+        bool allow_1080i = dwEnc & VIDEO_MODE_1080I;
+
```
JayFoxRox: Indentation is different from surrounding code

lib/hal/video.c
```diff
@@ -227,6 +233,22 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
 		if(pVidMode->refresh != refresh)
 			continue;
 
+		if(dwAdapter == AV_PACK_HDTV)
+		{
+			is_480p = pVidMode->dwMode & 0x80000000;
```
JayFoxRox: That is not what that bit means.
It is also set for 720p, 1080i and the VGA modes.

I'm not sure what exactly it stands for, but it's definitely not just `is_480p`.

My best naming was `is_hd`, but it's still bad.

lib/hal/video.c
```diff
@@ -198,7 +198,13 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
 	
 	DWORD dwAdapter = dwEnc & 0x000000FF;
 	DWORD dwStandard = dwEnc & 0x0000FF00;
-	
+
+        bool is_480p;
+        bool allow_480p = dwEnc & VIDEO_MODE_480P;
+        bool allow_480i = !allow_480p;
```
JayFoxRox: If you also use `!is_480p` below (instead of `!is_480i`), then why not also use `!allow_480p`(instead of `allow_480i`)?

This is also different from the 480p, 720p and 1080i modes. So a comment explaining why another check for 480p is necessary, would be good (`If 480p is enabled, reject 480i modes (forces a 480p mode to be picked)`).

lib/hal/video.h
```diff
@@ -23,6 +23,11 @@ extern "C"
 #define VIDEO_ENC_FLICKERFILTER     11
 #define VIDEO_ENC_SOFTEN_FILTER     14
 
+// Defines for video resolution modes
+#define VIDEO_MODE_720P   0X20000
+#define VIDEO_MODE_1080I  0X40000
```
JayFoxRox: `0X` &rarr; `0x`

lib/hal/video.c
dracc: This was solved by removing this line in its entirety, the same data is being fetched by "`XVideoGetEncoderSettings()`".

lib/hal/video.c
dracc: Name suggestion: `is_progressiveScan`?

lib/hal/video.c
dracc: Fixed

lib/hal/video.h
dracc: Fixed. :+1:

lib/hal/video.c
dracc: I went with `is_hd` for now.

lib/hal/video.c
```diff
@@ -230,6 +236,24 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
...
+			is_hd = pVidMode->dwMode & 0x80000000;
+
+			if(pVidMode->height == 720 && !allow_720p)
+				continue;
+
+			if(pVidMode->height == 1080 && !allow_1080i)
+				continue;
+
+//			If 480p support is enabled, reject 480i modes
+//			(forces a 480p mode to be picked) */
```
JayFoxRox: - Comment indentation is weird.
- This has `// comment` but also `comment */` without `/*`
- Rest of file also seems to use `/* comment */`

lib/hal/video.c
```diff
@@ -201,7 +201,13 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
 	
 	DWORD dwAdapter = dwEnc & 0x000000FF;
 	DWORD dwStandard = dwEnc & 0x0000FF00;
-	
+
+	bool is_hd;
```
JayFoxRox: Declare where you set it, not up here (which is also in the middle of the function)

lib/hal/video.c
```diff
@@ -230,6 +236,24 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
...
 
+		if(dwAdapter == AV_PACK_HDTV)
+		{
+			is_hd = pVidMode->dwMode & 0x80000000;
+
+			if(pVidMode->height == 720 && !allow_720p)
+				continue;
+
+			if(pVidMode->height == 1080 && !allow_1080i)
+				continue;
```
JayFoxRox: 720p and 1080i mode should follow 480 modes so you get:
- 480i rejection + including comment
- 480p check
- 720p check
- 1080i check

lib/hal/video.c
dracc: Fixed.
I used `/*  */` at first  but started looking at https://github.com/XboxDev/nxdk/blob/554da020e0805990b552d2c015b512aaa67640be/lib/hal/video.c#L378 and drew bad conclusions.

lib/hal/video.c
dracc: Fixed. :+1:

lib/hal/video.c
dracc: Fixed. :+1:

lib/hal/video.c
```diff
@@ -230,6 +235,24 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
...
 			continue;
 
+		if(dwAdapter == AV_PACK_HDTV)
+		{
+			bool is_hd = pVidMode->dwMode & 0x80000000;
+
+			/* If 480p support is enabled, reject 480i modes
+			   (forces a 480p mode to be picked) */
+			if(pVidMode->height == 480 && !is_hd && !allow_480i)
+				continue;
```
JayFoxRox: There's no 480i mode in our video-mode table for HDTV adapter.

We need to adopt / add table from https://sourceforge.net/projects/openxdk/files/Patches%20and%20Drivers/Zagon_s%20480i%20patch/zagons-480i-patch.tar.gz/download in this or a separate PR first (or we potentially break applications)

lib/hal/video.c
```diff
@@ -230,6 +234,27 @@ BOOLEAN XVideoListModes(VIDEO_MODE *vm, int bpp, int refresh, void **p)
 		if(pVidMode->refresh != refresh)
 			continue;
 
+		if(dwAdapter == AV_PACK_HDTV)
+		{
+			bool is_hd = pVidMode->dwMode & 0x80000000;
+
+			/* If 480p support is enabled, reject 480i modes
+			   (forces a 480p mode to be picked) */
+			if(allow_480p && (pVidMode->height == 480 && !is_hd))
```
LukeUsher: What if an application exists that should force 480i mode? For example, testing how a display deals with interlacing in something similar to 240p test suite.

This would make that impossible without manually setting video mode,  as nxdk would always force the application into progressive-scan.

Although this would have little use outside of that kind of 'display test' application, so it might not be worth worring about

JayFoxRox: The performance shouldn't be impacted when switching between 480i and 480p.
So generally the user should expect to get the best quality that their display supports (which is set up in the dashboard).

> Although this would have little use outside of that kind of 'display test' application, so it might not be worth worring about

Agreed; I assume that we'll see another API to force (or search for) possible modes explicitly, but that's a niche use-case that we should ignore for now.

dracc:
Just a friendly reminder that #193 has been merged.

dracc:
In order for this change to come into use as often as necessary, JFR kindly submitted some changes to XVideoSetMode. The evaluation inside the `while` loop could be inverted (look for negations and carry on instead of matching and doing the stuff) to better fit surrounding code - I'm not quite sure how to do that properly at this moment though.

lib/hal/video.c
```diff
@@ -310,67 +310,26 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
...
-//			IsrRegistered = TRUE;
-//	}
-
-	vmCurrent.width = pVidMode->width;
-	vmCurrent.height = pVidMode->height;
-	vmCurrent.bpp = bpp;
-	vmCurrent.refresh = vidRefresh;
-
-	return TRUE;
+	memset(&vmCurrent, 0x00, sizeof(VIDEO_MODE));
```
JayFoxRox: nit: empty line before this one, to separate the "success" and "failure" case more clearly.

dracc: Blank line added.

lib/hal/video.c
```diff
@@ -310,67 +335,28 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
...
-			continue;
-
-		if(pVidMode->dwStandard != dwStandard)
-			continue;
-
-		if(pVidMode->width != width || pVidMode->height != height)
+		if(vm.width != width || vm.height != height)
+		{
 			continue;
+		}
```
JayFoxRox: nit: existing code (old code which this replaces + code in `XVideoListModes`) do not have `{}` for these early-rejections

lib/hal/video.c
```diff
@@ -310,67 +335,28 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
...
+		DWORD dwMode = vidModes[(int)p-1].dwMode;
```
JayFoxRox: nit: should add a blank line before this, to separate the rejection ("failure"/"rejection"-case from "success")

dracc: Newline added.

lib/hal/video.c
dracc: Fixed - curly braces removed.

JayFoxRox:
I'll merge this in a couple of hours if nobody vetos.

I have tested this with the mesh sample with 480p disabled (480i) and 480p enabled (480p) mode in NTSC. I assume that 720p and 1080p will also work (although I did not test them).

Without this code I have to restart my VGA box (by replugging the cables) when switching from 480p dashboard to 480i nxdk apps. So this PR is critical to me; hence the pressure to merge it.

PR #143, opened 2019-07-11
Clean up README

### LINKED ISSUE #73: Avoid duplication of documentation
Currently, a lot of the infomation is duplicated between README.md and the Wiki.

I think we should simply link the wiki from the README, and keep our README more simple.

The wiki encourages documentation contributions, but we don't have to waste time reviewing it.
The README will also feel like a wall-of-text, whereas the wiki can be used to structure content.

thrimbor:
Updates and reduces the README. This is intended to fix #73.

I tried to reduce the instructions to a useful minimum, adding links to the Wiki for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,27 +20,7 @@ You will need the following tools:
 - [lld](http://lld.llvm.org/)
 - [Git](http://git-scm.com/)
 
-#### OS X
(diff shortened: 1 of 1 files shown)


dracc:
Looks good to me! :+1:

PR #144, opened 2019-07-13
extract-xiso: Update to latest master

thrimbor:
Updates extract-xiso from https://github.com/XboxDev/extract-xiso/commit/01fe0b682d0dc5ea0789ff7205c8ab977d67d91c to https://github.com/XboxDev/extract-xiso/commit/2da27caba8fda945a1c5013271c91f2f69da01f5 so we can profit from the fixes for warnings and errors that landed in the meantime.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ TARGET       = $(OUTPUT_DIR)/default.xbe
 CXBE         = $(NXDK_DIR)/tools/cxbe/cxbe
 VP20COMPILER = $(NXDK_DIR)/tools/vp20compiler/vp20compiler
 FP20COMPILER = $(NXDK_DIR)/tools/fp20compiler/fp20compiler
-EXTRACT_XISO = $(NXDK_DIR)/tools/extract-xiso/extract-xiso
+EXTRACT_XISO = $(NXDK_DIR)/tools/extract-xiso/build/extract-xiso
 TOOLS        = cxbe vp20compiler fp20compiler extract-xiso
 NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
                -ffreestanding -nostdlib -fno-builtin -fno-exceptions \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -162,7 +162,7 @@ $(FP20COMPILER):
 extract-xiso: $(EXTRACT_XISO)
 $(EXTRACT_XISO):
 	@echo "[ BUILD    ] $@"
-	$(VE)$(MAKE) -C $(NXDK_DIR)/tools/extract-xiso $(QUIET)
+	$(VE)(cd $(NXDK_DIR)/tools/extract-xiso && cmake -G "Unix Makefiles" . $(QUIET) && $(MAKE) $(QUIET))
```
JayFoxRox: Why not `cmake` it in a subdirectory (`build/`)? This would simplify the `distclean`

thrimbor: My intent was to keep the path to the extract-xiso binary the same. Probably unimportant, now that I think about it.

thrimbor: I changed it.

Makefile
```diff
@@ -173,7 +176,7 @@ clean:
 
 .PHONY: distclean 
 distclean: clean
-	$(VE)$(MAKE) -C $(NXDK_DIR)/tools/extract-xiso clean $(QUIET)
+	$(VE)rm -rf $(NXDK_DIR)/tools/extract-xiso/build
```
DiscoStarslayer: nit: Can we get away with just a `rm -r`? I'd rather not force `rm` if we don't have to.

thrimbor: That would break `distclean` when `build` doesn't exist. I could offer `|| true` as an alternative though.

DiscoStarslayer: 👍 good point. I'd prefer `-f` then instead of silently allowing this line to fail.

JayFoxRox:
This adds dependency on CMake; because at the time, I thought we'd be migrating most host-based tools (and potentially also guest-samples) to CMake. This did not happen.

I'm still fine with this being merged; not sure about others though.

Makefile
JayFoxRox: Does "Unix Makefiles" work on msys2?

thrimbor: It does, but extract-xiso is currently broken on msys2/mingw. I sumitted https://github.com/XboxDev/extract-xiso/pull/30 to fix it.

Makefile
mborgerson: The convention here so far has been to let the sub-project clean up itself (by invoking `disclean` on the associated Makefile) instead of wiping everything away with force, but in this case this approach is ok

mborgerson:
Thanks

PR #147, opened 2019-07-18
io: Allow opening the same file multiple times

### LINKED ISSUE #146: SDL_TTF cannot open the same .ttf twice
This is probably somehow related to the nxdk implementation of `fopen()`.
Using the same file for two `TTF_Font*`s works fine in Linux so I'd say this is an issue in nxdk.

dracc:
Closes #146

Small change usable until the winapi stuff merges.

### DIFF
--- a/lib/hal/fileio.c
+++ b/lib/hal/fileio.c
@@ -253,7 +253,7 @@ int XCreateFile(
 		&ioStatusBlock,
 		NULL,
 		flagsAndAttributes,
-		0,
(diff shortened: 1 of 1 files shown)


dracc:
Commit message is bad, I blame the GitHub UI on mobile.

thrimbor:
I've taken the liberty to rename the commit.
Given that it's been tested and the obviousness of the problem it fixes I'm merging this after a quick review.

PR #148, opened 2019-07-19
sdl: Update submodule

dracc:
This update adds SDL_Joystick and SDL_GameController support.


thrimbor:
The SDL work has been tested and reviewed in https://github.com/XboxDev/nxdk-sdl/pull/5 already so I'm merging this after only checking that it updates correctly.

dracc:
Thanks! :muscle:

PR #150, opened 2019-08-02
winapi improvements

thrimbor:
This is quite a large PR because it adds a lot of functionality.
All of the new functions (and `FindFirstFile` as well now) rely on proper symbolic links instead of `XConvertDOSFilenameToXBOX`, the file listing sample has been modified accordingly.
The new functions are: `CreateFile`, `CloseHandle`, `ReadFile`, `WriteFile`, `SetFilePointer`, `SetFilePointerEx`, `GetFileSize`, `GetFileSizeEx`, `DeleteFile`, `MoveFile`, `CreateDirectory` and `RemoveDirectory`. All of these were written to comply to Windows' API, and the same restrictions apply (`MoveFile` can't move between different volumes, `RemoveDirectory` can only delete empty directories etc.). Not all of these functions are feature complete already - for example, "overlapped"-mode for file I/O is not supported yet - but assertions are in place in case someone tries to use that missing functionality.

I have a [separate branch](https://github.com/thrimbor/nxdk/tree/winapi_tests) where I added a bunch of unit tests for the new stuff, written against W7 with VS 2017. Atm all tests pass (if you try them yourself, be aware that they leave behind a bunch of `file_`-prefixed files on volume C that need to be removed manually), but they don't test all functions yet (I'll probably add more soon).

**/edit:**
This also marks the X-style file I/O functions as deprecated, so that they cause a warning when they're used. My goal is to get everyone to use winapi & libc functions, move PDCLib away from using the X-style functions and then remove them entirely.


lib/winapi/fileapi.h
```diff
@@ -3,19 +3,45 @@
...
+BOOL DeleteFileA (LPCTSTR lpFileName);
+BOOL RemoveDirectoryA (LPCSTR lpPathName);
+BOOL CreateDirectoryA (LPCSTR lpPathName, LPSECURITY_ATTRIBUTES lpSecurityAttributes);
+BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName);
+
 #ifndef UNICODE
+#define CreateFile CreateFileA
 #define FindFirstFile FindFirstFileA
 #define FindNextFile FindNextFileA
+static BOOL DeleteFile (LPCTSTR lpFileName) { return DeleteFileA(lpFileName); }
```
JayFoxRox: Why is this one singled out?

samples/winapi_filefind/main.c
```diff
@@ -15,6 +16,17 @@ int main()
...
+    OBJECT_STRING MountPath = RTL_CONSTANT_STRING("\\??\\C:");
+    OBJECT_STRING DrivePath = RTL_CONSTANT_STRING("\\Device\\Harddisk0\\Partition2\\");
+    NTSTATUS status;
+    status = IoCreateSymbolicLink(&MountPath, &DrivePath);
+    if (!NT_SUCCESS(status)) {
+        debugPrint("Failed to mount C: drive!\n");
+        XSleep(5000);
+        return -1;
+    }
+
```
JayFoxRox: Should there be some helper API which does mounting for compatibility with old code or MS XAPI?

It could be called from CRT0 (but it would have to be optional, to keep side-effects low for low-level use-cases).

---

Even if we decide against that, consider moving it into a separate function, so users can copy/paste this part to their own projects more easily.

lib/winapi/fileapi.h
thrimbor: Afaik Windows uses a macro for `DeleteFile` too, but for us, this causes issues. In `xboxkrnl.h`, the struct `FILE_DISPOSITION_INFORMATION` has a field named `DeleteFile`, which was affected when including `fileapi.h` before `xboxkrnl.h`.

My original solution added a workaround to `xboxkrnl.h` using `#pragma` to store away the define, undefine it, and restore it at the end. @mborgerson suggested to use a static function instead to escape the whole macro issue, and after thinking about it for a while I decided to follow the suggestion.

As for why I didn't change the rest: Rebase conflicts. There are several commits touching these lines, and even that change caused a lot of conflicts that I had to resolve manually - so if we want to change all of them to static functions, I'd rather do that in a follow-up PR tbh.

The whole discussion about it happened in Discord, starting here: https://discordapp.com/channels/428359196719972353/428359408204906497/607000885716975628

samples/winapi_filefind/main.c
thrimbor: Imo there should be an API for that, but I haven't found a good fit yet. There are Windows APIs for similar functionality but those are bit different in what they actually do/expect, and I didn't see a candidate in XAPI either - in fact, the only code I've seen written against XDK that used something like this just used `IoCreateSymbolicLink` directly, too.
The fact that we don't have something like that yet is part of the reason why I'm not axing the X*-functions with this PR already.

I just pushed a change that moves the mounting in a separate function in the sample, though.

thrimbor:
https://github.com/thrimbor/nxdk/commits/winapi_tests should now be complete, meaning that all functions in this PR have working tests that were written against VS 2017.

lib/winapi/fileapi.h
JayFoxRox:
```c
#define DeleteFile(...) DeleteFileA(__VA_ARGS__)
```

is it possible to use a function macro like this?

I'm fine with changing rest in a follow-up. Ideally we'd provide a function alias or something.

We should also check how Microsoft did it for something like the XDK - because they'd have had the same issue.

thrimbor: Well, now I'm embarrassed I didn't think of that.
It in fact solves the conflict while argument type and count checks still work fine. I'll push an update soon.

I don't know about XDK, but we dug up a [MS header](https://github.com/microsoft/service-fabric/blob/6f2a641bdfe4b7e7100780cd4a4473f8eda30d5c/src/prod/ktl/src/linuxinc/ktlpal.h#L635-L638) which simply undefined the macro, while `ntddk.h` seems to just ignore the issue (which you may get away with depending on your include order).
I don't know how they solve the problem in their code files however - in our implementation of `DeleteFileA`, we need to access the `DeleteFile`-field, so we had the conflict there, too. They must've had the same issue.

thrimbor: I took a bit longer, but I just pushed an update that changes this (and all the defines in the commits that follow) to a variadic macro.

JayFoxRox:
Accepting without further reviews and tests, as follow-ups are piling up and nobody seems to look into this at this point.
If problems are found, feel free to leave review comments / and new issues.

Merged.

Issue #15, opened 2017-07-08
lwip: Use BSD socket API


JayFoxRox:
The existing code uses the lwip netbuf interface, however, for compatibility with existing code and ease of use, we should rewrite the httpd example to use the BSD style socket API.

We should also make sure that the BSD style API works flawless.
nxdk-rdt has a lot of issues with the netbuf API, however, it is still unclear where exactly these issues stem from (be it nxdk-rdt, lwip or the Xbox specific network driver). With the BSD API we could test behaviour on other platforms and we will not have to worry about splitting or joining the network packets manually like we currently do.

thrimbor:
Resolved in #151.

PR #151, opened 2019-08-10
Samples: Add httpd bsd sample

dracc:
The code in `httpserver.c` is inspired by [Beej's Guide to Network Programming](https://beej.us/guide/bgnet/).

Somewhat related to #15.

### DIFF
--- a/samples/httpd_bsd/Makefile
+++ b/samples/httpd_bsd/Makefile
@@ -1,0 +1,6 @@
+XBE_TITLE=httpd_bsd
+GEN_XISO = $(XBE_TITLE).iso
+SRCS += $(CURDIR)/main.c $(CURDIR)/httpserver.c
+NXDK_DIR = $(CURDIR)/../..
+NXDK_NET=y
+include $(NXDK_DIR)/Makefile
--- a/samples/httpd_bsd/httpserver.c
+++ b/samples/httpd_bsd/httpserver.c
@@ -1,0 +1,174 @@
+#include <stdio.h>
+#include <lwip/opt.h>
+#include <lwip/arch.h>
+#include <lwip/netdb.h>
+#include <lwip/sockets.h>
+
+#include "httpserver.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <hal/xbox.h>
+#include <xboxkrnl/xboxkrnl.h>
+#include <debug.h>
+
+#if LWIP_SOCKET
+
+#ifndef HTTPD_DEBUG
+#define HTTPD_DEBUG         LWIP_DBG_OFF
+#endif
+
+#define PORT "80"   // port we're listening on
+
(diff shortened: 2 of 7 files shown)


samples/httpd_bsd/Makefile
```diff
@@ -0,0 +1,7 @@
+DEBUG=y
```
thrimbor: This shouldn't be set by default imo.

samples/httpd_bsd/httpserver.c
```diff
@@ -0,0 +1,174 @@
+#include <stdio.h>
+#include "lwip/opt.h"
```
thrimbor: nit: `#include <lwip/opt.h>` should be preferred here.

samples/httpd_bsd/httpserver.c
```diff
@@ -0,0 +1,174 @@
...
+#include "lwip/netdb.h"
+#include "lwip/sockets.h"
+
+#include "httpserver.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <hal/xbox.h>
+#include <xboxkrnl/xboxkrnl.h>
+#include <xboxrt/debug.h>
```
thrimbor: Use `#include <debug.h>` instead, also see https://github.com/XboxDev/nxdk-sdl/pull/8

samples/httpd_bsd/httpserver.c
```diff
@@ -0,0 +1,174 @@
...
+
+#if LWIP_SOCKET
+
+#ifndef HTTPD_DEBUG
+#define HTTPD_DEBUG         LWIP_DBG_OFF
+#endif
+
+#define PORT "9034"   // port we're listening on
```
thrimbor: Why not 80?

samples/httpd_bsd/main.c
```diff
@@ -0,0 +1,107 @@
+#include "lwip/debug.h"
```
thrimbor: nit: As above, `#include <lwip/debug.h>` should be preferred here.

/edit: Eh, seems like this comes from the httpd sample. We should probably just clean up all occurrences like these in a follow-up PR.

samples/httpd_bsd/main.c
```diff
@@ -0,0 +1,107 @@
...
+
+struct netif nforce_netif, *g_pnetif;
+
+err_t nforceif_init(struct netif *netif);
+void http_server_netconn_init(void);
+static void packet_timer(void *arg);
+
+static void tcpip_init_done(void *arg)
+{
+  sys_sem_t *init_complete = arg;
```
thrimbor: Since the indentation was changed from the httpd-sample anyway, use 4 spaces instead.

thrimbor:
Alright, merging.

PR #154, opened 2019-08-24
winapi improvements #2

thrimbor:
~Based on #150 (will rebase once that one is merged to solve conflicts).~

Most of these commits were done as part of my work to port libc++, but `GetTickCount()` and `GetSystemInfo()` were added to reduce nxdk-specific changes to PDCLib's dlmalloc-port (with a few more enhancements to nxdk, it can now build as it does on Win32 without changes).

`GetThreadId()`, `GetCurrentThreadId()`, `Sleep()` and `SleepEx` were tested separately.


lib/hal/xbox.h
```diff
@@ -9,6 +9,8 @@ extern "C"
 
 void XReboot();
 int  XGetTickCount();
```
JayFoxRox: Why not also deprecated?

thrimbor: An oversight, it is now.

thrimbor:
Now rebased onto master to fix conflicts.

lib/winapi/profiling.c
```diff
@@ -0,0 +1,20 @@
...
+#include <assert.h>
+#include <stddef.h>
+#include <xboxkrnl/xboxkrnl.h>
+
+BOOL QueryPerformanceCounter (LARGE_INTEGER *lpPerformanceCount)
+{
+    assert(lpPerformanceCount != NULL);
+
+    lpPerformanceCount->QuadPart = KeQueryPerformanceCounter();
+    return 1;
```
JayFoxRox: Why not use the `BOOL` specific values like `TRUE`?

lib/winapi/sync.c
```diff
@@ -0,0 +1,66 @@
...
+{
+    LARGE_INTEGER duration;
+    duration.QuadPart = ((LONGLONG)dwMilliseconds) * -10000;
+    while (true) {
+        NTSTATUS status = KeDelayExecutionThread(UserMode, bAlertable, &duration);
+        if (status != STATUS_ALERTED) {
+            return status == STATUS_USER_APC ? WAIT_IO_COMPLETION : 0;
+        }
+    }
+
```
JayFoxRox: nit: empty line. why?

lib/winapi/sysinfoapi.h
```diff
@@ -0,0 +1,49 @@
...
+
+void GetSystemTimePreciseAsFileTime (LPFILETIME lpSystemTimeAsFileTime);
+DWORD GetTickCount (void);
+
+// Unprovided fields are intentionally disabled to catch code trying to access them
+typedef struct _SYSTEM_INFO
+{
+    union
+    {
+        //DWORD dwOemId;
```
JayFoxRox: Why dead code?

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -55,13 +55,14 @@ typedef LONG NTSTATUS;
 #define STATUS_INVALID_HANDLE ((NTSTATUS)0xC0000008L)
 #define STATUS_HANDLE_NOT_CLOSABLE ((NTSTATUS)0xC0000235L)
 
-#define WAIT_FAILED ((NTSTATUS)0xFFFFFFFFL)
+#define WAIT_FAILED ((NTSTATUS)0xFFFFFFFFL) // FIXME: Not really an NTSTATUS
```
JayFoxRox: What's there to `FIXUP` here? Consider splitting this off by an empty line and explaining the situation in a comment (if necessary).

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -2111,6 +2080,8 @@ XBAPI VOID NTAPI RtlInitializeCriticalSection
     IN PRTL_CRITICAL_SECTION CriticalSection
 );
 
+#define RtlDeleteCriticalSection(CriticalSection) ((void)CriticalSection)
```
JayFoxRox: A bit torn about including non-exports here; fine for now though.

Probably deserves a comment as to why this exists the way it does.

You might want to add a comment in the user though.
If you have confirmed behaviour in MS XAPI that should also be documented in the user.

JayFoxRox: This macro still feels dangerous for when `CriticalSection` is some address math (unlikely, but also shows the problem of macros).
Not sure if an `static`-function is much better though.

lib/winapi/sysinfoapi.h
JayFoxRox: Nevermind, there's a comment above. I don't fully agree with this approach though (offsets of fields shouldn't change on Xbox).

lib/winapi/profiling.c
thrimbor: An oversight from copying to code over from another branch. It uses `TRUE` now.

lib/winapi/sync.c
thrimbor: Accidental mistake, fixed.

lib/winapi/sysinfoapi.h
thrimbor: I'm not fully satisfied with my solution, either. Ideally, there would be an attribute to mark the fields so that the user gets an error or a warning, but there doesn't seem to be such a thing.

lib/xboxkrnl/xboxkrnl.h
thrimbor: The fix would be to remove `WAIT_FAILED` from the kernel headers, but PDCLib currently expects it to be there.

lib/xboxkrnl/xboxkrnl.h
thrimbor: On the Xbox, `RtlInitializeCriticalSection` only sets up stuff in the `RTL_CRITICAL_SECTION` you pass, that's why no `RtlDeleteCriticalSection` exists as an export - there simply doesn't need to be one. Afaik XAPI headers define `RtlDeleteCriticalSection` and `DeleteCriticalSection` as `((void)0)`.
I decided to put `RtlDeleteCriticalSection` there for compatibility, in case some code out there expects this to exist.

lib/xboxkrnl/xboxkrnl.h
JayFoxRox: Where should it be? Why can't we move it?

JayFoxRox: Oh I see - it's in the synchapi stuff, but we can't update PDCLib (to find it), because we'd also pull other PDCLib changes if we update the submodule.

dracc:
I've looked at the code, found nothing odd that's not already pointed out and those parts have been explained already.
Haven't tried building it (yet), doubt it won't work.

Lgtm. :+1:

JayFoxRox:
Somehow CI broke for this on merge: https://travis-ci.org/XboxDev/nxdk/jobs/589441844
It still worked during review: https://travis-ci.org/XboxDev/nxdk/jobs/578533447

PR #156, opened 2019-08-26
Libraryization

thrimbor:
This splits out `PDCLib`, `hal`, `Freetype` and `SDL_ttf` to be built as libraries (others are untouched to avoid getting too many merge conflicts).

I am well aware that this is still far from being a complete solution (everything is built with the same compiler flags, libs are automatically added by makefiles instead of the samples specifying their requirements, still the one large main Makefile, sub-Makefiles adding cleaning rules to the main Makefile via a variable etc.) - I consider this only a first step and, most importantly, a fix to get nxdk to build on msys2 again.

### DIFF
--- a/lib/Makefile-pdclib
+++ b/lib/Makefile-pdclib
@@ -1,0 +1,16 @@
+PDCLIB_SRCS := \
+	$(shell find $(NXDK_DIR)/lib/pdclib/functions/ -name "*.c") \
+	$(shell find $(NXDK_DIR)/lib/pdclib/platform/xbox/ -name "*.c") \
+	$(shell find $(NXDK_DIR)/lib/pdclib/platform/xbox/ -name "*.s")
+
+PDCLIB_OBJS = $(addsuffix .obj, $(basename $(PDCLIB_SRCS)))
+
+$(NXDK_DIR)/lib/libpdclib.lib: $(PDCLIB_OBJS)
+
+main.exe: $(NXDK_DIR)/lib/libpdclib.lib
+
+CLEANRULES += clean-pdclib
+
+.PHONY: clean-pdclib
+clean-pdclib:
+	$(VE)rm -f $(PDCLIB_OBJS) $(NXDK_DIR)/lib/libpdclib.lib
--- a/lib/hal/Makefile
+++ b/lib/hal/Makefile
@@ -1,0 +1,21 @@
+HAL_SRCS := \
+	$(NXDK_DIR)/lib/hal/audio.c \
+	$(NXDK_DIR)/lib/hal/debug.c \
+	$(NXDK_DIR)/lib/hal/fileio.c \
+	$(NXDK_DIR)/lib/hal/input.c \
+	$(NXDK_DIR)/lib/hal/io.c \
+	$(NXDK_DIR)/lib/hal/led.c \
+	$(NXDK_DIR)/lib/hal/video.c \
+	$(NXDK_DIR)/lib/hal/xbox.c
+
+HAL_OBJS = $(addsuffix .obj, $(basename $(HAL_SRCS)))
+
+$(NXDK_DIR)/lib/libnxdk_hal.lib: $(HAL_OBJS)
+
+main.exe: $(NXDK_DIR)/lib/libnxdk_hal.lib
+
+CLEANRULES += clean-hal
+
+.PHONY: clean-hal
+clean-hal:
+	$(VE)rm -f $(HAL_OBJS) $(NXDK_DIR)/lib/libnxdk_hal.lib
--- a/lib/sdl/Makefile
+++ b/lib/sdl/Makefile
@@ -1,5 +1,8 @@
 SDL_TTF_DIR = $(NXDK_DIR)/lib/sdl/SDL_ttf
 SDL_TTF_SRCS += $(SDL_TTF_DIR)/SDL_ttf.c
+SDL_TTF_OBJS = $(addsuffix .obj, $(basename $(SDL_TTF_SRCS)))
+
+$(NXDK_DIR)/lib/libSDL_ttf.lib: $(SDL_TTF_OBJS)
 
 FREETYPE_DIR = $(NXDK_DIR)/lib/sdl/SDL_ttf/external/freetype-2.4.12
 FREETYPE_SRCS += $(FREETYPE_DIR)/src/base/ftbase.c
@@ -13,9 +16,22 @@ FREETYPE_SRCS += $(FREETYPE_DIR)/src/raster/raster.c
 FREETYPE_SRCS += $(FREETYPE_DIR)/src/smooth/smooth.c
 FREETYPE_SRCS += $(FREETYPE_DIR)/src/autofit/autofit.c
 FREETYPE_SRCS += $(FREETYPE_DIR)/src/sfnt/sfnt.c
+FREETYPE_OBJS = $(addsuffix .obj, $(basename $(FREETYPE_SRCS)))
 
-SRCS += $(SDL_TTF_SRCS) $(FREETYPE_SRCS)
+$(NXDK_DIR)/lib/libfreetype.lib: $(FREETYPE_OBJS)
+
 CFLAGS += -I$(NXDK_DIR)/lib/sdl \
           -I$(SDL_TTF_DIR) \
           -I$(FREETYPE_DIR)/include \
           -DXBOX
+
+main.exe: $(NXDK_DIR)/lib/libSDL_ttf.lib $(NXDK_DIR)/lib/libfreetype.lib
+
+CLEANRULES += clean-sdl_ttf
+
+.PHONY: clean-sdl_ttf
+clean-sdl_ttf:
+	$(VE)rm -f $(SDL_TTF_OBJS) \
+	           $(FREETYPE_OBJS) \
+	           $(NXDK_DIR)/lib/libSDL_ttf.lib \
+	           $(NXDK_DIR)/lib/libfreetype.lib
--- a/lib/usb/Makefile
+++ b/lib/usb/Makefile
@@ -1,0 +1,32 @@
+USB_SRCS := \
+	$(NXDK_DIR)/lib/usb/host/ohci-hcd.c \
+	$(NXDK_DIR)/lib/usb/core/message.c \
+	$(NXDK_DIR)/lib/usb/core/hcd.c \
+	$(NXDK_DIR)/lib/usb/core/hcd-pci.c \
+	$(NXDK_DIR)/lib/usb/core/hub.c \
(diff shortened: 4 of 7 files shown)


dracc:
NevoX shrunk by 0.2MiB, no files were added to the .iso - did I miss something or does this do magic?

I also noted that I needed to add a link to `llvm-lib-8` named `llvm-lib` in my $PATH on Debian testing (Bullseye) - might be worth adding a note about that somewhere.

thrimbor:
(This was mostly answered in Discord already by others, but I'll still reply here for completeness)

> NevolutionX.xbe shrunk by 0.2MiB, no files were added to the .iso - did I miss something or does this do magic?

This is expected, and one of the reasons behind this change. Atm the linker just links together all object files we pass to it, resulting in lots of dead code ending up in the binary. When using static libraries, lld is able to grab only the required object files from the libraries, resulting in the size reduction. When we switch all of nxdk to be built as libraries, I expect even higher gains.

> I also noted that I needed to add a link to `llvm-lib-8` named `llvm-lib` in my $PATH on Debian testing (Bullseye) - might be worth adding a note about that somewhere.

I assume this is an issue of your setup - I didn't test on Debian, but the CI ran it on Ubuntu, macOS and Windows, where it worked fine.

thrimbor:
> Sorry, I forgot for a moment that Debian is not one of this projects supported OSs, please disregard my comment about `llvm-lib`.

@dracc I'm certainly interested in having nxdk build properly on Debian. What I meant was that I was suspecting it to be an issue of your specific setup - I have since tried the Debian Docker images though, and found out that it requires you to install the "llvm" package (in addition to clang and lld) to work. I added Debian-specific instructions to the Wiki.

Interestingly, this doesn't seem to be an issue on Ubuntu - CI just installs the llvm package as a dependency.

lib/hal/Makefile
```diff
@@ -0,0 +1,22 @@
...
+	$(NXDK_DIR)/lib/hal/fileio.c \
+	$(NXDK_DIR)/lib/hal/input.c \
+	$(NXDK_DIR)/lib/hal/io.c \
+	$(NXDK_DIR)/lib/hal/led.c \
+	$(NXDK_DIR)/lib/hal/video.c \
+	$(NXDK_DIR)/lib/hal/xbox.c
+
+HAL_OBJS = $(addsuffix .obj, $(basename $(HAL_SRCS)))
+
+$(NXDK_DIR)/lib/libnxdkhal.lib: $(HAL_OBJS)
```
JayFoxRox: I'm personally not a big fan of stacking acronyms; I'd suggest libnxdk_hal.lib and libnxdk_usb.lib instead.

thrimbor: Done.

JayFoxRox:
This has a conflict now.

mborgerson:
Which version of Ubuntu are you testing with? `llvm-lib` symlink doesn't exist for me on Ubuntu 18.04.

lib/usb/Makefile
```diff
@@ -0,0 +1,34 @@
...
+	$(NXDK_DIR)/lib/usb/sys/xremote.c \
+	$(NXDK_DIR)/lib/usb/sys/usbkey.c \
+	$(NXDK_DIR)/lib/usb/sys/usbmouse.c \
+	$(NXDK_DIR)/lib/usb/misc/misc.c \
+	$(NXDK_DIR)/lib/usb/misc/pci.c \
+	$(NXDK_DIR)/lib/usb/misc/malloc.c
+
+USB_OBJS = $(addsuffix .obj, $(basename $(USB_SRCS)))
+
+$(NXDK_DIR)/lib/libnxdk_usb.lib: $(USB_OBJS)
```
mborgerson: We shouldn't need to define the steps (including the echo line and lib command) in every Makefile, this could be moved out into a generic recipe in the top level Makefile

thrimbor:
> Which version of Ubuntu are you testing with? `llvm-lib` symlink doesn't exist for me on Ubuntu 18.04.

I've only tested Ubuntu via Travis CI, that should be 16.04. Do you have the package "llvm" installed (clang and lld don't seem to pull that in on Debian)?

lib/Makefile-pdclib
```diff
@@ -0,0 +1,18 @@
+PDCLIB_SRCS := \
```
mborgerson: Why isn't this in pdclib repository?

mborgerson:
@thrimbor 

> I've only tested Ubuntu via Travis CI, that should be 16.04

Why are you testing with such an old version? You can [add a line in your config file](https://docs.travis-ci.com/user/reference/bionic/) to request that a more recent version be used.

> Do you have the package "llvm" installed

Ya:
```
$ dpkg -l | grep llvm
ii  libllvm4.0:amd64                              1:4.0.1-10                                   amd64        Modular compiler and toolchain technologies, runtime library
ii  libllvm6.0:amd64                              1:6.0-1ubuntu2                               amd64        Modular compiler and toolchain technologies, runtime library
ii  libllvm7:amd64                                1:7-3~ubuntu0.18.04.1                        amd64        Modular compiler and toolchain technologies, runtime library
ii  libllvm7:i386                                 1:7-3~ubuntu0.18.04.1                        i386         Modular compiler and toolchain technologies, runtime library
ii  libllvm8:amd64                                1:8.0.1~svn369350-1~exp1~20190820121219.79   amd64        Modular compiler and toolchain technologies, runtime library
ii  llvm                                          1:6.0-41~exp5~ubuntu1                        amd64        Low-Level Virtual Machine (LLVM)
ii  llvm-4.0                                      1:4.0.1-10                                   amd64        Modular compiler and toolchain technologies
ii  llvm-4.0-dev                                  1:4.0.1-10                                   amd64        Modular compiler and toolchain technologies, libraries and headers
ii  llvm-4.0-runtime                              1:4.0.1-10                                   amd64        Modular compiler and toolchain technologies, IR interpreter
ii  llvm-6.0                                      1:6.0-1ubuntu2                               amd64        Modular compiler and toolchain technologies
ii  llvm-6.0-dev                                  1:6.0-1ubuntu2                               amd64        Modular compiler and toolchain technologies, libraries and headers
ii  llvm-6.0-runtime                              1:6.0-1ubuntu2                               amd64        Modular compiler and toolchain technologies, IR interpreter
ii  llvm-7                                        1:7-3~ubuntu0.18.04.1                        amd64        Modular compiler and toolchain technologies
ii  llvm-7-dev                                    1:7-3~ubuntu0.18.04.1                        amd64        Modular compiler and toolchain technologies, libraries and headers
ii  llvm-7-runtime                                1:7-3~ubuntu0.18.04.1                        amd64        Modular compiler and toolchain technologies, IR interpreter
ii  llvm-8                                        1:8.0.1~svn369350-1~exp1~20190820121219.79   amd64        Modular compiler and toolchain technologies
ii  llvm-8-dev                                    1:8.0.1~svn369350-1~exp1~20190820121219.79   amd64        Modular compiler and toolchain technologies, libraries and headers
ii  llvm-8-runtime                                1:8.0.1~svn369350-1~exp1~20190820121219.79   amd64        Modular compiler and toolchain technologies, IR interpreter
ii  llvm-runtime                                  1:6.0-41~exp5~ubuntu1                        amd64        Low-Level Virtual Machine (LLVM), bytecode interpreter
```

thrimbor:
> > I've only tested Ubuntu via Travis CI, that should be 16.04
> 
> Why are you testing with such an old version? You can [add a line in your config file](https://docs.travis-ci.com/user/reference/bionic/) to request that a more recent version be used.

It's just the version nxdk currently uses. If I remember correctly the newer images weren't available when we first added CI.

I looked into the issue in a Ubuntu Docker image, and it's a bug in Ubuntu's llvm-package, which I reported [upstream](https://bugs.launchpad.net/ubuntu/+source/llvm-defaults/+bug/1846275). As a temporary workaround, `PATH=$PATH:/usr/lib/llvm-6.0/bin` does the trick.

lib/usb/Makefile
thrimbor: Done.

lib/sdl/Makefile
```diff
@@ -13,9 +16,22 @@ FREETYPE_SRCS += $(FREETYPE_DIR)/src/raster/raster.c
...
+
+$(NXDK_DIR)/lib/libfreetype.lib: $(FREETYPE_OBJS)
 
-SRCS += $(SDL_TTF_SRCS) $(FREETYPE_SRCS)
 CFLAGS += -I$(NXDK_DIR)/lib/sdl \
           -I$(SDL_TTF_DIR) \
           -I$(FREETYPE_DIR)/include \
           -DXBOX
+
+main.exe: $(NXDK_DIR)/lib/libSDL_ttf.lib $(NXDK_DIR)/lib/libfreetype.lib
```
JayFoxRox: What main.exe is this?

thrimbor: The `main.exe` of whatever sample you're building at that moment. We're just adding a library dependecy here (only if SDL was requested ofc).

As we move away from the "one big makefile handles everything" approach, I expect samples to list their required libs explicitly in their own Makefile (with proper library directory set in the linker flags etc.) in the future, this is only for the transitioning phase.

JayFoxRox: I'd have expected libraries to be linked like libxboxkrnl.lib (which is linked from the Makefile in the nxdk root).

I didn't even know you could add more dependencies to the same output using this (I thought each main.exe line would create a separate target).

However, suspecting this would add dependencies to application default.xbe, my comment was also meant to hint at the fact that a library Makefile should be independent of the users project.

> this is only for the transitioning phase.

Fair enough. In that case this is fine with me.

JayFoxRox:
> I looked into the issue in a Ubuntu Docker image, and it's a bug in Ubuntu's llvm-package, which I reported upstream. As a temporary workaround, PATH=$PATH:/usr/lib/llvm-6.0/bin does the trick.

Upstream apparently has no intention of fixing this themselves. I also contacted doko (who uploaded the current revision) on freenode and they responded with: "https://wiki.ubuntu.com/StableReleaseUpdates" - I believe they imply it does not qualify or we should contact the respective SRU people.

We should have an issue about this after merge.
A possible workaround should be added to the wiki after merge.

---

I think this is fine and it should be merged.

LukeUsher:
Nice work, I think we can merge this as-is.

@thrimbor are you planniing on doing this for the rest of the libraries as a seperate PR?

thrimbor:
@LukeUsher Yes, I plan to have all libraries actually build as libraries eventually. Some are more complicated to adjust though (like SDL, which has its Makefile in the submodule, so it requires an update to the submodule *and* nxdk before it can work) or were/are undergoing larger changes currently, so this PR is intentionally "incomplete" to avoid problems.

PR #157, opened 2019-08-26
Travis CI for Windows/msys2

thrimbor:
Based on #156.

The only change compared to #156 is b9f3204d6eac7c762cf4c4e2cf7ce03005be8505, which adds support for building with msys2 on Windows with Travis CI. It uses the Chocolatey package manager for msys2 installation (also includes a hang workaround, see https://travis-ci.community/t/build-times-out-when-msys-was-installed/3498), the additional batch files were required because of whitespace issues when calling bash directly from `.travis.yml`.

Due to slower building and the time required to install msys2, this runs quite a bit slower than the Linux build and is even about a minute slower than the macOS build, but it works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os:
     - linux
     - osx
+    - windows
 
 dist: xenial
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -9,12 +10,18 @@ sudo: required
...
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get update -qq; fi
+    - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then choco install msys2; fi
 
 install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install llvm; fi
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -qq flex bison clang lld-6.0; fi
+    - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then cmd /c\ .travis_win_deps.bat; fi
+    # Build hang workaround
+    - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then taskkill -IM "gpg-agent.exe" -F; fi
```
JayFoxRox: Please provide more inforrmation about this workaround (like the URL which describes it) in the comment. Ideally there'd be a bugtracker issue we could track.

thrimbor: I added the url. Unfortunately there doesn't seem to be a bugtracker - well, [there is one](https://github.com/travis-ci/travis-ci), but it's not supposed to be used anymore

GXTX: I'm assuming it's not running into this issue on #168 because I'm having to install via chocolaty which has it's own install scripts separate to what AppVeyor sets up.

JayFoxRox:
This is still conflicting - it needs a rebase.

thrimbor:
Oops, completely forgot about updating this.
The conflicts are fixed now.

.travis_win_deps.bat
```diff
@@ -0,0 +1 @@
+C:\tools\msys64\usr\bin\bash.exe -lc "pacman --needed --noconfirm -S make bison flex mingw-w64-x86_64-llvm mingw-w64-x86_64-clang mingw-w64-x86_64-lld"
```
JayFoxRox: The build failed with`/bin/sh: line 2: cmake: command not found`.

https://github.com/XboxDev/nxdk/wiki/Install-the-Prerequisites must also still be updated for https://github.com/XboxDev/nxdk/pull/144#issuecomment-525822616

thrimbor: It's fixed now, and I updated the wiki as well.
I checked that it builds now on Travis, the GitHub check didn't complete yet though (there currently seems to be an issue that prevents the macOS CI VM from starting).

.travis.yml
JayFoxRox: The URL was removed again?

PR #158, opened 2019-09-01
pbkit: pb_push refactor

dracc:
- Remove the old `pb_push` macro
- Remove some of the repeated code
- ~~Define functions as `inline`~~
- Return a new pointer, remove the need for manual pointer incrementation
- Update the `mesh` sample to use these new capabilities
- Update the `triangle` sample to use these new capabilities

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -2205,7 +2205,7 @@ DWORD *pb_begin(void)
 #ifdef DBG
     if (pb_Put>=pb_Tail) debugPrint("ERROR! Push buffer overflow! Use pb_reset more often or enlarge push buffer!\n");
 
-    if (pb_BeginEndPair==1) debugPrint("pb_start without a pb_end earlier\n");
+    if (pb_BeginEndPair==1) debugPrint("pb_begin without a pb_end earlier\n");
     pb_BeginEndPair=1;
     pb_PushIndex=0;
     pb_PushNext=pb_Put;
@@ -2271,7 +2271,7 @@ void pb_end(DWORD *pEnd)
     }
     if (pb_BeginEndPair==0)
     {
-        debugPrint("pb_end without a pb_start\n");
+        debugPrint("pb_end without a pb_begin\n");
         assert(false);
     }
     pb_BeginEndPair=0;
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -2264,7 +2264,16 @@ void pb_end(DWORD *pEnd)
 #endif
 
 #ifdef DBG
-    if (pb_BeginEndPair==0) debugPrint("pb_end without a pb_start\n");
+    if (pEnd!=pb_PushNext)
+    {
+        debugPrint("pb_end: input pointer invalid or not following previous write addresses\n");
+        assert(false);
+    }
+    if (pb_BeginEndPair==0)
+    {
+        debugPrint("pb_end without a pb_start\n");
+        assert(false);
+    }
     pb_BeginEndPair=0;
 #endif
 
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -2319,83 +2319,48 @@ void pb_pushto(DWORD subchannel, DWORD *p, DWORD command, DWORD nparam)
     *(p+0)=EncodeMethod(subchannel,command,nparam);
 }
 
-void pb_push1to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1)
+DWORD *pb_push1to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push1to: new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push1to: missing pb_begin earlier\n");
-    pb_PushIndex+=2;
-    pb_PushNext+=2;
-    if (pb_PushIndex>128) debugPrint("pb_push1to: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(subchannel,command,1);
+    pb_pushto(subchannel,p,command,1);
     *(p+1)=param1;
+    return p+2;
 }
 
-void pb_push2to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2)
+DWORD *pb_push2to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push2to : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push2to : missing pb_begin earlier\n");
-    pb_PushIndex+=3;
-    pb_PushNext+=3;
-    if (pb_PushIndex>128) debugPrint("pb_push2to: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(subchannel,command,2);
+    pb_pushto(subchannel,p,command,2);
     *(p+1)=param1;
     *(p+2)=param2;
+    return p+3;
 }
 
-void pb_push3to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3)
+DWORD *pb_push3to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push3to : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push3to : missing pb_begin earlier\n");
-    pb_PushIndex+=4;
-    pb_PushNext+=4;
-    if (pb_PushIndex>128) debugPrint("pb_push3to: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(subchannel,command,3);
+    pb_pushto(subchannel,p,command,3);
     *(p+1)=param1;
     *(p+2)=param2;
     *(p+3)=param3;
+    return p+4;
 }
 
-void pb_push4to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4)
+DWORD *pb_push4to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push4to : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push4to : missing pb_begin earlier\n");
-    pb_PushIndex+=5;
-    pb_PushNext+=5;
-    if (pb_PushIndex>128) debugPrint("pb_push4to: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(subchannel,command,4);
+    pb_pushto(subchannel,p,command,4);
     *(p+1)=param1;
     *(p+2)=param2;
     *(p+3)=param3;
     *(p+4)=param4;
+    return p+5;
 }
 
-void pb_push4fto(DWORD subchannel, DWORD *p, DWORD command, float param1, float param2, float param3, float param4)
+DWORD *pb_push4fto(DWORD subchannel, DWORD *p, DWORD command, float param1, float param2, float param3, float param4)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push4fto : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push4fto : missing pb_begin earlier\n");
-    pb_PushIndex+=5;
-    pb_PushNext+=5;
-    if (pb_PushIndex>128) debugPrint("pb_push4fto: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(subchannel,command,4);
+    pb_pushto(subchannel,p,command,4);
     *((float *)(p+1))=param1;
     *((float *)(p+2))=param2;
     *((float *)(p+3))=param3;
     *((float *)(p+4))=param4;
+    return p+5;
 }
 
 void pb_push(DWORD *p, DWORD command, DWORD nparam)
@@ -2403,96 +2368,34 @@ void pb_push(DWORD *p, DWORD command, DWORD nparam)
     pb_pushto(SUBCH_3D,p,command,nparam);
 }
 
-void pb_push1(DWORD *p, DWORD command, DWORD param1)
+DWORD *pb_push1(DWORD *p, DWORD command, DWORD param1)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push1: new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push1: missing pb_begin earlier\n");
-    pb_PushIndex+=2;
-    pb_PushNext+=2;
-    if (pb_PushIndex>128) debugPrint("pb_push1: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(SUBCH_3D,command,1);
-    *(p+1)=param1;
+    return pb_push1to(SUBCH_3D,p,command,param1);
 }
 
-void pb_push2(DWORD *p, DWORD command, DWORD param1, DWORD param2)
+DWORD *pb_push2(DWORD *p, DWORD command, DWORD param1, DWORD param2)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push2 : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push2 : missing pb_begin earlier\n");
-    pb_PushIndex+=3;
-    pb_PushNext+=3;
-    if (pb_PushIndex>128) debugPrint("pb_push2: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(SUBCH_3D,command,2);
-    *(p+1)=param1;
-    *(p+2)=param2;
+    return pb_push2to(SUBCH_3D,p,command,param1,param2);
 }
 
-void pb_push3(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3)
+DWORD *pb_push3(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push3 : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push3 : missing pb_begin earlier\n");
-    pb_PushIndex+=4;
-    pb_PushNext+=4;
-    if (pb_PushIndex>128) debugPrint("pb_push3: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(SUBCH_3D,command,3);
-    *(p+1)=param1;
-    *(p+2)=param2;
-    *(p+3)=param3;
+    return pb_push3to(SUBCH_3D,p,command,param1,param2,param3);
 }
 
-void pb_push4(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4)
+DWORD *pb_push4(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push4 : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push4 : missing pb_begin earlier\n");
-    pb_PushIndex+=5;
-    pb_PushNext+=5;
-    if (pb_PushIndex>128) debugPrint("pb_push4: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(SUBCH_3D,command,4);
-    *(p+1)=param1;
-    *(p+2)=param2;
-    *(p+3)=param3;
-    *(p+4)=param4;
+    return pb_push4to(SUBCH_3D,p,command,param1,param2,param3,param4);
 }
 
-void pb_push4f(DWORD *p, DWORD command, float param1, float param2, float param3, float param4)
+DWORD *pb_push4f(DWORD *p, DWORD command, float param1, float param2, float param3, float param4)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push4f : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push4f : missing pb_begin earlier\n");
-    pb_PushIndex+=5;
-    pb_PushNext+=5;
-    if (pb_PushIndex>128) debugPrint("pb_push4f: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(SUBCH_3D,command,4);
-    *((float *)(p+1))=param1;
-    *((float *)(p+2))=param2;
-    *((float *)(p+3))=param3;
-    *((float *)(p+4))=param4;
+    return pb_push4fto(SUBCH_3D,p,command,param1,param2,param3,param4);
 }
 
-void pb_push_transposed_matrix(DWORD *p, DWORD command, float *m)
+DWORD *pb_push_transposed_matrix(DWORD *p, DWORD command, float *m)
 {
-#ifdef DBG
-    if (p!=pb_PushNext) debugPrint("pb_push_transposed_matrix : new write address invalid or not following previous write addresses\n");
-    if (pb_BeginEndPair==0) debugPrint("pb_push_transposed_matrix : missing pb_begin earlier\n");
-    pb_PushIndex+=17;
-    pb_PushNext+=17;
-    if (pb_PushIndex>128) debugPrint("pb_push_transposed_matrix : begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p++)=EncodeMethod(SUBCH_3D,command,16);
+    pb_pushto(SUBCH_3D,p++,command,16);
 
     *((float *)p++)=m[_11];
     *((float *)p++)=m[_21];
@@ -2513,9 +2416,9 @@ void pb_push_transposed_matrix(DWORD *p, DWORD command, float *m)
     *((float *)p++)=m[_24];
     *((float *)p++)=m[_34];
     *((float *)p++)=m[_44];
+
+    return p;
 }
-
-
 
 
 
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -1863,16 +1863,16 @@ void pb_target_back_buffer(void)
     dma_addr|=3;
 
     p=pb_begin();
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_WAIT_MAKESPACE,0); p+=2;
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x08,dma_addr); p+=3; //set params addr,data
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2; //calls subprogID PB_SETOUTER: does VIDEOREG(addr)=data
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x0C,dma_addr); p+=3;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2;
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x00,dma_flags); p+=3;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2;
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x04,dma_limit); p+=3;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT3,9); p+=2;
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_WAIT_MAKESPACE,0);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x08,dma_addr); //set params addr,data
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); //calls subprogID PB_SETOUTER: does VIDEOREG(addr)=data
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x0C,dma_addr);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x00,dma_flags);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID9Inst<<4)+0x04,dma_limit);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT3,9);
     pb_end(p);
 
     //DMA channel 11 is used by GPU in order to bitblt images
@@ -1882,16 +1882,16 @@ void pb_target_back_buffer(void)
     dma_addr|=3;
     
     p=pb_begin();
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_WAIT_MAKESPACE,0); p+=2;
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x08,dma_addr); p+=3; //set params addr,data
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2; //calls subprogID PB_SETOUTER: does VIDEOREG(addr)=data
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x0C,dma_addr); p+=3;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2;
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x00,dma_flags); p+=3;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2;
-    pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x04,dma_limit); p+=3;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); p+=2;
-    pb_push1_to(SUBCH_4,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT2,11); p+=2;
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_WAIT_MAKESPACE,0);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x08,dma_addr); //set params addr,data
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); //calls subprogID PB_SETOUTER: does VIDEOREG(addr)=data
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x0C,dma_addr);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x00,dma_flags);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x04,dma_limit);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push1_to(SUBCH_4,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT2,11);
     pb_end(p);
 
     depth_stencil=1;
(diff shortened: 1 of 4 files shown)


lib/pbkit/pbkit.h
```diff
@@ -51,7 +51,7 @@ extern "C"
 #define SUBCH_4                 4
 
 //fastest way to write a method for subchannel 3D (0)
-#define pb_push(p,command,nparam)       *(p)=((nparam<<18)+command)
+//#define pb_push(p,command,nparam)       *(p)=((nparam<<18)+command)
```
JayFoxRox: Remove dead code

lib/pbkit/pbkit.h
```diff
@@ -67,25 +67,42 @@ DWORD   pb_wait_for_vbl(void);  //waits for VBL, returns VBL counter value as bo
 void    pb_erase_depth_stencil_buffer(int x, int y, int w, int h);
 
 void    pb_reset(void); //forces a jump to push buffer head (do it at frame start)
-int pb_finished(void);  //prepare screen swapping at VBlank (do it at frame end)
+int  pb_finished(void); //prepare screen swapping at VBlank (do it at frame end)
```
JayFoxRox: Unintentional change?

lib/pbkit/pbkit.h
```diff
@@ -67,25 +67,42 @@ DWORD   pb_wait_for_vbl(void);  //waits for VBL, returns VBL counter value as bo
...
-void    pb_push1(DWORD *p, DWORD command, DWORD param1); //slow functions but with debug messages (targets SUBCH_3D (0))
-void    pb_push2(DWORD *p, DWORD command, DWORD param1, DWORD param2);
-void    pb_push3(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3);
-void    pb_push4(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4);
-void    pb_push4f(DWORD *p, DWORD command, float param1, float param2, float param3, float param4);
-void    pb_push_transposed_matrix(DWORD *p, DWORD command, float *m);
-void    pb_end(DWORD *pEnd);    //end a block with this (triggers the data sending to GPU)
+
+inline
+void    pb_pushto(DWORD subchannel, DWORD *p, DWORD command, DWORD nparam);
```
JayFoxRox: I'd put the inline on the same line as the rest of the declaration

lib/pbkit/pbkit.c
```diff
@@ -2287,183 +2287,126 @@ void pb_end(DWORD *pEnd)
             }
         }
     }
+    return NULL;
 }
 
-
-
-void pb_push1to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1)
+inline void pb_pushto(DWORD subchannel, DWORD *p, DWORD command, DWORD nparam)
```
JayFoxRox: These should probably still return the next valid `p`?

lib/pbkit/pbkit.c
```diff
@@ -2234,7 +2234,7 @@ void pb_stop_log(void)
 #endif
 
 
-void pb_end(DWORD *pEnd)
+DWORD *pb_end(DWORD *pEnd)
```
JayFoxRox: This function should also check the correctness of `pEnd` (same conditions as for `pb_pushto` input apply: it must be at the expected location + the expected `pb_PushNext` location should probably be made invalid so it's not used errornously).

lib/pbkit/pbkit.c
```diff
@@ -2287,183 +2287,126 @@ void pb_end(DWORD *pEnd)
...
-    pb_PushNext+=4;
-    if (pb_PushIndex>128) debugPrint("pb_push3to: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(subchannel,command,3);
+    pb_pushto(subchannel,p,command,3);
     *(p+1)=param1;
     *(p+2)=param2;
     *(p+3)=param3;
 }
```
JayFoxRox: Return value missing?

*(Review other instances, too)*

lib/pbkit/pbkit.c
```diff
@@ -2287,183 +2287,126 @@ void pb_end(DWORD *pEnd)
...
-    pb_PushNext+=4;
-    if (pb_PushIndex>128) debugPrint("pb_push3: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(SUBCH_3D,command,3);
+    pb_push(p,command,3);
     *(p+1)=param1;
     *(p+2)=param2;
     *(p+3)=param3;
+    return p+4;
```
JayFoxRox: This stuff can all be reduced to `return pb_push3to(...)` (and equivalent) with the subchannel being set to the 3D one.

lib/pbkit/pbkit.h
```diff
@@ -67,25 +67,42 @@ DWORD   pb_wait_for_vbl(void);  //waits for VBL, returns VBL counter value as bo
...
+inline
+DWORD   *pb_push3to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3);
+inline
+DWORD   *pb_push4to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4);
+
+inline
+void    pb_push(DWORD *p, DWORD command, DWORD nparam);
+
+inline
+DWORD   *pb_push1(DWORD *p, DWORD command, DWORD param1); //slow functions but with debug messages (targets SUBCH_3D (0))
```
JayFoxRox: I'd remove the comment about this being slow. It's no longer true, and it always used to be very vague. These functions were never slow, unless debugging was enabled.

lib/pbkit/pbkit.c
```diff
@@ -3383,60 +3316,55 @@ int pb_init(void)
...
+    p=pb_push2to(SUBCH_4,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT1,3,11);
+    p=pb_end(p); //calls pb_start() which will trigger the reading and sending to GPU (asynchronous, no waiting)
 
     //setup needed for color computations
     p=pb_begin();
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_SET_OBJECT0,3);
-    *(p++)=2;
-    *(p++)=3;
-    *(p++)=3;
+    p=pb_push3(p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT0,2,3,3);
```
JayFoxRox: Avoid this step here. The GPU will auto-increment the command index, so it's possible to send multiple commands with on header. We should handle this differently, so it's more obvious which commands are send.

I'm not sure about this particular case, but it could be up to 3 different individual commands, so we should not merge them into one line (it hides the intention). Keep the `pb_push` variant for now.

JayFoxRox: Note: It's generally fine when only submitting 1 parameter, because then we know it's only 1 command being submitted. It's only problematic for 2 or more parameters as they *could* (and likely will) belong to different commands.

lib/pbkit/pbkit.c
```diff
@@ -2549,14 +2492,9 @@ void pb_fill(int x, int y, int w, int h, DWORD color)
...
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,2);     //sets rectangle coordinates
-    *(p++)=((x2-1)<<16)|x1;
-    *(p++)=((y2-1)<<16)|y1;
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_DEPTH,3);     //sets data used to fill in rectangle
-    *(p++)=0;           //(depth<<8)|stencil
-    *(p++)=color;           //color
-    *(p++)=0xF0;            //triggers the HW rectangle fill (0x03 for D&S)
-    pb_end(p);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,((x2-1)<<16)|x1,((y2-1)<<16)|y1); //sets rectangle coordinates
+    p=pb_push3(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_DEPTH,0,color,0xF0); //sets data used to fill in rectangle
```
JayFoxRox: This is a case where 3 different commands are being submitted. \- so avoid this step.

*(See my other note about merging `pb_push` into a single line)*

lib/pbkit/pbkit.c
```diff
@@ -2580,14 +2518,9 @@ void pb_erase_depth_stencil_buffer(int x, int y, int w, int h)
...
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,2);     //sets rectangle coordinates
-    *(p++)=((x2-1)<<16)|x1;
-    *(p++)=((y2-1)<<16)|y1;
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_DEPTH,3);     //sets data used to fill in rectangle
-    *(p++)=0xffffff00;      //(depth<<8)|stencil
-    *(p++)=0;           //color
-    *(p++)=0x03;            //triggers the HW rectangle fill (only on D&S)
-    pb_end(p);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,((x2-1)<<16)|x1,((y2-1)<<16)|y1); //sets rectangle coordinates
+    p=pb_push3(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_DEPTH,0xffffff00,0,0x03); //sets data used to fill in rectangle
```
JayFoxRox: 3 different commands

lib/pbkit/pbkit.c
```diff
@@ -2580,14 +2518,9 @@ void pb_erase_depth_stencil_buffer(int x, int y, int w, int h)
...
     p=pb_begin();
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,2);     //sets rectangle coordinates
-    *(p++)=((x2-1)<<16)|x1;
-    *(p++)=((y2-1)<<16)|y1;
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_DEPTH,3);     //sets data used to fill in rectangle
-    *(p++)=0xffffff00;      //(depth<<8)|stencil
-    *(p++)=0;           //color
-    *(p++)=0x03;            //triggers the HW rectangle fill (only on D&S)
-    pb_end(p);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,((x2-1)<<16)|x1,((y2-1)<<16)|y1); //sets rectangle coordinates
```
JayFoxRox: 2 different commands (`NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ` and the vertical equvialent)

lib/pbkit/pbkit.c
```diff
@@ -2549,14 +2492,9 @@ void pb_fill(int x, int y, int w, int h, DWORD color)
...
     p=pb_begin();
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,2);     //sets rectangle coordinates
-    *(p++)=((x2-1)<<16)|x1;
-    *(p++)=((y2-1)<<16)|y1;
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_DEPTH,3);     //sets data used to fill in rectangle
-    *(p++)=0;           //(depth<<8)|stencil
-    *(p++)=color;           //color
-    *(p++)=0xF0;            //triggers the HW rectangle fill (0x03 for D&S)
-    pb_end(p);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_CLEAR_VALUE_HORIZ,((x2-1)<<16)|x1,((y2-1)<<16)|y1); //sets rectangle coordinates
```
JayFoxRox: 2 commands (see other instance below)

JayFoxRox:
[I left some more comments on Discord #nxdk](https://discordapp.com/channels/428359196719972353/428360618102226946/617872451858333696) - I assume they'll be integrated during review fixups, so I won't relay them here.

See my code comments for most of my important thoughts. Additionally, the commit "pbkit: Make pb_push* return a pointer position" should be renamed to hint *which* pointer position is returned ("next command word" / "end of buffer"? - something along those lines).

lib/pbkit/pbkit.c
dracc: All but `pb_push4f` fixed due to it's type casting stuff.

lib/pbkit/pbkit.c
dracc: Should be fixed now.

lib/pbkit/pbkit.c
dracc: Should be fixed now.

lib/pbkit/pbkit.h
dracc: True. Should be fixed now.

lib/pbkit/pbkit.h
dracc: (Probably) fixed now.

lib/pbkit/pbkit.h
dracc: Done. :+1:

lib/pbkit/pbkit.h
dracc: Done. :+1:

lib/pbkit/pbkit.c
JayFoxRox: We should potentially consider to add those `push4fto` for completeness *in the future*.

dracc: Consider it done. :+1:

lib/pbkit/pbkit.c
dracc: Reverted. :+1:

lib/pbkit/pbkit.c
dracc: Reverted. :+1:

lib/pbkit/pbkit.c
dracc: Reverted. :+1:

lib/pbkit/pbkit.c
dracc: Reverted. :+1:

lib/pbkit/pbkit.c
dracc: Reverted. :+1:

lib/pbkit/pbkit.c
dracc: I believe this has been taken care of now.

lib/pbkit/pbkit.c
```diff
@@ -2353,32 +2353,32 @@ inline DWORD *pb_push4fto(DWORD subchannel, DWORD *p, DWORD command, float param
...
 
 inline DWORD *pb_push2(DWORD *p, DWORD command, DWORD param1, DWORD param2)
 {
-    return pb_push2to(SUBCH_3D,p,command,param1,param2);
+    return pb_push2_to(SUBCH_3D,p,command,param1,param2);
 }
 
 inline DWORD *pb_push3(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3)
 {
-    return pb_push3to(SUBCH_3Dp,command,param1,param2,param3);
```
JayFoxRox: This had a bug in an earlier commit?

lib/pbkit/pbkit.c
```diff
@@ -2260,8 +2260,10 @@ DWORD *pb_end(DWORD *pEnd)
 #endif
 
 #ifdef DBG
+    if (pEnd!=pb_PushNext) debugPrint("pb_end: input pointer invalid o r not following previous write addresses\n");
```
JayFoxRox: "o r"

lib/pbkit/pbkit.c
```diff
@@ -2260,8 +2260,10 @@ DWORD *pb_end(DWORD *pEnd)
...
     if (pb_BeginEndPair==0) debugPrint("pb_end without a pb_start\n");
     pb_BeginEndPair=0;
+    pb_PushNext=NULL;
```
JayFoxRox: nit: actually we don't need this `NULL`-ing, because `pb_BeginEndPair` probably guards all use? Otherwise, for completeness, we'd probably also want to `NULL` the other globals (I believe `pb_PushIndex`?)

lib/pbkit/pbkit.c
dracc: That much is true. I'll try to sort it out in that earlier commit. :+1:

lib/pbkit/pbkit.c
dracc: I feel like we'd rather have less NULLings than more - removed the `pb_PushNext=NULL;` line.

lib/pbkit/pbkit.c
dracc: Great catch, fixed!

JayFoxRox:
Some commits are weirdly structured, such as  `pbkit: Make pb_push* return the end of buffer pointer` also containing the redirection from `push*` to `push*to`.
I'd also suggest squashing `pbkit: Add pb_push4fto` and the change where `push*` redirects to `push*to`, so it immediately has 1 user.
We can also consider `pb_push_transposed_matrix_to` in the same commit, as it's the only function which doesn't have a `push*to` variant (although.. not really necessary either).

Ideally `pbkit: Rename pb_push*to -> pb_push*_to` would also be the very first commit, however, I can see how restructuring this will also be very annoying (unless you rebuild the entire history manually). I'm also not a big fan of `->` in commit titles, but I don't have a better symbol or word to use in this context either (seeing how `Rename *to to *_to` isn't very nice either).

Aside from that, the code looks good. There's a couple of things we *could* improve, but we can also do them in follow-ups (more checks in `pb_end` for example).

---

We need someone to test this.

Ideally someone else would do the final review as I was quite involved with the design of this, including the new style of `p=pb_push*()` and even `p=pb_end()` (which is unlike C `free`, so I wouldn't be surprised if someone was offended).

We can refactor the other samples in a follow-up PR (unless any of them broke - which shouldn't be possible?).

lib/pbkit/pbkit.c
```diff
@@ -2365,6 +2365,23 @@ void pb_push4to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD p
     *(p+4)=param4;
 }
 
+void pb_push4fto(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4)
```
JayFoxRox: Argument types wrong in commit `pbkit: Add pb_push4fto function `

JayFoxRox:
Rename `pbkit: pb_push refactoring` commit to `pbkit: pb_push* refactoring` to make clear that it's a bigger refactor of all involved functions.

lib/pbkit/pbkit.c
```diff
@@ -3349,33 +3349,33 @@ int pb_init(void)
     *(p++)=12;
     pb_push(p++,NV20_TCL_PRIMITIVE_3D_ACTIVATE_COLORS,1);
```
JayFoxRox: I noticed this on by accident - looks like it could be `pb_push1` (as part of commit `pbkit: Use new pb_push* in pbkit functions`)? The one before this, too.
However, also acceptable to do this in a full pbkit refactor (as it's easy to introduce errors again).

JayFoxRox:
Thanks for taking the time to refactor the history - it's much cleaner now.
Some errors slipped in (which even broke CI), but some quick fixups in the respective commits will fix this easily.

Once fixed, this should be merged after a quick-test / quick-review.

lib/pbkit/pbkit.c
dracc: Great find, thanks!
Fixed.

JayFoxRox: CI actually had an error because of this. I didn't have to look very hard.

lib/pbkit/pbkit.c
dracc: Good catch. I believe I had touched these two in an earlier attempt. They're the only ones using the same command and having a "standard" number of parameters not using one of the helper functions. Ought to be fixed now. :+1:

samples/mesh/main.c
```diff
@@ -301,33 +301,30 @@ static void init_shader(void)
...
+    p=pb_push1(p, NV097_SET_TRANSFORM_PROGRAM_LOAD, 0);
 
     for (i=0; i<sizeof(vs_program)/8; i++) {
         pb_push(p++, NV097_SET_TRANSFORM_PROGRAM, 4);
         memcpy(p, &vs_program[i*4], 4*4);
         p+=4;
     }
 
-    pb_end(p);
+    p=pb_end(p);
```
mborgerson: Ok but not really necessary

samples/mesh/main.c
```diff
@@ -301,33 +301,30 @@ static void init_shader(void)
...
 
     /* Setup fragment shader */
     p = pb_begin();
     #include "ps.inl"
-    pb_end(p);
+    p=pb_end(p);
```
mborgerson: Ok but not really necessary. We never use this pointer after here. Same comment for all instances.

lib/pbkit/pbkit.c
mborgerson: Why does this return a value (always `null`) now?

lib/pbkit/pbkit.c
```diff
@@ -2287,163 +2288,101 @@ void pb_end(DWORD *pEnd)
...
+    if (p!=pb_PushNext) debugPrint("pb_push_to: new write address invalid or not following previous write addresses\n");
+    if (pb_BeginEndPair==0) debugPrint("pb_push_to: missing pb_begin earlier\n");
+    pb_PushIndex += 1 + nparam;
+    pb_PushNext += 1 + nparam;
+    if (pb_PushIndex>128) debugPrint("pb_push_to: begin-end block musn't exceed 128 dwords\n");
 #endif
 
-    *(p+0)=EncodeMethod(subchannel,command,1);
-    *(p+1)=param1;
+    p=EncodeMethod(subchannel,command,nparam);
```
mborgerson: This doesn't look right. `EncodeMethod` doesn't return a pointer

lib/pbkit/pbkit.h
```diff
@@ -75,17 +73,20 @@ void pb_wait_until_gr_not_busy(void);
...
-void    pb_push3to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3);
-void    pb_push4to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4);
-void    pb_push1(DWORD *p, DWORD command, DWORD param1); //slow functions but with debug messages (targets SUBCH_3D (0))
-void    pb_push2(DWORD *p, DWORD command, DWORD param1, DWORD param2);
-void    pb_push3(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3);
-void    pb_push4(DWORD *p, DWORD command, DWORD param1, DWORD param2, DWORD param3, DWORD param4);
-void    pb_push4f(DWORD *p, DWORD command, float param1, float param2, float param3, float param4);
-void    pb_push_transposed_matrix(DWORD *p, DWORD command, float *m);
-void    pb_end(DWORD *pEnd);    //end a block with this (triggers the data sending to GPU)
+inline DWORD   *pb_push_to(DWORD subchannel, DWORD *p, DWORD command, DWORD nparam);
```
mborgerson: Compiler is likely to just ignore this `inline` and iirc we are not using lto. An alternative approach could be to define it as a static function in this header file so the compiler can optimize in the object file

lib/pbkit/pbkit.c
JayFoxRox: So a user can indent all pbkit commands with `p = `; it also makes clear that the command stream pointer is dead after `pb_end()`; this prevents re-use of invalid pointers.

There's no real downside for having this feature as people can simply ignore the return value if they don't want it.

The only downside would be if we decide to use the return value for something else.

samples/mesh/main.c
JayFoxRox: The weird thing here is also the different formatting of the `pb_begin` and `pb_end` line.

See my other comment for rationale of `pb_end` return value.

lib/pbkit/pbkit.c
mborgerson: I don't really see a lot of value there, I think just having the `pb_end()` not indented is more obvious. But it's fine #bikeshed

lib/pbkit/pbkit.h
JayFoxRox: Described in PR description. Consider this preparation for #156 

Moving code into the header would be an ugly temporary hack and could cause issues with object files of different pbkit versions.

mborgerson: > Moving code into the header would be an ugly temporary hack

Not really? Just put them in a .h or .inl and `#include` it

> could cause issues with object files of different pbkit versions
Huh? This is not an issue. We aren't keeping old object files around, and Make resolves the include dependency

lib/pbkit/pbkit.c
dracc: Good catch. This change has been reverted. Thanks!

samples/mesh/main.c
dracc: `p = pb_*` formatting fixed throughout the file.

lib/pbkit/pbkit.c
```diff
@@ -2287,163 +2288,101 @@ void pb_end(DWORD *pEnd)
...
-    if (pb_BeginEndPair==0) debugPrint("pb_push2 : missing pb_begin earlier\n");
-    pb_PushIndex+=3;
-    pb_PushNext+=3;
-    if (pb_PushIndex>128) debugPrint("pb_push2: begin-end block musn't exceed 128 dwords\n");
-#endif
-
-    *(p+0)=EncodeMethod(SUBCH_3D,command,2);
-    *(p+1)=param1;
-    *(p+2)=param2;
+    pb_push_to(SUBCH_3D,p,command,nparam);
```
JayFoxRox: This function must return this.

JayFoxRox:
This doesn't work in XQEMU; after compiling the triangle sample and running I get the following error:

`qemu-system-i386: /home/fox/Data/Projects/xqemu/xqemu2/hw/xbox/nv2a/nv2a_pfifo.c:186: pfifo_run_puller: Assertion `engine == ENGINE_GRAPHICS' failed.`

The error appears to happen during `pb_init`. The sample still compiles / runs fine on nxdk master.

lib/pbkit/pbkit.c
```diff
@@ -1878,17 +1878,17 @@ void pb_target_back_buffer(void)
...
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_WAIT_MAKESPACE,0);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x08,dma_addr); //set params addr,data
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER); //calls subprogID PB_SETOUTER: does VIDEOREG(addr)=data
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x0C,dma_addr);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x00,dma_flags);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    p=pb_push2(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,NV_PRAMIN+(pb_DmaChID11Inst<<4)+0x04,dma_limit);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_SETOUTER);
+    pb_push1_to(SUBCH_4,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT2,11);
```
JayFoxRox: This does not modify `p` anymore

lib/pbkit/pbkit.c
```diff
@@ -3322,15 +3322,15 @@ int pb_init(void)
...
-    pb_push2_to(SUBCH_4,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT1,3,11); p+=3;
-    pb_end(p); //calls pb_start() which will trigger the reading and sending to GPU (asynchronous, no waiting)
+    pb_push1_to(SUBCH_2,p,NV20_TCL_PRIMITIVE_SET_MAIN_OBJECT,14);
+    pb_push1_to(SUBCH_3,p,NV20_TCL_PRIMITIVE_SET_MAIN_OBJECT,16);
+    pb_push1_to(SUBCH_4,p,NV20_TCL_PRIMITIVE_SET_MAIN_OBJECT,17);
+    pb_push1_to(SUBCH_3D,p,NV20_TCL_PRIMITIVE_SET_MAIN_OBJECT,13);
+    pb_push1_to(SUBCH_2,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT0,7);
+    pb_push1_to(SUBCH_3,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT5,17);
+    pb_push1_to(SUBCH_3,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT_UNKNOWN,3);
+    pb_push2_to(SUBCH_4,p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT1,3,11);
```
JayFoxRox: None of this modifies p anymore

lib/pbkit/pbkit.c
```diff
@@ -3345,37 +3345,35 @@ int pb_init(void)
     *(p++)=3;
     *(p++)=3;
     *(p++)=8;
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_SET_OBJECT8,1);
-    *(p++)=12;
-    pb_push(p++,NV20_TCL_PRIMITIVE_3D_ACTIVATE_COLORS,1);
-    *(p++)=0;
-    pb_end(p);
+    pb_push1(p,NV20_TCL_PRIMITIVE_3D_SET_OBJECT8,12);
+    pb_push1(p,NV20_TCL_PRIMITIVE_3D_ACTIVATE_COLORS,0);
```
JayFoxRox: This doesn't modify `p` anymore

lib/pbkit/pbkit.c
```diff
@@ -3406,37 +3345,35 @@ int pb_init(void)
...
+    p=pb_push4f(p,NV097_SET_EYE_POSITION,0.0f,0.0f,0.0f,1.0f);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_EDGE_FLAG,1);
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_TX_SHADER_PREVIOUS,0x00210000); //(PSTextureInput) What previous stage is used at each stage
+    p=pb_push1(p,NV097_SET_COMPRESS_ZBUFFER_EN,0); //
+    p=pb_push1(p,NV097_SET_SHADOW_ZSLOPE_THRESHOLD,0x7F800000);
+    p=pb_push1(p,NV097_SET_ZMIN_MAX_CONTROL,1); //CULL_NEAR_FAR_EN_TRUE | ZCLAMP_EN_CULL | CULL_IGNORE_W_FALSE
+    p=pb_end(p);
 
     p=pb_begin();
     pb_push_transposed_matrix(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_A(0),pb_IdentityMatrix); p+=17;
```
JayFoxRox: These can use the return value of this function now.

lib/pbkit/pbkit.c
```diff
@@ -2290,193 +2290,97 @@ void pb_end(DWORD *pEnd)
...
     *((float *)(p+2))=param2;
     *((float *)(p+3))=param3;
     *((float *)(p+4))=param4;
+    return p+5;
 }
 
-void pb_push(DWORD *p, DWORD command, DWORD nparam)
+inline DWORD *pb_push(DWORD *p, DWORD command, DWORD nparam)
 {
     pb_pushto(SUBCH_3D,p,command,nparam);
```
JayFoxRox: Return value still broken in commit "pbkit: pb_push* refactoring"

lib/pbkit/pbkit.c
JayFoxRox: This was fixed, but in the wrong commit. I created a new review comment below.

lib/pbkit/pbkit.c
dracc: All of these instances ought to have been fixed now.

lib/pbkit/pbkit.c
dracc: Fixed now. :+1:

lib/pbkit/pbkit.c
dracc: This change has been reverted. Thanks for your input, both of you.

lib/pbkit/pbkit.c
```diff
@@ -2260,6 +2260,7 @@ void pb_end(DWORD *pEnd)
 #endif
 
 #ifdef DBG
+    if (pEnd!=pb_PushNext) debugPrint("pb_end: input pointer invalid or not following previous write addresses\n");
```
mborgerson: We should stop and scream (`assert`). Using the `debugPrint` macro can lead to missing these important error messages. Won't block merge for this but it should be changed

lib/pbkit/pbkit.c
```diff
@@ -2290,160 +2291,97 @@ void pb_end(DWORD *pEnd)
...
-#endif
+    pb_push_to(subchannel,p,command,1);
+    *(p+1)=param1;
+    return p+2;
+}
 
-    *(p+0)=EncodeMethod(subchannel,command,2);
+inline DWORD *pb_push2_to(DWORD subchannel, DWORD *p, DWORD command, DWORD param1, DWORD param2)
+{
+    pb_push_to(subchannel,p,command,2);
```
mborgerson: Return value ignored

lib/pbkit/pbkit.c
```diff
@@ -2464,9 +2402,9 @@ void pb_push_transposed_matrix(DWORD *p, DWORD command, float *m)
     *((float *)p++)=m[_24];
     *((float *)p++)=m[_34];
     *((float *)p++)=m[_44];
-}
-
 
+    return p+17;
```
mborgerson: This doesn't look right...We incremented the pointer while populating, now we are returning (p+16)+17

lib/pbkit/pbkit.c
JayFoxRox: Agreed.

We couldn't do this before this PR because any use of `pb_push` would automatically trigger these.

I had also [commented on Discord yesterday](https://discordapp.com/channels/428359196719972353/428360618102226946/628372619330060303):

> I have tried #158 in `DBG` mode (using XQEMU) and triangle sample seems to work, but mesh sample seems to generate warnings - however, at least some of them also happen in master.
> In particular, during startup, both samples cause "pb_push_to: new write address invalid or not following previous write addresses".
>
> [Note: I think the following comments are about the mesh sample - not sure if triangle is also affected]
> - master also causes it during runtime (and then causes XQEMU to assert due CPU / GPU both writing at the same time)
> - #158 continues but also causes "pb_push_to: begin-end block musn't exceed 128 dwords"

Since then, the last issue should have been fixed in #183 (Issue #178).

I had forgotten about the other ones, but my Discord messages imply those messages still trigger.
We should probably look into this before a merge.
I believe this PR should have fixed those messages.

lib/pbkit/pbkit.c
JayFoxRox: This is intended: `pb_push_to` returns `p+3` - it includes the gap for the data (which is written by this function).

So `pb_push_to` writes `*(p+0)` and this function writes the rest using the old pointer; on return, this function does the same increment that `pb_push_to` did (which was discarded).

I agree that this code is confusing.

I guess an alternative is:

```c
uint32_t* p_new = pb_push_to(subchannel,p,command,2);
*(p+1) = param1;
*(p+2) = param2;
return p_new;
```

*(I don't think that helps very much though)*

lib/pbkit/pbkit.c
dracc: Good eye, thanks!

lib/pbkit/pbkit.c
dracc: Fixed by adding assertions in dcb7a59 and bae0bc8 respectively. :+1:

JayFoxRox: This will definitely need testing with all nxdk samples now - I assume they'll hit the assertion in `DBG` mode now (unless the bugs were caused by the `p+17` matrix return value).

lib/pbkit/pbkit.c
```diff
@@ -2260,7 +2261,16 @@ void pb_end(DWORD *pEnd)
 #endif
 
 #ifdef DBG
-    if (pb_BeginEndPair==0) debugPrint("pb_end without a pb_start\n");
+    if (pEnd!=pb_PushNext)
+    {
+        debugPrint("pb_end: input pointer invalid or not following previous write addresses\n");
+        assert(false);
```
JayFoxRox: Compilation issue: `error: use of undeclared identifier 'false';`

I can't stress this enough:
- At the very least compile each build before pushing
- Also test your builds regularly

These repeated compilation issues cause a lot of delays in review.

lib/pbkit/pbkit.c
```diff
@@ -32,6 +32,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
+#include <assert.h>
```
JayFoxRox: nit: Probably only needed `#ifdef DBG`

lib/pbkit/pbkit.c
dracc: I just rebuilt the `mesh`, `sdl` and `triangle` samples using `make clean && make DBG=y DEBUG=y` and tested them in XQEMU, works the same as without the flags.

lib/pbkit/pbkit.c
JayFoxRox: `<stdbool.h>` seems to have been added now, but it's included even if `DBG` ist not defined (when it wouldn't be necessary). If we respect it for `assert.h`, we should also respect it for `stdbool.h`. It should just be consistent - I'm also fine with ignoring `DBG` for both of these paticular includes (having them regardless doesn't have any disadvantages).

dracc: I opted for putting them both inside the `#ifdef` guard.

lib/pbkit/pbkit.c
dracc: Done.

samples/mesh/main.c
```diff
@@ -297,20 +297,17 @@ static void init_shader(void)
...
+                 | MASK(NV097_SET_TRANSFORM_EXECUTION_MODE_RANGE_MODE, NV097_SET_TRANSFORM_EXECUTION_MODE_RANGE_MODE_PRIV));
 
-    pb_push1(p, NV097_SET_TRANSFORM_PROGRAM_CXT_WRITE_EN, 0);
-    p += 2;
+    p = pb_push1(p, NV097_SET_TRANSFORM_PROGRAM_CXT_WRITE_EN, 0);
 
     /* Set cursor and begin copying program */
-    pb_push1(p, NV097_SET_TRANSFORM_PROGRAM_LOAD, 0);
-    p += 2;
+    p = pb_push1(p, NV097_SET_TRANSFORM_PROGRAM_LOAD, 0);
```
JayFoxRox: This change conflicts with #187 which will have to be merged first (as this PR would hit `assert` otherwise).

The exact same code is present in the triangle sample. I say, fix it here, so mesh sample itself is consistently using new API; and we can fix it in triangle in a follow-up PR.

JayFoxRox: triangle sample is also updated by this PR now, so this comment isn't relevant anymore.

dracc:
I've rebased upon f66a9410 and updated the `triangle` sample in a new commit.

lib/pbkit/pbkit.c
JayFoxRox: Those issues have been fixed by now (#187 and #183, possibly more).

JayFoxRox:
I plan to do a final review (and potentially merge this) by the end of this week.

@mborgerson Your last review was "Changes requested", but I believe most of the feedback has been integrated. The state of this PR is probably "better-than-master" and only minor complaints remain (disagreement about `inline` for example).
So if you still want to veto this, please leave a comment / review.

lib/pbkit/pbkit.c
```diff
@@ -34,6 +36,7 @@
...
```
JayFoxRox: This is included twice, once for `DBG` and once unconditionally

lib/pbkit/pbkit.c
```diff
@@ -2267,7 +2267,16 @@ void pb_end(DWORD *pEnd)
...
 #ifdef DBG
-    if (pb_BeginEndPair==0) debugPrint("pb_end without a pb_start\n");
+    if (pEnd!=pb_PushNext)
+    {
+        debugPrint("pb_end: input pointer invalid or not following previous write addresses\n");
+        assert(false);
+    }
+    if (pb_BeginEndPair==0)
+    {
+        debugPrint("pb_end without a pb_start\n");
```
JayFoxRox: nit: This should be `pb_begin` instead of `pb_start`.

It is already a bug in the existing code.
A similar bug exists in `pb_begin` which calls itself `pb_start`.

lib/pbkit/pbkit.c
```diff
@@ -2297,7 +2297,7 @@ void pb_end(DWORD *pEnd)
 }
 
 
-void pb_pushto(DWORD subchannel, DWORD *p, DWORD command, DWORD nparam)
+inline DWORD *pb_pushto(DWORD subchannel, DWORD *p, DWORD command, DWORD nparam)
```
JayFoxRox: Considering https://github.com/XboxDev/nxdk/pull/158/commits/2bcb78718f251614cb68873ec1e33debd46efe1d#r330211066, I'd suggest to make this return `void` for now?

At least it can't be misused then.

lib/pbkit/pbkit.c
```diff
@@ -3387,33 +3387,31 @@ int pb_init(void)
...
-*/  pb_push4f(p,NV097_SET_FOG_PLANE,0.0f,0.0f,1.0f,0.0f); p+=5;
-    pb_push1(p,NV20_TCL_PRIMITIVE_3D_VP_UPLOAD_CONST_ID,0x0000003C); p+=2; //set shader constants cursor at C-36
+    p=pb_push_transposed_matrix(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_A(0),pb_IdentityMatrix);
+    p=pb_push_transposed_matrix(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_A(4),pb_IdentityMatrix);
+    p=pb_push_transposed_matrix(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_A(8),pb_IdentityMatrix);
+    p=pb_push_transposed_matrix(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_A(12),pb_IdentityMatrix);
+/*  p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_ENABLE(0),0x2202); 
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_ENABLE(1),0x2202); 
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_ENABLE(2),0x2202); 
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_CLIP_PLANE_ENABLE(3),0x2202);
```
JayFoxRox: These 4 lines have a space before the linebreak.

lib/pbkit/pbkit.c
dracc: Conditional include removed.

lib/pbkit/pbkit.c
dracc: A separate commit was filed, listed last.

lib/pbkit/pbkit.c
dracc: Fixed. File seems to be riddled by lines ending with whitespace btw.

lib/pbkit/pbkit.c
dracc: Fixed - they both return `void` now.

dracc:
The `inline` manual optimization has now been removed as requested by @mborgerson and later by @JayFoxRox.

lib/pbkit/pbkit.c
dracc: Taken care of now - `pb_push` and `pb_push_to` now return `void`.

lib/pbkit/pbkit.c
```diff
@@ -20,8 +20,9 @@
 #include <hal/io.h>
 #include <xboxkrnl/xboxkrnl.h>
 #include <hal/debug.h>
+#ifdef DBG
 #include <stdbool.h>
-#include <assert.h>
+#endif
```
JayFoxRox: `stdbool` is also used unconditionally in functions like `pb_set_color_format` and `pb_init`

lib/pbkit/pbkit.c
JayFoxRox: Keep the `assert` include above - it already existed in master

dracc: Changes to includes reverted. Fixed.

lib/pbkit/pbkit.c
dracc: Changes to includes reverted. Fixed.

mborgerson:
Thanks!

Issue #130, opened 2019-06-10
Update the lwip submodule


JayFoxRox:
Our lwip is from early 2016. There have been some changes which also requires updating the arch specific code.

There's probably various security, stability and performance upgrades we'd benefit from.

JayFoxRox:
Shallow clones of the lwip submodule are problematic. This also affects GitHub Actions CI (which, at the time of writing, is proposed to become our CI). This is not working:

```
- uses: actions/checkout@v1
  with:
    fetch-depth: 1
    submodules: recursive
```

The issue can be recreated locally (although it might not be the exact action/checkout behaviour):

```
$ git init
Initialized empty Git repository in /tmp/lwip/.git/
$ git remote add o https://git.savannah.gnu.org/git/lwip.git
$ git fetch o b9d0d8094638e042f92e15c24015a172cacc65c4
error: Server does not allow request for unadvertised object b9d0d8094638e042f92e15c24015a172cacc65c4
```

The message shows that git.savannah.gnu does not allow fetching of any unadvertised object, so we'd have to use [an existing ref](http://git.savannah.nongnu.org/cgit/lwip.git/refs) instead.
However, we currently use some random commit from early 2016.

Fetching an advertised object (ref) works:
```
$ git fetch o STABLE-2_1_2_RELEASE --depth=1
remote: Counting objects: 438, done.
remote: Compressing objects: 100% (390/390), done.
remote: Total 438 (delta 66), reused 186 (delta 27)
Receiving objects: 100% (438/438), 1.29 MiB | 1.76 MiB/s, done.
Resolving deltas: 100% (66/66), done.
From https://git.savannah.gnu.org/git/lwip
 * tag               STABLE-2_1_2_RELEASE -> FETCH_HEAD
```

The closest tag to our current revision would be http://git.savannah.nongnu.org/cgit/lwip.git/tag/?h=STABLE-2_0_0_RC1
We should at least switch to that revision, to improve the ability to clone.

I also hope that someone knows a better solution, just in case we still want to diverge from lwip tags in the future.
For example, there's also https://git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtshallow to force shallow clones of a submodule (`true`); unfortunately it can not be used to force a deep clone (`false` or similar are unspecified) which could work around this issue.

mborgerson:
Closed by #244

PR #168, opened 2019-09-07
ci: GitHub Actions for CI tasks

### LINKED ISSUE #36: Setup CI build system

GXTX:
It is expected that on November 13th GitHub is releasing to the public a feature called 'Actions'. We can use Actions as a CI/CD and hence this PR. It currently is in beta so I'm leaving this PR as a draft until it's public and fleshed out.

There are a couple of caveats I've noticed as of now:

- Actions require a single workflow to be present in `master` before it'll run on different branches
- Removing/Renaming a workflow will leave previously ran instances in the Actions list
- You are not allowed to remove or delete previously ran workflows, including ones with artifacts
- Not related to NXDK but XQEMU: `upload-artifact` has no way to link to a 'latest build' and requires users to navigate

Even with these drawbacks I believe this solution will be better than relying on 3rd party services such as AppVeyor and Travis-CI both of which have their own unique drawbacks, such as with Travis-CI having outdated OS builds, and a Windows builder who's been in beta for over a year with no updates. So the end goal would be to remove these from the repo and rely on the integrated solution.

Suggestions welcomed.

Closes #36.


.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,51 @@
...
+      - uses: actions/checkout@master
+        with:
+          submodules: recursive
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update -y && sudo apt-get install -y flex bison clang lld
+      - name: Build
+        run: sh ./.ci_build_samples.sh
+  macOS:
+    name: MacOS
```
JayFoxRox: Why is this MacOS instead of macOS?

GXTX: Stylized, is Apple's official way lower case even at the beginning on a sentence?

GXTX: I've pushed the fix for this now.

.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,51 @@
...
+  macOS:
+    name: macOS
+    runs-on: macOS-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          submodules: recursive
+      - name: Install Dependencies
+        run: |
+          brew update && brew install llvm
```
thrimbor: I don't know how much of a problem this is or will be, but on Travis, which uses quite old images, brew cleanup gets triggered every time. I considered PRing a change to fix this and shave a bit off of our build time, but I expect this to replace Travis.

If you want to add the change to be on the safe side when it comes to minimizing build times, this should do the trick:
`export HOMEBREW_NO_INSTALL_CLEANUP=1`

.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,51 @@
+name: Build Samples
+
+on: 
+  push:
+    branches: master
```
thrimbor: Recommend to change this to `'*'`, otherwise forked repos will only build master as well.

.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          submodules: recursive
+      - name: Install MSYS2 & Dependencies
+        run: |
+          choco install msys2
+          C:\tools\msys64\usr\bin\bash.exe -lc "pacman --needed --noconfirm -S make bison flex mingw-w64-x86_64-llvm"
+          C:\tools\msys64\usr\bin\bash.exe -lc "pacman --needed --noconfirm -S mingw-w64-x86_64-clang mingw-w64-x86_64-lld"
```
thrimbor: Why two separate invocations? Also this will have to install CMake now, too.

.github/workflows/verify_samples.yml
GXTX: The reason I set it like this is to avoid having it run twice if someone makes a self-PR from within the Project. I can change it still if you wish.

.github/workflows/verify_samples.yml
GXTX: I have it be 2 different calls to avoid a needlessly long line (although it's already over 80col so not sure why I care), and yes I just noticed we updated the exiso submodule.

If you want, again, I can change this.

JayFoxRox: Change it.

Two invocations might affect behavior (dependencies for example) and should be avoided on CI, unless there's actual reasons as to why it's necessary.

There's probably also a way to split lines in the string (typically a backslash) if it's really necessary.

.github/workflows/verify_samples.yml
JayFoxRox: I currently don't have an opinion on this.

> if someone makes a self-PR from within the Project.

I strongly doubt we'd do that.
It also has the negative side-effect of forks having the temporary branch, which is something we avoid.

---

I have yet to try GitHub Actions.

- Does GitHub Actions allow deleting of builds? (if not, we might want to keep the "master" variant to avoid accidental pushes)
- Does it allow manual configuration overrides outside of this yml? (if so, we could fix this upstream)

GXTX: > Does GitHub Actions allow deleting of builds?

From what I can tell, no.

.github/workflows/verify_samples.yml
GXTX: Fixed.

.github/workflows/verify_samples.yml
GXTX: Fixed.

.github/workflows/verify_samples.yml
GXTX: > Does it allow manual configuration overrides outside of this yml?

If you're talking about a single run where the configuration is changed, no.

JayFoxRox: I'm not sure if we are talking about the same thing. So I'll explain what I was thinking:

I believe Travis and AppVeyor allow to specify a configuration in their Web-UI. This allows projects to ignore hostile changes in PRs (which might affect the visibility of the build) or to allow storing a CI configuration for forks, while using a different configuration upstream.

So if GitHub Actions allows overriding configuration fields in a web UI, then we can keep our CI config in the git repository configured for users of forks (build `*`); our own repository would be using a Web-UI override to only  build `master` (although that is pointless if that override would also be included in forks).

JayFoxRox:
nit: "Ci" in commit title triggers me. - I had already mentioned this on Discord a while ago. Ideally the maintainer who eventually merges this will change it manually (or the author will).

---

Also inviting other @XboxDev maintainers to review. I don't want to deal with CI related stuff personally (I have never used this CI system in particular, and it's not my strong suit either). This PR has been open for a while and author has been pushing for more action/review on Discord.

Keep in mind that we should disable Travis-CI before merging this, so we don't error Travis on the merge anymore (as this also removes the configuration).

.github/workflows/verify_samples.yml
GXTX: Ah then no, it seems it does not allow such things.

.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,50 @@
...
+  windows:
+    name: Windows
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
+          submodules: recursive
+      - name: Install MSYS2 & Dependencies
+        run: |
+          choco install msys2
```
thrimbor: There is some issue with chocolatey and the way GitHub handles its output, causing thousands of progress messages getting spammed into the log. It can be fixed by replacing this line with `choco install msys2 --no-progress` - I tested this change myself and can confirm that it's working and has no unwanted side effects.

JayFoxRox: Where can I see a log for either variant?

*(I feel like a link to CI logs should have been added to the first post)*

thrimbor: The actual branches I used for testing are gone, but the logs are here:
[Without the fix](https://github.com/thrimbor/nxdk/runs/327551909)
[With the fix](https://github.com/thrimbor/nxdk/runs/327572236)

It affects the `Install MSYS2 & Dependencies` part.

.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,50 @@
+name: Build Samples
+
+on: [push, pull_request]
+
+jobs:
+  linux:
+    name: Ubuntu
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
JayFoxRox: "Run actions/checkout@master" in log is quite ugly.

XQEMU CI seems to handle this step better in general:

```
- name: Clone Tree
  uses: actions/checkout@v1
  with:
    fetch-depth: 1
```

Note how this also restricts fetch-depth, which improves CI performance.
It also picks a stable version of the checkout action.

.github/workflows/verify_samples.yml
JayFoxRox: Yes. Fix should be added. I don't think it hides anything we have to see.

.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,50 @@
+name: Build Samples
+
+on: [push, pull_request]
+
+jobs:
```
JayFoxRox: nit: I prefer jobs / OS lists in this order:

- Windows
- macOS
- Linux (Ubuntu in this case)

For user-facing stuff, Windows and macOS are more popular than Linux. For CI stuff, we should still keep it for consistency, and it also reflects what's most important for most developers (as the order of OS popularity is the other way around).

Also Windows and macOS are more likely to have artifacts, so keeping those OS on top would be nice.

JayFoxRox:
While we are at it, we could add CI badges (in a separate commit in this PR, or a follow-up PR - depending on @GXTX frustration level).
I had been meaning to close #36 as there wasn't any veto - but CI badges are the last missing bit (to me).

.github/workflows/verify_samples.yml
JayFoxRox: Note: XQEMU actually split platforms into different workflows (my preferred order is lost then); different workflows is useful for having different CI badges (not sure if this is possible otherwise).
I'm not sure, but tasks might also run in parallel then (not sure how jobs handle this)?

.github/workflows/verify_samples.yml
GXTX: Both logs appear the same beyond a misplaced `\n` causing double spacing. There is an argument about `stable` versions so I will change back to v1.

.github/workflows/verify_samples.yml
GXTX: Even if unintentional it has been the preferred order on XQEMU to run Linux (Ubuntu) then macOS on travis-ci, i'd like to keep with that and add Windows at the bottom.

.github/workflows/verify_samples.yml
GXTX: Good find! It was rather annoying having my browser window crash when watching the the progress.

.github/workflows/verify_samples.yml
JayFoxRox: > Both logs appear the same beyond a misplaced \n causing double spacing.

Where? I'm not sure I follow.

You seem to imply that the log doesn't respect the `--depth` either (which should probably differ in log - although I did not verify it myself)?

---

I meant the following differences:

1. The step has a different name in XQEMU ("Clone Tree") vs. nxdk CI ("Run actions/checkout@master").
2. Should be more stable in the future (by naming an explicit version)
3. Download less data for XQEMU (only the latest revision, other than downloading the entire history)

GXTX: Misunderstood. I don't see a point in adding a name, I think it's rather clear what the external action is doing.

.github/workflows/verify_samples.yml
JayFoxRox: As I have said in my previous comment: XQEMU splits platforms into workflows with the Actions CI. The workflows seems to be alphabetical order (or round-robin or something?) - I don't think we order it explicitly anymore. So there's no "preferred order". The implicit order is fine - there's no point in adding more code just to have the explicit order.

However, generally, if we explicitly list platforms I ensure we do this in this order: Windows, macOS, Linux (which would be the preferred order).

*(We did have some older lists in XboxDev and XQEMU which might still be unordered, but most user facing changes had been changed in the past, and new lists generally follow this order as I bring it up in review)*

---

Splitting into different workflows for this nxdk CI might also be beneficial for different CI badges (also explained in my previous comment) - it will be implicit order then; but we might have other benefits (different badges per platform + possibly faster jobs?).

.github/workflows/verify_samples.yml
JayFoxRox: Helps with consistency and hides implementation details (avoids confusing nxdk@v1 with action@v1 for example).
The current name sticks out like a sore thumb.

.github/workflows/verify_samples.yml
GXTX: I disagree in this ordering.

.github/workflows/verify_samples.yml
GXTX: In my opinion it should, it's an important step and I would rather know when a step is using a external workflow instead of internal.

.github/workflows/verify_samples.yml
```diff
@@ -0,0 +1,52 @@
+name: Build Samples
```
JayFoxRox: nit: Inconsistent with filename (verify vs. build)

JayFoxRox:
I had recommended a shallow clone during review. However, this causes issues with lwip: https://github.com/XboxDev/nxdk/issues/130#issuecomment-560164537
Rushing a change for the lwip version might cause issues.
I'd suggest that we do deep clones for now (removing the `fetch-depth` from this PR).
Once the revision is changed or a workaround is found, we can re-introduce shallow clones.

.github/workflows/verify_samples.yml
JayFoxRox: Regarding the 3 benefits:
1. Name change: This was not done, but I feel strongly about this.
    We disagree, so some other maintainer should take a look (@XboxDev).
    I think that information-hiding is important, especially for a user-facing portions (the name is shown in web-ui).
    Also we don't explicitly mark other external resources in the step names either. ("chocolatey" for example).
2. Stable revision of workflow: This has been changed. Thanks.
3. Shallow clone: This has been tried, but fails. I'd recommend to drop the `depth-fetch`; explained in https://github.com/XboxDev/nxdk/pull/168#issuecomment-560164656.

.github/workflows/verify_samples.yml
JayFoxRox: This has been changed. @GXTX please always leave a comment on discussions *after* pushing a fix (and mention what you changed). This makes it easier for maintainers to know which discussions can be resolved.

.github/workflows/verify_samples.yml
JayFoxRox: The ordering was changed.
This discussion remains unresolved because splitting workflows per platform might have benefits (but wasn't done yet):

- Probably easier to add more steps for things other than building samples.
- [GitHub badges are per workflow](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository) (ability to separate platforms). At the time of writing, there don't seem to be per-job badges.
- Separate the configuration more cleanly into separate files (which might avoid git merge conflicts).

That said: I could tolerate the current PR (in fact, I prefer how the GitHub Actions CI displays this workflow over the one proposed for XQEMU which is per-platform workflows).

Also, I'm not confident in either solution; I simply lack the understanding of how GitHub intends users to use GitHub Actions. I don't fully understand the differences between workflows, jobs and the actions themselves.

However, I worry we'd change this to per-platform-workflow soon. My understanding is, that @mborgerson had also expressed that per-platform workflows have benefits (or did I misunderstand this?).
Also XQEMU CI is probably going to do per-platform-workflows, so at least some maintainers will have slightly more expertise with that soon.

So it might be of interest to change this in this PR to per-platform-workflows (either by creating a duplicate of this PR for per-platform, *or* by sending a PR to @GXTX - if he'd be willing to accept such changes).

.github/workflows/verify_samples.yml
GXTX: Depth has been fixed.

.github/workflows/verify_samples.yml
GXTX: Fixed.

.github/workflows/verify_samples.yml
GXTX: I disagree enough about splitting per-OS and am not going to change this and have created an issue to track this at #243.

.github/workflows/build_samples.yml
```diff
@@ -0,0 +1,49 @@
+name: Build Samples
+
+on: [push, pull_request]
+
+jobs:
+  windows:
+    name: Windows
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v1
```
JayFoxRox: I consider name is still "bad"; see https://github.com/XboxDev/nxdk/pull/168#discussion_r352331736.

*(Extra discussion as old discussion has diverged; with 2/3 items resolved)*

GXTX:
Unfortunately due to some things I'll be unable to continue working on this PR and as such I'm going to consider it abandoned and close it out, it seems for the most part it is finished so maybe someone else can continue working where i've left it off.

PR #185, opened 2019-10-02
vcruntime: Add pure virtual function call handling

thrimbor:
This is a very MS and C++ specific piece of code - unfortunately, it's impossible to test this meaningfully without some large parts of my libc++ work (we would need to try and call a pure virtual function, but such code will emit RTTI references we don't provide yet).

As the name suggests, clang will emit code calling `_purecall` whenever the program tries to call a pure virtual function. The program is able to set it's own handler (which should never return), but there also has to be default handling (we're just printing info here and enter an endless loop).

As for the new header: While being considered part of vcruntime, `_get_purecall_handler` and `_set_purecall_handler` are supposed to be provided by `stdlib.h`. To not clutter PDCLib with non-standard stuff, I have spoken with upstream to get extension hooks (which we'll get with the merge of https://github.com/XboxDev/nxdk-pdclib/pull/12) which we can use to include additional stuff into the standard headers.
When the PDCLib update (and this) is merged, we can simply add the option to the PDCLib config to activate this.

/edit:
Using the extension hooks for this can be tested by adding `#define _PDCLIB_EXTEND_STDLIB_H <stdlib_ext_.h>` to `platform/xbox/include/pdclib/_PDCLIB_config.h` in the PDCLib directory after checking out my branch from https://github.com/XboxDev/nxdk-pdclib/pull/12.

### DIFF
--- a/lib/xboxrt/libc_extensions/stdlib_ext_.h
+++ b/lib/xboxrt/libc_extensions/stdlib_ext_.h
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2019 Stefan Schmidt
+ *
+ * Licensed under the MIT License
+ */
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -5,6 +11,11 @@ extern "C" {
 #ifdef _CRT_RAND_S
 int rand_s (unsigned int *randomValue);
 #endif
+
+typedef void (__cdecl *_purecall_handler)(void);
+
+_purecall_handler __cdecl _get_purecall_handler (void);
+_purecall_handler __cdecl _set_purecall_handler (_purecall_handler function);
 
 #ifdef __cplusplus
 }
--- a/lib/xboxrt/vcruntime/purecall.c
+++ b/lib/xboxrt/vcruntime/purecall.c
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2019 Stefan Schmidt
+ *
+ * Licensed under the MIT License
+ */
+
+#include <stdbool.h>
+#include <stdlib_ext_.h>
+#include <windows.h>
+#include <hal/debug.h>
+
+_purecall_handler current_purecall_handler = NULL;
+
+int __cdecl _purecall ()
+{
+    if (current_purecall_handler) {
+        current_purecall_handler();
+    }
+
+    debugPrint("Pure virtual function called!\n");
+
+    // Make sure the handler doesn't return
+    while(true) {
+        Sleep(1000);
+    }
+
+    // return value is not documented by MS (unreachable anyway)
+    return 0;
+}
+
+_purecall_handler __cdecl _get_purecall_handler (void)
+{
+    return current_purecall_handler;
+}
+
+_purecall_handler __cdecl _set_purecall_handler (_purecall_handler function)
+{
+    _purecall_handler old_handler = current_purecall_handler;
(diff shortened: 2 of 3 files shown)


lib/xboxrt/vcruntime/purecall.c
```diff
@@ -0,0 +1,37 @@
...
+_purecall_handler current_purecall_handler = NULL;
+
+int __cdecl _purecall ()
+{
+    if (current_purecall_handler) {
+        current_purecall_handler();
+    }
+
+    DbgPrint("Pure virtual function called!\n");
+    debugPrint("Pure virtual function called!\n");
```
JayFoxRox: I'd rather avoid these double-prints for different interfaces.

I'm worried that we'll search/replace all `debugPrint` in the future, and then we end-up with double-prints.
It also adds risk of only modifying one instance, so you'd also want to use `const char* message = "Pure virtual function called!\n";`.

If need be, we should just replicate `debugPrint` output to `DbgPrint` (might not work from interrupt handlers?) by changing debug.c.

lib/xboxrt/vcruntime/purecall.c
```diff
@@ -0,0 +1,37 @@
...
+
+    // Make sure the handler doesn't return
+    while(true) {
+        Sleep(1000);
+    }
+
+    // return value is not documented my MS
+    return 0;
```
JayFoxRox: Should document that this is currently unreachable, too?

thrimbor: Done.

lib/xboxrt/vcruntime/purecall.c
thrimbor: I removed the `DbgPrint`. It's mostly a leftover from trying my libc++ branch in xqemu with a debug bios.

JayFoxRox:
Will wait for #192 so reviewers can test this.

lib/xboxrt/vcruntime/purecall.c
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    debugPrint("Pure virtual function called!\n");
+
+    // Make sure the handler doesn't return
+    while(true) {
+        Sleep(1000);
+    }
+
+    // return value is not documented my MS (unreachable anyway)
```
JayFoxRox: `my` &rarr; `by`

lib/xboxrt/vcruntime/purecall.c
```diff
@@ -0,0 +1,36 @@
+#include <assert.h>
```
JayFoxRox: This isn't used here at the moment.

lib/xboxrt/vcruntime/purecall.c
thrimbor: Fixed.

lib/xboxrt/vcruntime/purecall.c
thrimbor: Removed.

lib/xboxrt/libc_extensions/stdlib_ext_.h
```diff
@@ -0,0 +1,12 @@
+#ifdef __cplusplus
```
mborgerson: How would you like to license this code?

thrimbor: MIT license note added.

PR #186, opened 2019-10-02
libc: Add strdup / _strdup as extensions

thrimbor:
Like #185, this is supposed to use the new PDCLib extension hooks once they're merged, see that PR for details.

This adds `strdup` and the MS-specific alias `_strdup`. While not part of the C standard (yet, C2x will probably add it), it's required by libc++.

/edit:
Using the extension hooks for this can be tested by adding `#define _PDCLIB_EXTEND_STRING_H <string_ext_.h>` to `platform/xbox/include/pdclib/_PDCLIB_config.h` in the PDCLib directory after checking out my branch from https://github.com/XboxDev/nxdk-pdclib/pull/12.

### DIFF
--- a/lib/xboxrt/libc_extensions/string_ext_.c
+++ b/lib/xboxrt/libc_extensions/string_ext_.c
@@ -1,0 +1,16 @@
+#include <string.h>
+#include <stdlib.h>
+
+char *strdup (const char *s)
+{
+    if (s == NULL) {
+        return NULL;
+    }
+
+    char *new_s = malloc(strlen(s) + 1);
+    if (new_s != NULL) {
+        strcpy(new_s, s);
+    }
+
+    return new_s;
+}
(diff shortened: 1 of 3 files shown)


JayFoxRox:
Will wait for #192 so reviewers can test this.

lib/xboxrt/libc_extensions/string_ext_.c
```diff
@@ -0,0 +1,16 @@
+#include <string.h>
+#include <stdlib.h>
+
+char *strdup (const char *s)
```
mborgerson: Why is this not just in functions/string/strdup.c instead?

thrimbor: Imho it makes sense to keep it separate to make it clear that this is a downstream extension (upstream only want C functionality, not POSIX), and to make PDCLib easier to rebase by keeping the changes to non-xbox code small.

JayFoxRox: `strdup` was only added in C2x is very recent and I assume pdclib will take some time to support it (if it's even planned).

So I think this is fine to keep it in our tree for now (as GNU/POSIX/... extension to earlier C standards, which are common versions we compile with).

JayFoxRox:
Tested and worked fine. Thanks!

PR #187, opened 2019-10-04
samples: Fix VP upload

### LINKED ISSUE #184: Mesh sample: "begin-end block musn't exceed 128 dwords" in pbkit
The mesh sample submits large `pb_begin/end` blocks in `init_shader`.

Before #158 is merged, this error is hidden because the `pb_push` macro doesn't do these checks (instead, `pb_push` currently causes other errors).
After #158 is merged, this error will trigger.

The included shader is 44 instructions long. Each instruction is 4 DWORDs long.
This would make a total of 44 \* 4 = 176 DWORDs (excluding further overhead for GPU commands).
This is already too much.

However, the mesh sample errornously assumes that each instruction is only 8 bytes long; it uses this to calculate the instruction count from the VP binary size (176 DWORDs * 4 byte = 704 bytes).
This means a total of 704 / 8 = 88 instructions are uploaded instead.
This makes a total of 88 \* 4 = 352 DWORDs (excluding further overhead for GPU commands).

To fix this:
- The instruction count must be be calculated correctly: `sizeof(vs_program)/8` &rarr; `sizeof(vs_program)/16`.
- Individual `pb_begin/end` blocks must be used within the instruction upload loop.

The same issue is also present in the triangle sample but does not exceed the limit due to the shorter VP. The overflow is still bad and should be fixed.

JayFoxRox:
Closes #184 

Fixes VP upload errors in the samples:

**Uploads the correct number of instructions now.**
This used to be miscalculated which caused overflows.

**Uploads each instruction in its own `pb_begin` / `pb_end` block.**
This avoids long pbkit blocks; pbkit warnings suggest that 128 DWORDs is the maximum number of DWORD per block. I believe this is the GPU cache size, although I don't think the hardware would struggle when submitting more. For now, we'll just respect this pbkit limit (it will be enforced using `assert` by #158), but we should eventually review why it exists.

---

I've only tested this in XQEMU

### DIFF
--- a/samples/mesh/main.c
+++ b/samples/mesh/main.c
@@ -308,17 +308,22 @@ static void init_shader(void)
     pb_push1(p, NV097_SET_TRANSFORM_PROGRAM_CXT_WRITE_EN, 0);
     p += 2;
 
-    /* Set cursor and begin copying program */
+    pb_end(p);
+
+    /* Set cursor for program upload */
+    p = pb_begin();
     pb_push1(p, NV097_SET_TRANSFORM_PROGRAM_LOAD, 0);
     p += 2;
+    pb_end(p);
 
-    for (i=0; i<sizeof(vs_program)/8; i++) {
+    /* Copy program instructions (16-bytes each) */
+    for (i=0; i<sizeof(vs_program)/16; i++) {
+        p = pb_begin();
         pb_push(p++, NV097_SET_TRANSFORM_PROGRAM, 4);
         memcpy(p, &vs_program[i*4], 4*4);
         p+=4;
+        pb_end(p);
     }
-
-    pb_end(p);
 
     /* Setup fragment shader */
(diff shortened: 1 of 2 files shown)


dracc:
I end up with this when cherrypicking this on top of 8f7db3b0c100738491779facefd691b02c6d10f3 from #158 and running `mesh`:
mesh

`triangle` still works though:
triangle

JayFoxRox:
I was unable to reproduce that issue. I have rebased it myself in https://github.com/JayFoxRox/nxdk/pull/25 and it worked with the last commit from dracc PR modified (so this fix remains) or that last commit removed. Maybe a different base?

LukeUsher:
with both @dracc approving and @JayFoxRox stating that 'this PR is fine', we can merge this now.

I don't see any reason to delay this further.

PR #188, opened 2019-10-04
makefile: Cleanup

JayFoxRox:
The idea is so that if someone includes our makefile, they can use `NXDK_*FLAGS` if they want to call the tools themselves.

Additionally, the nxdk makefile can then be interacted with using standard `*FLAGS` variables, without them being modified by nxdk (in case the makefile that included nxdk makefile wants to use them again).
This is still very dirty, but a bit more consistent and predictable.

I've also added the `-MP` flag to avoid errors which would appear after building with a feature-branch (which had headers which are missing when switching back to master): 
```
$ make
make: *** No rule to make target 'nxdk/samples/triangle/../../lib/xgu/xgu.h', needed by 'nxdk/samples/triangle/main.obj'.  Stop.
$
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,9 +70,6 @@ NXDK_CXXFLAGS += -g -gdwarf-4
 NXDK_LDFLAGS += -debug
 endif
 
-NXDK_CFLAGS += $(CFLAGS)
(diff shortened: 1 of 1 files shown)


mborgerson:
Thanks!

PR #192, opened 2019-10-06
pdclib: Update submodule to 96a8f82

### LINKED ISSUE #174: warning: 'WAIT_FAILED' macro redefined
This warning happens quite often.

It's currently in https://github.com/XboxDev/nxdk/blob/a75692f24532bf654addab88c45de69cb8e183ce/lib/winapi/synchapi.h#L23 but also in https://github.com/XboxDev/nxdk/blob/master/lib/xboxkrnl/xboxkrnl.h#L58

The warning can be ignored, but we should still fix it.

thrimbor:
Integrates changes listed in https://github.com/XboxDev/nxdk-pdclib/pull/12.

/edit:
Now also includes removal of duplicated `WAIT_FAILED` definition (as this is now possible with the updated PDCLib). Closes #174.


dracc:
NevolutionX builds fine and runs as good as usual with this. Is there anything else I should pay attention to when testing?

PR #193, opened 2019-10-06
video: Add 480i component video modes

### LINKED ISSUE #66: No support for AV_PACK_HDTV in combination with VIDEO_REGION_PAL
While XDK code works fine with the component cable when the Xbox is configured to PAL, the screen remains black (or Xbox even crashes?) with the same combination for nxdk.

The code implementing this is here: https://github.com/XboxDev/nxdk/blob/d1c2ff3513c9365d1cf90b21cf74689b35a56a8b/lib/hal/video.c#L58

The code does list modes for `VIDEO_REGION_NTSCM` and `VIDEO_REGION_NTSCJ` in combination with `AV_PACK_HDTV`. However, a mode for `VIDEO_REGION_PAL` is missing.
We should also review other adapters and add fallback modes if possible.

dracc:
Needed for #142
Closes #66 

Adds 480i video mode settings for the HDTV AV pack.

Source for the additions is https://sourceforge.net/projects/openxdk/files/Patches%20and%20Drivers/Zagon_s%20480i%20patch/zagons-480i-patch.tar.gz/download


LukeUsher:
I'm unable to test this myself hardware because I don't have a HDTV av pack, but the mode values do match what we have in Cxbx-R, so I assume they're correct.

I'm happy to merge this end of today, unless there are any objections.

Issue #197, opened 2019-10-07
fp20compiler: Unable to define different colors for general-combiners and final-combiner


JayFoxRox:
GL doesn't allow unique colors for final-combiner, but NV2A does. See http://xboxdevwiki.net/NV2A/Pixel_Combiner#Per_stage_constant-colors

The syntax of nvparse only allows global-constants (type `ConstColor`) before `GeneralCombiners` (if any are present).
They are used to initialize the local-constants of each general-combiner, but they are also used as input for `FinalCombiner`

This means it's now allowed to do stuff like `GeneralCombiners ConstColor FinalCombiner`:

```c
!!RC1.0

// GeneralCombiners
{ rgb { spare0.rgb = spare0.rgb; } }

// ConstColor
const0 = (0,0,0,0); // error!

// FinalCombiner
out.rgb = const0.rgb;
out.a = const0.a;
```

This syntax of nvparse could be improved by allowing `ConstColor` declarations before general-combiners (for all general stages), but also before `FinalCombiner` (for final combiner).
The value before general-combiners would still be used to initialize the final-combiner constants.
This is similar to per-stage locals being initialized by globals.

I believe the easiest way of achieving this is to combine `ConstColor` and `FinalCombiner`:

```yacc
ColoredFinalCombiner : ConstColor ConstColor FinalCombiner`
                     | ConstColor FinalCombiner`
                     | FinalCombiner`
```

Currently we allow `ConstColor FinalCombiner` which would look like his:

```c
!!RC1.0

// ConstColor
const0 = (0,0,0,0); // This belongs to combinersStruct.
                    // Can be used in GeneralCombiners and FinalCombiner.

// FinalCombiner
out.rgb = const0.rgb;
out.a = const0.a;
```

Instead, we could replace this by `ColoredFinalCombiner` so this becomes:

```c
!!RC1.0

// ColoredFinalCombiner
const0 = (0,0,0,0); // This belongs to finalCombinerStruct now.
                    // Can only be used in FinalCombiner.
out.rgb = const0.rgb;
out.a = const0.a;
```

This would still work for `ConstColor GeneralCombiners ColoredFinalCombiner`:

```c
!!RC1.0

// ConstColor
const0 = (0,0,0,0); // This belongs to combinersStruct.
                    // Can be used in GeneralCombiners and FinalCombiner.
                    // In this case, used by GeneralCombiners.
const1 = (1,1,1,1); // This belongs to combinersStruct.
[...]

PR #199, opened 2019-10-08
fp20compiler: Fix constant-color emitter

### LINKED ISSUE #132: fp20compiler generates invalid pbkit pushes
Cg code like `return color * I.col * 0.3;` will introduce constants which will use code like this:

```c
pb_push4f(p, NV097_SET_COMBINER_FACTOR0 + 0 * 4, 0.300000, 0.300000, 0.300000, 0.000000);
```

However, the `NV097_SET_COMBINER_FACTOR0/1` and `NV097_SET_SPECULAR_FOG_FACTOR` registers are single u32 registers, with 4x 8bit fields: `0xAARRGGBB`

JayFoxRox:
Closes #132.

For background infomation, see [this XboxDevWiki article for a description of register-combiners](http://xboxdevwiki.net/NV2A/Pixel_Combiner#Register_combiners).

The old emitter code..
- ..erroneously chose per-stage constant-colors (`FACTOR#_EACH_STAGE` instead of `FACTOR#_SAME_FACTOR_ALL`), depending on number of constant-colors defined in global scope. Instead, this selection should depend on the number of local-colors (per-stage) instead. This needed a larger refactor, see below.
- ..emitted code had a typo for final combiner `const1` which lead to compilation errors if it was used.
- ..pushed colors in 4 floats. This is wrong and caused GPU command overflows / garbage colors.
- ..updated wrong constant in almost any case (used wrong variable in emitter-`printf`).

This PR fixes all of those issues.

The old GL code from nvparse was restored for better `FACTOR#_EACH_STAGE` / `FACTOR#_SAME_FACTOR_ALL` mode selection and constant handling, but minor changes were made for NV2A:
- As the final-combiners constant-colors are separate in NV2A, we always emit our global constant-colors for the final-combiner. This could change in the future, see #197.
- For `FACTOR#_SAME_FACTOR_ALL`, we also have to emit the global-constants into the first stage . See comment in code and XboxDevWiki for details.

Some minor issues:

- When mixing globals and locals, the globals are inlined.
    There are no checks if constants are even used, so this leads to dead-code.
    This is an issue / feature of the original nvparse implementation.
    It's a minor annoyance for most use-cases (useless writes to color registers).
    We should create an issue after merge.
- We don't split `FACTOR#_SAME_FACTOR_ALL` / `FACTOR#_EACH_STAGE` for `const0` and `const1`.
    This is possible with NV2A, but not in GL, so fp20compiler has no code to handle this.
    At the moment, if any `const0` or `const1` is defined locally, both will be set to `FACTOR#_EACH_STAGE`.
    This might affect performance.
    We should create an issue after merge.

---

I have a lot of different tests for this feature in https://github.com/JayFoxRox/nxdk/pull/34.

### DIFF
--- a/tools/fp20compiler/rc1.0_combiners.cpp
+++ b/tools/fp20compiler/rc1.0_combiners.cpp
@@ -21,37 +21,56 @@ void CombinersStruct::Validate()
 
 void CombinersStruct::Invoke()
 {
-    // for (int i = 0; i < numConsts; i++)
-    //     glCombinerParameterfvNV(cc[i].reg.bits.name, &(cc[i].v[0]));
     assert(numConsts <= 2);
-
-    printf("pb_push1(p, NV097_SET_COMBINER_CONTROL,");
-    printf("\n    MASK(NV097_SET_COMBINER_CONTROL_FACTOR0, %s)",
-        numConsts >= 1 ? "NV097_SET_COMBINER_CONTROL_FACTOR1_SAME_FACTOR_ALL"
-            : "NV097_SET_COMBINER_CONTROL_FACTOR0_EACH_STAGE");
-    printf("\n    | MASK(NV097_SET_COMBINER_CONTROL_FACTOR1, %s)",
-        numConsts >= 2 ? "NV097_SET_COMBINER_CONTROL_FACTOR1_SAME_FACTOR_ALL"
-            : "NV097_SET_COMBINER_CONTROL_FACTOR1_EACH_STAGE");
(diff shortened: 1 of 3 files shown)


dracc:
I've ran the tests as linked in the description.
They all worked as intended on real hw.
The code also looks good to me. :+1:

JayFoxRox:
I'll probably merge this in a couple of hours, unless there's a veto. If anybody still wants to review, do that now.

PR #202, opened 2019-10-13
Cpp build fix for SDL_TTF and friends

### LINKED ISSUE #201: NXDK_SDL=y only works for C, but not in C++
Renaming samples/sdl_ttf/main.c to main.cpp will cause the following error:

```
samples/sdl_ttf/main.cpp:5:10: fatal error: 'SDL.h' file not found
#include <SDL.h>
```

This is not specific to that sample and happens with all C++ projects which use nxdk makefiles.
That happens, because we only modify `CFLAGS`, but not `CXXFLAGS` in the SDL makefile:

https://github.com/XboxDev/nxdk/blob/b7d9ac8eee07e9a1db0215c091ec191098004c04/lib/sdl/Makefile#L22-L27

Additionally, we should be extending the `NXDK_` variants of `CFLAGS`/`CXXFLAGS` (to make it more consistent, as already partially done in #188).

*Similar issues might also be present in other nxdk code (which might require more checking for C++ header compatibility).*

dracc:
Closes #201 
This fixes building C++ apps with SDL_ttf and freetype.
We'd need an update to the nxdk-sdl repo too for it to work all out though.
XboxDev/nxdk-sdl#12 could be that fix.

### DIFF
--- a/lib/sdl/Makefile
+++ b/lib/sdl/Makefile
@@ -25,6 +25,11 @@ CFLAGS += -I$(NXDK_DIR)/lib/sdl \
           -I$(FREETYPE_DIR)/include \
           -DXBOX
 
+CXXFLAGS += -I$(NXDK_DIR)/lib/sdl \
+          -I$(SDL_TTF_DIR) \
+          -I$(FREETYPE_DIR)/include \
+          -DXBOX
+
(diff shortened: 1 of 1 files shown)


dracc:
Submodule update commit added.

Voxel9:
Can confirm this has fixed the problems I was experiencing last week. Everything's still working fine after removing all temporary hacks and doing a clean compile.

PR #203, opened 2019-10-14
Remove XReboot and return value from main

### LINKED ISSUE #91: XReboot() should not be part of the samples
@JayFoxRox suggested that `XReboot()` should not be a part of the nxdk samples, returning from `main` should be enough.
Changing the default behavior on returning from main might be necessary though.

kl0wn:
This PR

1. changes samples to int main() from void
2. removes XReboot() from main as pdclib handles this after returning form main

Fixes #91 

Some of the samples were returning -1 on pb_init() failure, @JayFoxRox suggested returning 1 for failure so I've made that change, but that can be up for discussion if others disagree

### DIFF
--- a/samples/hello/main.c
+++ b/samples/hello/main.c
@@ -5,7 +5,7 @@
 #include <windows.h>
 #include "stdio.h"
 
-void main(void)
+int main(void)
 {
     XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
 
@@ -14,8 +14,7 @@ void main(void)
         case 0: break;
         default:
             Sleep(2000);
-            XReboot();
-            return;
+            return 1;
     }
 
     pb_show_debug_screen();
@@ -26,5 +25,5 @@ void main(void)
     }
 
     pb_kill();
-    XReboot();
+    return 0;
 }
--- a/samples/httpd/main.c
+++ b/samples/httpd/main.c
@@ -39,7 +39,7 @@ static void packet_timer(void *arg)
   sys_timeout(PKT_TMR_INTERVAL, packet_timer, NULL);
 }
 
-void main(void)
+int main(void)
 {
 	sys_sem_t init_complete;
 	const ip4_addr_t *ip;
@@ -77,7 +77,7 @@ void main(void)
 	                     NULL, nforceif_init, ethernet_input);
 	if (!g_pnetif) {
 		debugPrint("netif_add failed\n");
-		return;
+		return 1;
 	}
 
 	netif_set_default(g_pnetif);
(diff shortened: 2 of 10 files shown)


samples/httpd/main.c
```diff
@@ -28,82 +28,82 @@ static void packet_timer(void *arg);
...
+    debugPrint("\n");
 
-	http_server_netconn_init();
-	while (1) NtYieldExecution();
-	Pktdrv_Quit();
-	return;
+    http_server_netconn_init();
+    while (1) NtYieldExecution();
+    Pktdrv_Quit();
+    return 0;
```
dracc: This file is now consistently spaces instead of tabs plus some spaces. I'll allow it.

JayFoxRox: Generally a good idea, but I feel like we should have this in a separate commit.

kl0wn: I reverted the spacing back to its original state, I think I hit auto format in VS in that file which caused that

dracc: Seems to have worked out. Nice!

samples/winapi_filefind/main.c
```diff
@@ -16,7 +16,7 @@ int mount_drive_c ()
     if (!NT_SUCCESS(status)) {
         debugPrint("Failed to mount C: drive!\n");
         Sleep(5000);
-        return -1;
+        return 1;
```
JayFoxRox: I'd keep this at `-1` and instead of returning `ret` from main (after checking result from `mount_drive_c`) return `1` regardless.

That is cleaner because in C world (outside of `main`) we'd also expect negative numbers for errors. Someone might extend this function with more return codes in the future, so it shouldn't be tied directly to `main` in the first place.

kl0wn: I've changed it to return -1 in the mount_drive_c function and return 1 from main on failure.

samples/hello/main.c
```diff
@@ -5,7 +5,7 @@
 #include <windows.h>
 #include "stdio.h"
 
-void main(void)
+int main()
```
JayFoxRox: As discussed on Discord, all of these have to be either:
```c
int main(void)
```
or
```c
int main(int argc, char* argv[])
```

see https://www.pdf-archive.com/2014/10/02/ansi-iso-9899-1990-1/ansi-iso-9899-1990-1.pdf page 6 (Page 16 in PDF).

Both styles have their pros and cons. I'd suggest using `(void)` for now.
If we ever introduce command line argument parsing from debuggers, we can add new samples which showcase the `argc`/`argv` style.

kl0wn: Changed to (void)

PR #204, opened 2019-10-15
libc: Add extension to always define errno_t in errno.h

thrimbor:
This adds `errno_ext_.h`, which, with the help of the extension hook system, ensures that `errno_t` always defined when including `errno.h` (as expected by libc++).

### DIFF
--- a/lib/xboxrt/libc_extensions/errno_ext_.h
+++ b/lib/xboxrt/libc_extensions/errno_ext_.h
@@ -1,0 +1,6 @@
+// Make sure errno_t is always defined when including errno.h, mimicking the
+// behavior of Win32 headers as expected by libc++.
+#ifndef _PDCLIB_ERRNO_T_DEFINED
+#define _PDCLIB_ERRNO_T_DEFINED _PDCLIB_ERRNO_T_DEFINED
+typedef int errno_t;
+#endif


JayFoxRox:
Note to others, that [this is an extension for this code](https://github.com/DevSolar/pdclib/blob/4f5b4339a48890599d8fb85bfa2fc6f031d47fe6/include/errno.h#L182-L189) which conditionally declares `errno_t`.

I wonder if libc++ might simply depend on `__STDC_WANT_LIB_EXT1__` / C11. In that case, solving it without this extension might be a better idea.

Which standard requires `errno_t` to always be defined? What platform is libc++ targeting?

thrimbor:
> I wonder if libc++ might simply depend on `__STDC_WANT_LIB_EXT1__` / C11. In that case, solving it without this extension might be a better idea.

libc++ doesn't depend on C11, as support for that is very poor on Windows so far. Solving it with `__STDC_WANT_LIB_EXT1__` would be possible, but that touches a whole bunch of other stuff.

> Which standard requires `errno_t` to always be defined?

None, afaik. But Win32 does it and libc++ expects it.

> What platform is libc++ targeting?

Win32, in this case.

JayFoxRox:
> > I wonder if libc++ might simply depend on `__STDC_WANT_LIB_EXT1__` / C11. In that case, solving it without this extension might be a better idea.
> 
> libc++ doesn't depend on C11, as support for that is very poor on Windows so far. Solving it with `__STDC_WANT_LIB_EXT1__` would be possible, but that touches a whole bunch of other stuff.
>
> [...]
>
> > What platform is libc++ targeting?
> 
> Win32, in this case.

Sorry, I meant which C standard and environment libc++ targets.
Or, put differently: Where do they get `errno_t` from, if it's not C11 / `__STDC_WANT_LIB_EXT1__`. You actually did answer it by saying it comes from Win32.

> > Which standard requires `errno_t` to always be defined?
> 
> None, afaik. But Win32 does it and libc++ expects it.

But where does it come from in Win32? Is it also just declared unconditionally in MS errno.h?
(In that case, I'd call it a "MS standard / extension / variant").

We should document why / where such things come from.
For `strdup` this wasn't too critical to me in a recent PR (although, in hindsight we should have added a comment about it being POSIX / MS / .. compatibility + link to man-page maybe).
For niche features like `errno_t`, we should document it as not many people will be familiar with it. A comment should mention which standard / convention it is from, or, if it's our own hack, we should document why that hack exists and when / how we could possibly get rid of it (short-note in code + issue on GitHub probably).

thrimbor:
> But where does it come from in Win32? Is it also just declared unconditionally in MS errno.h?
> (In that case, I'd call it a "MS standard / extension / variant").

It's declared unconditionally in `corecrt.h` which gets included by `errno.h`.

> We should document why / where such things come from.
> For `strdup` this wasn't too critical to me in a recent PR (although, in hindsight we should have added a comment about it being POSIX / MS / .. compatibility + link to man-page maybe).
> For niche features like `errno_t`, we should document it as not many people will be familiar with it. A comment should mention which standard / convention it is from, or, if it's our own hack, we should document why that hack exists and when / how we could possibly get rid of it (short-note in code + issue on GitHub probably).

True, I updated the code comment.

PR #205, opened 2019-10-15
libc: Add extension to provide _vsnprintf alias

thrimbor:
This adds `stdio_ext_.h`, which, with the help of the extension hook system, defines `_vsnprintf`, which is a Microsoft-specific alias for `vsnprintf`, and is required by libc++.

### DIFF
--- a/lib/xboxrt/libc_extensions/stdio_ext_.h
+++ b/lib/xboxrt/libc_extensions/stdio_ext_.h
@@ -1,0 +1,3 @@
+// Define required MS-specific aliases
+#define _vsnprintf vsnprintf
+


lib/xboxrt/libc_extensions/stdio_ext_.h
```diff
@@ -0,0 +1,3 @@
+// Define required MS-specific aliases
+#define _vsnprintf vsnprintf
```
JayFoxRox: I'll remind about https://github.com/XboxDev/nxdk/pull/150#discussion_r313995171.

However, in this case, I feel the non-function-macro is fine because of the underscore (so the name is unlikely to be re-used).
Also the issue with the function macro is that you can probably no longer get a function pointer (something I did not think about in the linked comment).

We should still rethink all of our function aliases in the future. I believe this could be solved with function-pointers or compiler extensions (to create an actual alias).

PR #213, opened 2019-10-24
xboxrt: Add include-guard to string_ext_.h

dracc:
Including `string_ext_.h` results in compilation errors due to redefinition of `_strdup`.
Let's fix that.


dracc:
I did a booboo, missed the fact that you should include `string.h` and not this file. Closing.

PR #215, opened 2019-10-27
samples: Remove pb_* from samples not using pbkit

### LINKED ISSUE #90: Samples pb_init() is ugly
Most samples do a switch case for pb_init() similar to
```c
switch(pb_init()) {
  case 0: break;
  default:
    XSleep(2000);
    XReboot();
    return -1;
}
```

This is not pleasing on the eyes.
@JayFoxRox suggested a more standard
```c
int ret = pb_init();
if (ret != 0) {
  XSleep(2000);
  XReboot();
  return -1;
}
```
which looks nicer IMO. This is also the way most other libs I've worked with suggests initialization checking to be done.

### LINKED ISSUE #212: Samples initialize pbkit, even though they never use it
Some samples intialize pbkit, but continue with software rendering.
I guess the intention is to call `pb_show_debug_screen` - but that shouldn't be necessary.
pbkit doesn't even own that framebuffer..

Currently, the framebuffer is owned by nobody, but should probably be (optionally) claimable by XVideo (See #169).
XVideo is responsible for setting up the display mode and the framebuffer address used for output.

Who owns the framebuffer depends on who draws to it:
* Allocation for CPU rendering should probably be done in XVideo (reclaim the buffer used on application entry) or the application (creating a new buffer).
* Allocation for GPU rendering should probably be done in an abstraction layer over pbkit (important for features like render-to-texture).

So, pbkit should only be used if the GPU is required (for drawing or computations).
pbkit shouldn't own any framebuffer (at very least none that it doesn't use itself) - it should only own optimization structures (DMA descriptors / compression tables / ...).

---

I noticed this problem because `pb_init` fails if the framebuffer isn't 32-bit.
It also increases RAM usage because of GPU framebuffers (which pbkit currently owns, even though it probably shouldn't) and helper structures.
Not having to deal with pbkit would also simplify code which only uses CPU rendering.

However, there's one exception why we might still need pbkit (or a minimal GPU driver):
I'm not sure if it can happen, but maybe the GPU can be left in a bad state when switching applications. That could have side-effects in RAM. To avoid that, we'd have to stop the GPU at startup. Currently, `pb_init` takes care of such things.
Someone should construct a test to see if this is an issue.

dracc:
Closes #212

Edit: Closes #90

Edit2: Copied from 212:
**H**owever, there's one exception why we might still need pbkit (or a minimal GPU driver):
I'm not sure if it can happen, but maybe the GPU can be left in a bad state when switching applications. That could have side-effects in RAM. To avoid that, we'd have to stop the GPU at startup. Currently, pb_init takes care of such things.
Someone should construct a test to see if this is an issue.

### DIFF
--- a/samples/sdl/main.c
+++ b/samples/sdl/main.c
@@ -10,8 +10,6 @@
   freely.
 */
 /* Simple program:  Create a native window and attach an SDL renderer */
-#include <hal/debug.h>
-#include <pbkit/pbkit.h>
 #include <hal/xbox.h>
 #include <hal/video.h>
 #include <windows.h>
@@ -24,7 +22,6 @@ static void printSDLErrorAndReboot(void)
     debugPrint("SDL_Error: %s\n", SDL_GetError());
     debugPrint("Rebooting in 5 seconds.\n");
     Sleep(5000);
-    pb_kill();
     XReboot();
 }
 
@@ -367,14 +364,7 @@ void demo(void)
 int main(void)
 {
     XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
-    if (pb_init() != 0)
-    {
-        Sleep(2000);
-        return 1;
-    }
 
-    pb_show_debug_screen();
     demo();
-    pb_kill();
     return 0;
 }
--- a/samples/hello++/main.cpp
+++ b/samples/hello++/main.cpp
@@ -1,5 +1,4 @@
 #include <hal/debug.h>
-#include <pbkit/pbkit.h>
 #include <hal/video.h>
 #include <windows.h>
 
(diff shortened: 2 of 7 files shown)


samples/sdl/main.c
```diff
@@ -290,7 +287,7 @@ void demo(void)
         SDL_WINDOW_SHOWN);
     if(window == NULL)
     {
-        debugPrint( "Window could not be created!\n");
+        debugPrint("Window could not be created!\n");
```
dracc: Unintended change. Will let it stay as the extra whitespace was probably unintentional.

mborgerson:
Thanks

PR #218, opened 2019-10-31
ci: Fix Travis on Windows by ignoring whether taskkill failed

thrimbor:
Apparently something changed in Travis CI, chocolatey or msys2, which results in `gpg-agent.exe` no longer being around. This makes `taskkill` fail, which fails our builds currently.
Since I'm not sure why this is the case and whether it's going to stay that way, I added `|| true` to make it no longer fail the build instead of removing it completely.

I expect this to get replaced by #168 anyway.


thrimbor:
Given #168, the low risk of this change and the fact that the problem currently makes all the PRs fail, I've decided to go ahead and merge this early.

PR #219, opened 2019-11-05
C++ Makefile changes

thrimbor:
This upstreams these two necessary changes from my libc++ branch:
* Thread-safe initialization of static variables gets disabled. C++11 by default guarantees that static variables will get initialized in a thread-safe way the first time they're encountered, but this relies on functions we do not provide yet. This should be easy to tackle properly after adding C++11 thread support.
* RTTI generation gets disabled. Generating RTTI causes references to internal data structures that aren't in place yet. This can get re-enabled once we implement these (which will happen together with exception support).

Both changes are necessary to avoid linker errors when using libc++.


mborgerson:
We should probably create an issue for these two things so we can track why we have disabled them and that we plan to address them down the road.

PR #220, opened 2019-11-05
vcruntime exception and type_info classes

thrimbor:
Another PR with changes from my libc++ branch.

On Win32, `std::exception` (and a few classes inheriting from that) and `std::type_info` are actually implement as part of vcruntime, and not the C++ standard library. libc++ also relies on these vcruntime implementations, so this implements the required classes and functions.

I copied the behavior of MS's functions as good as I could and copied their naming of private member variables and functions. I also added one MS-specific extension, the `(const char *const, int)` constructor for `std::exception` - this constructor gets a second parameter, which is actually ignored, and only used to overload the constructor in order to be able to avoid string buffer allocations when creating an exception object, which saves memory and allows exceptions to be thrown even when out of memory.

The exception object itself is rather trivial, `*_What` points to the exception message which gets returned by `what()`, and `_DoFree` specifies whether a buffer was allocated to store the exception message. If a buffer was allocated, a new one gets allocated and the content copied when the exception is copied, and the buffer is released upon destruction of the object.

`type_info` is also as close to the MS-version as possible, but with the shortcoming that `name()`, which is supposed to return a string containing the demangled name of the type, isn't implemented yet. This will get resolved when proper RTTI support will be added.

### DIFF
--- a/lib/xboxrt/vcruntime/vcruntime_exception.cpp
+++ b/lib/xboxrt/vcruntime/vcruntime_exception.cpp
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2019 Stefan Schmidt
+ *
+ * Licensed under the MIT License
+ */
+
+#include <assert.h>
+#include <stdlib.h>
+#include <string.h>
+#include <vcruntime_exception.h>
+
+extern "C"
+{
+    void __cdecl __std_exception_copy (const __std_exception_data *_From, __std_exception_data *_To)
+    {
+        if (_From->_DoFree) {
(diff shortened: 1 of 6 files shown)


lib/xboxrt/vcruntime/vcruntime_exception.cpp
```diff
@@ -0,0 +1,33 @@
+#include <assert.h>
```
mborgerson: How would you like to license this code?

thrimbor: MIT license note added.

PR #222, opened 2019-11-11
pbkit: Fix type warnings by replacing DWORD with uint32_t

### LINKED ISSUE #175: warning: incompatible pointer types passing 'uint32_t *' to parameter of type 'DWORD *'
We cause many warnings in calls to `pb_push1` in our sample code.

`DWORD` is typically used by MS.
However, pbkit probably shouldn't be limited to running on MS platforms (think of offline FIFO compilation), so it should probably use `uint32_t`, or a custom type to reflect pointers.

Alternatively, we can assume that pbkit is always ran on MS platform, and then use `DWORD` in the callers instead.

thrimbor:
Fixes #175.

This doesn't replace all occurrences of `DWORD`, only those that currently produce warnings. Tested samples continued to work as they did before.

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -117,9 +117,9 @@ static  DWORD           *pb_DmaBuffer2; //points at 32 contiguous bytes (Dma Cha
 static  DWORD           *pb_DmaBuffer7; //points at 32 contiguous bytes (Dma Channel ID 7 buffer)
 
 static  DWORD           pb_Size=512*1024;//push buffer size, must be >64Kb and a power of 2
-static  DWORD           *pb_Head;   //points at push buffer head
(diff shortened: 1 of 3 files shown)


JayFoxRox:
I'm fine with this, but we should still fix *all* data types in pbkit in the near future.

mborgerson:
I don't love the mix-n-match of both MS and stdint.h and would prefer they all be moved over to one style or the other. I prefer stdint.h style. We can fix the rest in a follow up.

haxar:
I prefer to use `stdint.h` style types from now on, so that the integer type size is known & for portability purposes, and I currently do when importing winapi prototypes to xexec's own xboxkrnl implementation, which does winapi to POSIX call translation (but not the other way around yet).

PR #225, opened 2019-11-17
libc: Provide _sys_nerr macro via stdlib.h

thrimbor:
Win32 provides a macro called `_sys_nerr` to get the highest valid errno value. libc++ requires this, so this commit provides it.

### DIFF
--- a/lib/xboxrt/libc_extensions/stdlib_ext_.h
+++ b/lib/xboxrt/libc_extensions/stdlib_ext_.h
@@ -17,6 +17,10 @@ typedef void (__cdecl *_purecall_handler)(void);
 _purecall_handler __cdecl _get_purecall_handler (void);
 _purecall_handler __cdecl _set_purecall_handler (_purecall_handler function);
 
+// Win32 extension providing the highest valid errno number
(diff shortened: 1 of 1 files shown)


lib/xboxrt/libc_extensions/stdlib_ext_.h
```diff
@@ -17,6 +17,8 @@ typedef void (__cdecl *_purecall_handler)(void);
 _purecall_handler __cdecl _get_purecall_handler (void);
 _purecall_handler __cdecl _set_purecall_handler (_purecall_handler function);
 
+#define _sys_nerr _PDCLIB_ERRNO_MAX
```
JayFoxRox: `#include "pdclib/_PDCLIB_config.h"` is only ever done indirectly in the pdclib stdlib.h
Should we do it explicitly here, or just assume that is has happened?

thrimbor: I think it's fine, afaik the config is (indirectly) included in every header.

JayFoxRox: As usual, I think we should add a comment about *what* kind of extension this is.
For what API / standard and maybe a link to documentation.
\- I'm surprised I didn't mention this in my initial review.

thrimbor: Done.

PR #226, opened 2019-11-17
libc: Add stubs for strtod(), strtof(), strtold()

thrimbor:
libc++ depends on the presence of these three functions. PDCLib doesn't provide them yet, so this provides stubs until the functionality is in place. I added asserts to catch calls to these functions instead of silently returning invalid results, but these functions were never called by libc++ itself during my tests.


JayFoxRox:
Skimmed over the code and it looks fine; implicit casts for the return values are okay, too.

I'm a bit surprised that this isn't stubbed upstream. I'm fine with it being in our codebase for now though (as a non-working stub upstream might cause issues, too).

PR #228, opened 2019-11-17
libc: Add locale.h extensions required by libc++

thrimbor:
This adds the necessary stubs for functionality libc++ expects from `locale.h`. `_create_locale` is actually ran by libc++ when using it, so this code is somewhat tested.

### DIFF
--- a/lib/xboxrt/libc_extensions/locale_ext_.h
+++ b/lib/xboxrt/libc_extensions/locale_ext_.h
@@ -1,0 +1,27 @@
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// We don't really have locale support.
+// These are only Windows-specific stubs and placeholders, providing just the bare minimum for libc++.
+
+typedef void *_locale_t;
+
+static _locale_t _create_locale (int category, const char *locale)
+{
+    return NULL;
+}
+
+static void _free_locale (_locale_t locale) {}
+
+#define _ENABLE_PER_THREAD_LOCALE 1
+#define _DISABLE_PER_THREAD_LOCALE 2
+static int _configthreadlocale (int per_thread_locale_type)
+{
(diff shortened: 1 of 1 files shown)


lib/xboxrt/libc_extensions/locale_ext_.h
```diff
@@ -0,0 +1,26 @@
...
+    return NULL;
+}
+
+static void _free_locale (_locale_t locale) {}
+
+#define _ENABLE_PER_THREAD_LOCALE 1
+#define _DISABLE_PER_THREAD_LOCALE 2
+static int _configthreadlocale (int per_thread_locale_type)
+{
+    return _DISABLE_PER_THREAD_LOCALE;
```
JayFoxRox: MSDN says `-1` on failure (not being able to change it should be considered a failure to me). Would that work for libc++?

thrimbor: I checked, and libc++ uses the return value as a parameter to this same function later on, but apart from that, there's no checking or other use of it. As long as we ignore the parameter completely, returning `-1` seems to be fine, so I changed it.

lib/xboxrt/libc_extensions/locale_ext_.h
```diff
@@ -0,0 +1,27 @@
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// We don't really have locale support.
+// These are only stubs, providing just the bare minimum for libc++.
+
+typedef void *_locale_t;
+
+static _locale_t _create_locale (int category, const char *locale)
```
JayFoxRox: These functions (and type?) should be marked as Windows specific stuff. If we also add some POSIX extensions later we don't want to lose track.

thrimbor: The comment above it now mentions that it's Windows-specific.

PR #229, opened 2019-11-17
Implement Win32 threads

thrimbor:
This implements large parts of the Win32 threads API for nxdk. I originally intended for this PR to be much smaller, but SDL, which I used for testing, has quite a few requirements. On the positive side, this PR is enough to enable the Win32 threads backend in SDL with minimal changes, which I've done locally (test results were shown off in Discord).

A couple of notes on some of the commits:
* 076dc4dbba46e7aa02751be1402840d63688144b The code for `RaiseException` actually comes from my local work on getting exceptions working. It has been tested as part of that work.
* e6f64b11d72c3229db0187400f551bdce0c98d0a The implementation for `GetCurrentThread` might seem weird, but is in fact correct (and tested together with `GetThreadId()`). Like on Win32, it returns a pseudo-handle that refers to the currently running thread.
* 09512ffdcaed58ead1d6a0e5d645d1d6bccec48b currently assumes that we can't detect the presence of a debugger. Afaik SDL only uses `IsDebuggerPresent()` to decide whether to advertise thread names to the debugger, so this implementation is good enough for SDL to be happy, but I don't know enough about interacting with debugging stuff on the Xbox to know whether or how we can actually properly implement this function.
* f5f2890b687effbd3608ef3e8c8a171498393d06 `InitializeCriticalSectionAndSpinCount()` allows specifying the number of spins before the thread yields. This is meant for fine-tuning the locking behavior on multiprocessor machines, and obviously not useful (or provided) on the Xbox, so we can get just ignore the spin count.
* 01c0aeb4e3da0311f2a1fdf58233700a128b4cf8 The code for `CreateThread` is almost identical to the current C11 threads code - the main difference is the API.
* c4feb0a5a280f9f16df538f242f5c1908d5608be This implements the FLS (fiber-local storage) API - we don't actually provide any support for fibers, but this API is used by libc++ and SDL because, unlike the TLS API, it allows to specify destructors. It's roughly based on the code I wrote for C11, but it inverts the meaning of the bits in the bitmap to avoid having to use GNU extensions for initializing it, and it's a bit more complex because it implements the Win32 behavior of running destructors for other threads on `FlsFree` calls.
* 84cbac81de6664169e012215ad666a850a01714e Implements the newer thread functions used by SDL. Apparently there were some memory leak issues on Windows with the old API when statically linking, but for nxdk, we can simply wrap the `CreateThread` API here.


lib/winapi/winbase.h
```diff
@@ -14,6 +14,15 @@ extern "C"
...
 #define INFINITE 0xFFFFFFFF
 
+#define THREAD_PRIORITY_ABOVE_NORMAL 1
+#define THREAD_PRIOIRTY_BELOW_NORMAL (-1)
+#define THREAD_PRIORITY_HIGHEST 2
+#define THREAD_PRIORITY_IDLE (-15)
+#define THREAD_PRIORITY_LOWEST (-2)
+#define THREAD_PRIORITY_NORMAL 0
+#define THREAD_PRIORITY_TIME_CRITICAL 15
+#define THREAD_PRIORITY_ERROR_RETURN 2147483647
```
JayFoxRox: nit: why not `0x7FFFFFFF`?

lib/winapi/debug.c
```diff
@@ -1,8 +1,14 @@
-#include <winbase.h>
+#include <windows.h>
 #include <winnt.h>
 
 #include <string.h>
 
+BOOL IsDebuggerPresent (VOID)
+{
+    // FIXME: Can we detect debugger presence on the Xbox?
+    return FALSE;
```
JayFoxRox: There's `KdDebuggerEnabled` / `KdDebuggerNotPresent`.

XBDM also sets up some hooks into the KPCR or something. I think that's where it communicates performance counters - which [even work for retail games, as shown in my FPS overlay](https://github.com/JayFoxRox/xbox-fps-overlay).

Obviously this is something we should be discussing in a follow-up PR / issue, not here.

lib/winapi/sync.c
```diff
@@ -9,6 +9,12 @@ VOID InitializeCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
     RtlInitializeCriticalSection(lpCriticalSection);
 }
 
+BOOL InitializeCriticalSectionAndSpinCount (LPCRITICAL_SECTION lpCriticalSection, DWORD dwSpinCount)
+{
+    InitializeCriticalSection(lpCriticalSection);
```
JayFoxRox: We should add a comment that [cites the MSDN](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-initializecriticalsectionandspincount):

> On single-processor systems, the spin count is ignored and the critical section spin count is set to 0 (zero).

lib/winapi/thread.c
```diff
@@ -1,7 +1,64 @@
+#include <assert.h>
+#include <string.h>
 #include <processthreadsapi.h>
 #include <winbase.h>
+#include <pdclib/_PDCLIB_xbox_tls.h>
 #include <xboxkrnl/xboxkrnl.h>
 
+VOID NTAPI WinapiThreadStartup (PKSTART_ROUTINE StartRoutine, PVOID StartContext)
```
JayFoxRox: This feels like something that should be `static`?
It's not mentioned in the header and I couldn't find docs on MSDN.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -208,6 +208,8 @@ typedef CONST UNICODE_STRING *PCUNICODE_STRING;
     (Entry)->Next = (ListHead)->Next; \
     (ListHead)->Next = (Entry);
 
+#define CONTAINING_RECORD(Address, Type, Field) ((Type *)((PCHAR)(Address) - (ULONG_PTR)(&((Type *)0)->Field)))
```
JayFoxRox: I wonder if we should use https://gcc.gnu.org/onlinedocs/gcc/Offsetof.html#Offsetof

I don't remember why specifically, but the `NULL` pointer solution can fail or isn't valid C/C++ (warnings?).

lib/winapi/thread.c
```diff
@@ -1,7 +1,64 @@
...
+
+    // Zero-initialize the rest
+    RtlZeroMemory((char *)TlsData + TlsDataSize, _tls_used.SizeOfZeroFill);
+
+    int res;
+    res = (*(LPTHREAD_START_ROUTINE)StartRoutine)(StartContext);
+    PsTerminateSystemThread(res);
+}
+
+HANDLE CreateThread (LPSECURITY_ATTRIBUTES lpThreadAttributes, SIZE_T dwStackSize, LPTHREAD_START_ROUTINE lpStartAddress, LPVOID lpParameter, DWORD dwCreationFlags, LPDWORD lpThreadId)
```
JayFoxRox: `assert(lpThreadAttributes == NULL)` ?
Maybe even return error for invalid arguments?
At least there should be a comment about this not being supported.

lib/winapi/thread.c
```diff
@@ -1,7 +1,64 @@
...
+    }
+
+    ULONG tlssize;
+    // Sum up the required amount of memory, round it up to a multiple of
+    // 16 bytes and add 4 bytes for the self-reference
+    tlssize = (_tls_used.EndAddressOfRawData - _tls_used.StartAddressOfRawData) + _tls_used.SizeOfZeroFill;
+    tlssize = (tlssize + 15) & ~15;
+    tlssize += 4;
+
+    status = PsCreateSystemThreadEx(&handle, 0, dwStackSize, tlssize, (PHANDLE)lpThreadId, (PKSTART_ROUTINE)lpStartAddress, lpParameter, dwCreationFlags & CREATE_SUSPENDED ? TRUE : FALSE, FALSE, WinapiThreadStartup);
```
JayFoxRox: `dwCreationFlags` is never checked for `STACK_SIZE_PARAM_IS_A_RESERVATION`

lib/winapi/sync.c
thrimbor: Done.

lib/winapi/thread.c
thrimbor: Yeah that's a function I came up with - Windows has such a function as well, but I don't know the name they chose. Anyway, this should have been static, and it is now.

lib/winapi/thread.c
thrimbor: I added a comment, but I'd like to keep ignoring the parameter - using it doesn't make sense on the Xbox, but there may be code out there using it. Ignoring it should be fine for conforming to MSDN descriptions.

lib/winapi/thread.c
thrimbor: Right, but that's because we don't support that - the kernel always expects the whole stack to be committed memory, growing is not supported (that's why we simply check for problems in `chkstk` instead of probing the memory).

lib/winapi/winbase.h
thrimbor: Changed.

lib/xboxkrnl/xboxkrnl.h
thrimbor: I'm not aware of problems with the `NULL` pointer solution, but I switched to `__builtin_offsetof` anyway to get rid of any issues there might be while not having to include C headers.

lib/winapi/thread.c
JayFoxRox: We should probably add a comment in the functions which receive this parameter, that it isn't supported (or where you'd expect a declaration to be, so people don't just declare it themselves).

lib/winapi/debug.c
JayFoxRox: "winapi: Implement IsDebuggerPresent() "

Commit title is bad; this merely stubs it.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -2386,7 +2386,7 @@ XBAPI VOID NTAPI READ_PORT_BUFFER_UCHAR
     IN ULONG Count
 );
 
-XBAPI volatile OBJECT_TYPE PsThreadObjectType;
+XBAPI OBJECT_TYPE PsThreadObjectType;
```
JayFoxRox: The volatile keyword seems to exist for all kernel exports.
I'm not sure if it's needed (does the compiler recognize `DATA` from an import lib might be modified by other code?).

However, if it's not needed for this `ObjectType`, then it's likely not needed for *any* `ObjectType`.

I feel like this is a premature optimization or incomplete code-cleanup which should be in a separate PR.

lib/winapi/basetyps.h
```diff
@@ -0,0 +1,5 @@
+#include <guiddef.h>
+
+typedef GUID IID;
+typedef IID *REFIID;
```
JayFoxRox: These types never seem to be used in this PR.
What will they be used for?

I mostly know these from COM - and we likely won't implement such features in nxdk.

lib/winapi/thread.c
thrimbor: Actually, declaring (and using) `STACK_SIZE_PARAM_IS_A_RESERVATION` is fine (I just now added the define for compatibility). It will simply fall back to using committed memory instead.

I added a comment before the declaration of `CreateThread` though to make people aware that the flag is ignored.

lib/winapi/debug.c
thrimbor: Fixed.

lib/xboxkrnl/xboxkrnl.h
thrimbor: `PsThreadObjectType` is not expected to change, so the `volatile` is unnecessary either way (we're only passing it's address to kernel APIs anyway). It's removed here because keeping it would mean that code using it (e.g. passing it to `ObReferenceObjectByHandle`, which is done in code submitted here) causes warnings about discarding qualifiers.

lib/winapi/basetyps.h
thrimbor: These typedefs are required by SDL's threading implementation which I used to test this code.

lib/winapi/processthreadsapi.h
```diff
@@ -1,16 +1,31 @@
...
 #include <windef.h>
 
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
 #endif
 
+typedef DWORD (__stdcall *LPTHREAD_START_ROUTINE) (LPVOID lpThreadParameter);
+// The STACK_SIZE_PARAM_IS_A_RESERVATION flag has no effect, stacks are always committed memory in the Xbox
```
JayFoxRox: Comment should be with the constant; [also affects `_beginthread`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/beginthread-beginthreadex?view=vs-2019) (which is where I originally noticed the lack of support)

lib/xboxkrnl/xboxkrnl.h
JayFoxRox: If we'd get a warning about `volatile` qualifier discarding here, wouldn't we also get one in the drive mounting PR for those object-types in similar contexts?
If we make such assumptions about one `ObjectType`, shouldn't we make it for others, too (for consistency)?

lib/winapi/basetyps.h
JayFoxRox: The only one we might include is <nxdk-sdl>/src/core/windows/SDL_windows.h which also drags in a lot of other Windows garbage (\- COM related; `CoInitializeEx` etc.). How is that dealt with in nxdk?

It feels like we should disable those parts in SDL for Xbox compilation.

Or am I looking at wrong code? Where do these appear?

lib/xboxkrnl/xboxkrnl.h
thrimbor: The drive mounting PR doesn't touch any `OBJECT_TYPE`s, so no warnings there. Afaik we only use `ObReferenceObjectByHandle` in [GetThreadId](https://github.com/XboxDev/nxdk/blob/ef7735d177c8877b002ef19ac569c9382063978b/lib/winapi/thread.c#L16), but the explicit cast seems to suppress the warning.

Removing `volatile` should be fine for (at least) all `OBJECT_TYPE` exports from the kernel - I didn't want to do more changes than necessary to get this working, but I can integrate the changes for the other occurrences if you want.

lib/winapi/basetyps.h
thrimbor: `SDL_windows.h` is indeed the file where this is required. We currently never build any code that uses the stuff declared there, that's why we don't get any unresolved references or have to deal with `CoInitialize`-stuff.

It's probably better to work around this in SDL instead - if you agree that I should remove it here I'll probably do that later today.

lib/winapi/processthreadsapi.h
thrimbor: Done.

lib/winapi/basetyps.h
JayFoxRox: Yes, we should remove this stuff from SDL, when compiling for Xbox

lib/xboxkrnl/xboxkrnl.h
JayFoxRox: Oops! I could have sworn that the drive mounting PR used `ObDirectoryObjectType` or `ObSymbolicLinkObjectType`.

I wonder if we should just tolerate the warning for now (= remove this commit), then have a different PR (or Discord discussion) about the volatile keyword.
I think there are good reasons to keep them (things like `KeTickCount`), but also good reasons to remove them (useless if data is passed around / some kernel objects aren't expected to change).
There are also good reasons to remove them everywhere (consistency), but also good reasons to only remove them for some variables (different usage-patterns of different variables).

Removing only *this* volatile keyword seems like a hack that introduces inconsistencies and unexpected user-code (where they manually have to add volatile sometimes; even for similar variables). But removing them everywhere might have unexpected consequences, too.

I'm also not familiar enough with the `volatile` implications to know how effective it really is (does a volatile pointer imply volatile access on the data it points to?).
I'm also not familiar enough with the kernel variables to know which ones *should* be volatile (if any).
I'm also not sure if it makes a difference that these variables are from an import lib (where they are being turned into pointers in the kernel stub anyway).

thrimbor: I dropped the commit for now. I'll probably create a separate PR for it sometime later.

> I'm also not sure if it makes a difference that these variables are from an import lib
I'd think so - it seems like this would be the sane thing to do, but it's just a guess. I'll have to read up on the details before even considering changing anything other than the `OBJECT_TYPE`s.

lib/winapi/basetyps.h
thrimbor: Done.

lib/xboxkrnl/xboxkrnl.h
JayFoxRox: >  I'll have to read up on the details before even considering changing anything other than the `OBJECT_TYPE`s.

If you don't create a PR for `volatile` anytime soon, create an issue please (so we don't forget + someone might know how to deal with this).

lib/winapi/basetyps.h
JayFoxRox: Clarification: "Done" by removing these types.
No nxdk-sdl changes have been shown yet.

mborgerson:
Thanks!

lib/winapi/debug.c
JayFoxRox: We should create an issue about this (marking it as a stub, and discussing the possibility / implementation options)

lib/xboxkrnl/xboxkrnl.h
JayFoxRox: Still needs issue or new PR

PR #230, opened 2019-11-20
Implement drive mounting

thrimbor:
This implements @LukeUsher 's proposal from [here](https://discordapp.com/channels/428359196719972353/428360618102226946/572440894704189460).

Three new functions are added:
* `XMountDrive` creates the symbolic link for a drive letter. It fails when the link already exists and reports errors winapi-style.
* `XUnMountDrive` removes a symbolic link for a drive letter, errors are also reported winapi-style.
* `XReMountDrive` tries to remove a symbolic link first, but ignores whether that fails, and then creates a new symbolic link.

`XMountDrive` was previously provided by `hal/fileio.c` - I removed that, it wasn't used anywhere in nxdk. My new code is fresh and not based on the old one.

Following the above proposal, another function is automatically run on startup, which checks whether `D:` is already available, and if not, it uses the path of the currently running XBE to create it.

The file listing sample has also been updated to utilize the new functions.


lib/winapi/xapi.c
```diff
@@ -5,9 +5,44 @@
  */
 
 #include <xapi.h>
+#include <assert.h>
 #include <stdio.h>
+#include <string.h>
 #include <xboxkrnl/xboxkrnl.h>
 
+__attribute__((constructor)) static VOID setup_d_drive (VOID)
```
JayFoxRox: This should be using `void` instead of `VOID`.

JayFoxRox: I don't like the idea of forcing constructors on people.

One solution would be to split this into a separate library, so only if that library is linked, it would do the setup.
That way, no source-code changes are necessary (for *very* naive ports) and the user can also avoid including this code at all if they don't want it (for low-level programs for example).

lib/winapi/xapi.c
```diff
@@ -5,9 +5,44 @@
...
+
+    // Retrieve the XBE path
+    strncpy(targetPath, XeImageFileName->Buffer, XeImageFileName->Length);
+    targetPath[XeImageFileName->Length] = '\0';
+
+    // Cut off the XBE file name by inserting a null-terminator
+    PCHAR filenameStr;
+    filenameStr = strrchr(targetPath, '\\');
+    assert(filenameStr != NULL);
+    *(filenameStr + 1) = '\0';
```
JayFoxRox: This retrieval of the XBE path should ideally be a separate function as we might want to re-use that in the future?
Or more generally, some `dirname`-like function?

lib/winapi/xapi.c
```diff
@@ -5,9 +5,44 @@
...
+    HANDLE handle;
+    ANSI_STRING driveString = RTL_CONSTANT_STRING("\\??\\D:");
+    OBJECT_ATTRIBUTES objattr;
+
+    InitializeObjectAttributes(&objattr, &driveString, OBJ_CASE_INSENSITIVE, NULL, NULL);
+    status = NtOpenSymbolicLinkObject(&handle, &objattr);
+    if (NT_SUCCESS(status)) {
+        NtClose(handle);
+        return;
+    }
```
JayFoxRox: This could be its own function to test if a drive is mounted?

lib/winapi/xapi.c
```diff
@@ -0,0 +1,89 @@
...
+
+    return TRUE;
+}
+
+BOOL XReMountDrive (CHAR driveLetter, PCSTR path)
+{
+    XUnMountDrive(driveLetter);
+    return XMountDrive(driveLetter, path);
+}
```
JayFoxRox: I don't think this function is necessary? If the path is already mounted, then the user should probably look into dealing with it (why was it mounted? maybe it had a purpose?).

I think a more suitable API design would possibly be to be able to `bool force` a `XMountDrive`.

Eitherway, I see drawbacks when unmounting succeeds, but remounting fails.

---

Realisticly, I assume that we'll provide some optional XDK XAPI-like lib which respects XBE flags and does the XDK-like mounting at startup. Most applications would probably use that. So it's unlikely the user would want to manually mount something (and if they do, they are probably fine with more steps).

JayFoxRox:
I have already criticized the placement of this on Discord. I don't think it should be in winapi (which implies documention in MSDN, availability on Windows and design by MS).

The argument for having it in winapi was that it uses the error reporting mechanism and interacts with the winapi file-system functions.
If we think this is enough (I disagree) then we should still move it in some winapi_extensions or winapi/extensions to make clear that this isn't MS stuff. If we really keep it in winapi, I'd also want to avoid the X prefix (as winapi wouldn't typically do that either).
The extensions should also be their own library (so winapi can be linked without them).

I think we should have an Xbox specific lib (or multiple thereof). We also have an X prefixed API for the LED in hal/led.c which might be suitable for that.
Currently that lib is "hal", although that's a bad name in my opinion.

lib/winapi/xapi.c
```diff
@@ -0,0 +1,89 @@
...
+    status = IoCreateSymbolicLink(&drivePath, &mountPath);
+    if (!NT_SUCCESS(status)) {
+        SetLastError(RtlNtStatusToDosError(status));
+        return FALSE;
+    }
+
+    return TRUE;
+}
+
+BOOL XUnMountDrive (CHAR driveLetter)
```
JayFoxRox: I prefer `Unmount` over `UnMount`

lib/winapi/xapi.c
LukeUsher: Are we sure we need to setup D: ourselves? The kernel should handle this based on the title path passed to the launch data page on reboot. Official XDK games don't handle setting up D themselves, but rely on the value set by the previously running Xbe (passed to the kernel via the launch data page when calling HalReturnToFirmware)

dracc:
I mean, I guess I could (should?) use a device path instead of this nice naming scheme but when writing a DVD dumper, having D: mounted to the xbe location is not helpful.

lib/winapi/xapi.c
thrimbor: Done.

lib/winapi/xapi.c
thrimbor: Removed.

> So it's unlikely the user would want to manually re-mount something (and if they do, they are probably fine with more steps).
If we're talking about games or something similar here, I agree. But I think it's quite likely that utility programs want to access the full partitions instead of only game-specific paths.

lib/winapi/xapi.c
thrimbor: I moved it do a `XGetXbePath` function now.

thrimbor:
@JayFoxRox A separate Xbox specific lib would be fine with me. I struggle with coming up with a name though - `hal` really is a bad name (and one I'm trying to slowly get rid of), and naming it `xapi` could make people think that this is an XDK-compatible implementation.

@dracc Actually, having `D:` point to the Xbe location is what pretty much every code I saw out there expects (even our current `hal` code does that, although in a horrible way). A DVD dumper should always mount the drive explicitly (and probably not even to `D:`), it can't expect a pre-mounted path to point to the drive.

lib/winapi/xapi.c
JayFoxRox: Resolving this. If you want to continue discussing the D:/ situation, create a new discussion / review comment chain.

JayFoxRox: My idea was to always link this lib by default, but provide an option to *not* link it (using some Makefile variable, analogous to how we have the `NXDK_SDL=y` stuff).
Similar approaches would also still work in future build-systems probably.

We could also use weak symbols / functions in the future to have some "override" library which just stubs these kind of constructors (so the code doesn't even end up in the binary).

lib/winapi/xapi.c
JayFoxRox: I was specifically thinking of something like a file-manager. Those utilities would want to ensure that the paths they re-mounted are mounted properly (as they don't want users to accidentally manipulate the wrong filesystem). So they are fine with extra-steps and more careful error-checking.

JayFoxRox:
@thrimbor "xapi" is equally bad.

If we split it into a new lib, then I think "nxdk" would be an okay name (which could be split into different libs for different subsystems, like "nxdk_paths", "nxdk_led", "nxdk_video", ...) or "xbox_helper".

If it remains in winapi (which might make sense) then it should be in "winapi/extensions/xbox_paths.c" (or winapi_extensions/xbox_paths.c) or something; the "X" prefix for functions should be removed, and it should be built into a separate lib like `winapi_xbox_paths.lib` (rather fine-grained).

The auto-mounter could be something like `winapi_xbox_auto_mount_d.lib`. Configuration could work using something like `NXDK_AUTO_MOUNT_D=<Path>`.
It could be "NONE", "XBE_PATH" or an actual path. The first option would be handled in the Makefile (disabling the lib). The last option would be handled by a macro / some symbol we'd add to the binary (to communicate the path, or lack thereof, to the constructor).

thrimbor:
Ok, I've refactored the PR. This is how it's set up now:

There is a new lib called `nxdk`, which includes all the new stuff except the constructor responsible for mounting `D:`. The `xapi` name is gone, instead, the header is now called `mount.h` (it's a bit of a generic name, but since this library's directory is not one of the header search paths, there's only a low risk of conflicts - it always has to be included with `#include <nxdk/mount.h>`). I kept the X-prefix though (I think it makes sense to prefix all Xbox-specific functionality like this).

Additionally, there's a second lib called `nxdk_automount_d`. It only gets linked against if `NXDK_DISABLE_AUTOMOUNT_D` is not set to `y`. It also sets an additional linker flag to not ignore the constructor, which would otherwise happen because it's not referenced from outside the library.
I tested this new build option, and it does work as expected.

I decided to keep it (relatively) simple instead of providing a full-fledged `NXDK_AUTO_MOUNT_D=<Path>` - I assume that everyone not wanting automatic mounting will be fine with doing the mounting manually anyway.

lib/nxdk/mount.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <winbase.h>
+
+#ifdef __cplusplus
+{
+#endif
+
+BOOL XIsDriveMounted (CHAR driveLetter);
+BOOL XMountDrive (CHAR driveLetter, PCSTR path);
+BOOL XUnmountDrive (CHAR driveLetter);
```
JayFoxRox: If we don't keep this in winapi, then I'd very much prefer C types like `bool` and `char`

lib/nxdk/xbe.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifdef __cplusplus
+{
+#endif
+
+#ifndef __cplusplus
+#include <stdbool.h>
+#endif
+
+void XGetXbePath (char *path, bool includeFileName);
```
JayFoxRox: My original idea for moving this into a more generic `paths.h` was to make this more extensible.
There's other standard parts the user might want to query like `XGetSavedataPath(..., uint32_t titleId);` or `XGetDashboardXbePath(...)`.

The path of the current XBE also isn't really a part of the XBE itself. I'd prefer to reserve the `xbe.h` name for functions and types that actually deal with XBE images.

JayFoxRox: As indicated by my other comment regarding `XGetDashboardXbePath`, it would make more sense to be more explicit like `XGetCurrentXbePath()` (`CurrentXbe`? `RunningXbe`? `LoadedXbe`? `TitleXbe`? ...)

JayFoxRox: As these allocate new memory, the `includeFileName` parameter is no longer needed if we also provide something like `dirname()` and `basename()`:

```c
// void XGetXbePath (char *path, bool includeFileName); 
XGetXbePath(tmp, true);  // D:/default.xbe
XGetXbePath(tmp, false);  // D:/

// void XGetXbePath (char *path); 
XGetXbePath(tmp); // D:/default.xbe
dirname(strdup(tmp)); // D:/
basename(strdup(tmp)); // default.xbe
```

I feel this would be a more powerful API that is more extensible in the future (although somewhat minimal in this instance).

JayFoxRox:
> I decided to keep it (relatively) simple instead of providing a full-fledged NXDK_AUTO_MOUNT_D=<Path> - I assume that everyone not wanting automatic mounting will be fine with doing the mounting manually anyway.

Fair enough, the new variable is a niche use-case, so breaking it in the future isn't *too* bad probably.
The idea behind the full-style was future-proofness; e.g: `NXDK_AUTO_MOUNT_C="\\Device\\Harddisk0\\Partition2\\"`; maybe I'm just overcomplicating it though.

samples/winapi_filefind/main.c
```diff
@@ -1,33 +1,20 @@
...
 {
     XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
 
     // Mount C:
-    int ret = mount_drive_c();
-    if (ret != 0) {
+    BOOL ret = XMountDrive('C', "\\Device\\Harddisk0\\Partition2\\");
+    if (!ret) {
+        debugPrint("Failed to mount C: drive!\n");
+        Sleep(5000);
```
JayFoxRox: Note: We could avoid this by just adding more files to the XBE path and querying it.

lib/nxdk/xbe.h
JayFoxRox: Does this return a path like `D:/default.xbe` or `\Device\\Harddisk0\\Partition2\\default.xbe`?

lib/nxdk/mount.h
thrimbor: Done.

lib/nxdk/xbe.h
thrimbor: I wasn't really sure where to put it, but yeah, `path.h` makes more sense.

thrimbor: Renamed to `XGetCurrentXbePath`.

thrimbor: I don't follow - what allocates memory?

As for the POSIX functions: While I'd usually agree that they're better than having an additional parameter, my intention was to keep things simple for the constructor (not having to check for failed allocations etc.), as it is run very early. Tbh if you prefer to have the parameter gone, I'd rather move the `strrchr` stuff to the constructor and leave the task of implementing `dirname` and `basename` for later.

thrimbor: Like `\Device\CdRom0\default.xbe`.

samples/winapi_filefind/main.c
thrimbor: True, but I think it makes sense to include a simple mounting demonstration here (the downside is that the output of this depends on the user's system, so it's not perfect as a test).

lib/nxdk/xbe.h
JayFoxRox: Sorry, I meant they use newly allocated memory (the function doesn't modify the `XeImageFileName`); my post was still claiming that it did allocate memory itself as I thought that was the case, until I constructed the actual code example (and I forgot to change the first sentence).

> Tbh if you prefer to have the parameter gone, I'd rather move the strrchr stuff to the constructor and leave the task of implementing dirname and basename for later.

Yes, I'd prefer that.

JayFoxRox: We should document this, as most people would probably expect a mounted path instead.

I'm always concerned about the future-proofness, as I want to avoid breakage of nxdk apps if we have to change APIs.

If we add more path-related functions, we might want to make them return the same path style, too. If it doesn't make sense (it might make sense to address some paths using drive letters for example).
Maybe we should name this `XGetXbeDevicePath` to clarify this / avoid confusion in the future?

Also, would these paths also work for our libc `fopen` etc.?

thrimbor: Done.

thrimbor: Following our Discord discussion, this is now named `XGetCurrentXbeNtPath`. I added Doxygen-style documentation as well.

This path style is not compatible with winapi and libc.

samples/winapi_filefind/main.c
JayFoxRox: I think that mounting should be it's only sample in the future (not necessarily in this PR).
For the moment, I'd personally just prefer to see our samples simplified and reduced to "do one thing, but do it well".

However, we can also do that in a follow-up PR which contributes a new "mount"-sample.

lib/nxdk/mount.h
```diff
@@ -0,0 +1,26 @@
...
+{
+#endif
+
+#ifndef __cplusplus
+#include <stdbool.h>
+#endif
+
+bool XIsDriveMounted (char driveLetter);
+bool XMountDrive (char driveLetter, char *path);
+bool XUnmountDrive (char driveLetter);
```
JayFoxRox: Still torn about these prefixes: X is used in the original XDK for XGraphics, XAudio (I think?), XInput and maybe other APIs; this could lead to conflicts.
That way we don't have to mix MS and our APIs (as I'd like to implement XInput compatibility probably, but potentially also a new nxdk input layer).

Therefore I still consider another prefix.

We could consider `nx`/`nX` for **nx**dk or something different.
I don't have an ideal solution; I just want to raise concerns over the proposed `X` prefix.

LukeUsher: nx prefix works for me: It's unlikely to conflict with other libraries and it's pretty common to see such prefixes in console development kits.

A decision doesn't have to be made right now, but this shouldn't hold back this PR imo.

Perhaps it's worth leaving it as is for now, and creating an issue for a discussion, and once a decision is made, we can make all of nxdk consistent with a prefix. 

This might break some existing projects, but there aren't that many yet, better to do something like this sooner than later.

thrimbor: I'm fine with the `nx` prefix, too. I've updated the PR to use it instead.

lib/nxdk/Makefile
```diff
@@ -13,3 +13,21 @@ CLEANRULES += clean-nxdk
...
+NXDK_AUTOMOUNT_D_SRCS := \
+	$(NXDK_DIR)/lib/nxdk/automount_d.c
+
+NXDK_AUTOMOUNT_D_OBJS = $(addsuffix .obj, $(basename $(NXDK_AUTOMOUNT_D_SRCS)))
+
+$(NXDK_DIR)/lib/libnxdk_automount_d.lib: $(NXDK_AUTOMOUNT_D_OBJS)
+
+ifneq ($(NXDK_DISABLE_AUTOMOUNT_D),y)
+main.exe: $(NXDK_DIR)/lib/libnxdk_automount_d.lib
+NXDK_LDFLAGS += -include:_automount_d_drive
```
JayFoxRox: Why is this necessary?

lib/nxdk/Makefile
```diff
@@ -0,0 +1,33 @@
...
+	$(VE)rm -f $(NXDK_OBJS) $(NXDK_DIR)/lib/libnxdk.lib
+
+NXDK_AUTOMOUNT_D_SRCS := \
+	$(NXDK_DIR)/lib/nxdk/automount_d.c
+
+NXDK_AUTOMOUNT_D_OBJS = $(addsuffix .obj, $(basename $(NXDK_AUTOMOUNT_D_SRCS)))
+
+$(NXDK_DIR)/lib/libnxdk_automount_d.lib: $(NXDK_AUTOMOUNT_D_OBJS)
+
+ifneq ($(NXDK_DISABLE_AUTOMOUNT_D),y)
```
JayFoxRox: nit: We should have a sample for this; if we don't add one in this PR, we'll need to create an issue after merge (otherwise this feature will be forgotten about).

The sample (or issue) should also explain that D: might still be mounted by the kernel (from [LaunchDataPage->launch_path](https://xboxdevwiki.net/Kernel/LaunchDataPage)); the Discord search doesn't work right now - but we discussed this a while ago (at least RadWolfie, LukeUsher and myself were involved). I believe I concluded that the `launch_path` was like `[<DVD path>;]<XBE path>` - but nobody tested or documented it. I don't fully remember if it supports relative paths for DVDs / absolute paths for XBE.

JayFoxRox:
I have done multiple review passes, code seems to be okay (new API seems to have an okay design).
I have also ran the code (including `NXDK_DISABLE_AUTOMOUNT_D=y`) and everything seems to work as intended.

Merging.

@thrimbor Please create a new issue about the automount sample / mount sample.
Also please clarify the `-include:` option in the review comment above; it also seems to work for me without it?

lib/nxdk/Makefile
thrimbor: > Is this to prevent optimizing it away?

Yes. The linker only pulls in symbols that are referenced from libraries, ignoring the attribute, unfortunately. I've had issues with this before when I played around with LTO, and afaik our only option is to use linker options here.

I only noticed this issue here when I tested on my Xbox by launching from a dashboard, where `D:` isn't automatically mounted and the constructor wasn't run.

PR #231, opened 2019-11-24
wchar stubs & PDCLib update

thrimbor:
This adds all the stubs in `wchar.h` required by libc++. The PDCLib update is included, because it's required for this to work (the PDCLib update includes the typedefs in `wctype.h`).

This probably is the last preparation PR for libc++.

### DIFF
--- a/lib/xboxrt/Makefile
+++ b/lib/xboxrt/Makefile
@@ -2,6 +2,7 @@ SRCS += \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/stat.c \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/strings.c \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/wchar.c \
+	$(NXDK_DIR)/lib/xboxrt/libc_extensions/wchar_ext_.c \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/stdlib_ext_.c \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/string_ext_.c \
 	$(NXDK_DIR)/lib/xboxrt/c_runtime/_alldiv.s \
--- a/lib/xboxrt/libc_extensions/wchar.c
+++ b/lib/xboxrt/libc_extensions/wchar.c
@@ -1,4 +1,5 @@
 #include "wchar.h"
+#include <assert.h>
 
 wchar_t * wcscat( wchar_t * XBOXRT_RESTRICT s1, const wchar_t * XBOXRT_RESTRICT s2 )
 {
@@ -322,4 +323,98 @@ wchar_t * wmemmove( wchar_t * dest, const wchar_t * src, size_t n )
         }
     }
     return rv;
+}
+
+wchar_t * wmemset(wchar_t *wcs, wchar_t wc, size_t n)
+{
+    wchar_t * p = wcs;
+    while ( n-- )
+    {
+        *p++ = wc;
+    }
+    return wcs;
+}
+
+size_t wcrtomb (char *mbchar, wchar_t wchar, mbstate_t *mbstate)
+{
+    assert(0);
+    return 0;
+}
+
+size_t mbrtowc (wchar_t *wchar, const char *mbchar, size_t count, mbstate_t *mbstate)
+{
+    assert(0);
+    return 0;
+}
+
+double wcstod(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr)
+{
+    assert(0);
+    return 0.0;
+}
+
+float wcstof(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr)
+{
+    assert(0);
+    return 0.0;
+}
+
+long double wcstold(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr)
+{
+    assert(0);
+    return 0.0;
+}
+
+long wcstol(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base)
+{
+    assert(0);
+    return 0;
+}
+
+long long wcstoll(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base)
+{
+    assert(0);
+    return 0;
+}
+
+unsigned long wcstoul(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base)
+{
+    assert(0);
+    return 0;
+}
+
+unsigned long long wcstoull(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base)
+{
+    assert(0);
+    return 0;
+}
+
+size_t mbrlen(const char *s, size_t n, mbstate_t *ps)
+{
+    assert(0);
+    return 0;
+}
+
+wint_t btowc(int c)
+{
+    assert(0);
+    return 0;
+}
+
+size_t mbsrtowcs(wchar_t *dest, const char **src, size_t len, mbstate_t *ps)
+{
+    assert(0);
+    return 0;
+}
+
+int wctob(wint_t c)
+{
+    assert(0);
+    return 0;
+}
+
+int wcrtomb_s(size_t * XBOXRT_RESTRICT retval, char * XBOXRT_RESTRICT s, size_t ssz, wchar_t wc, mbstate_t * XBOXRT_RESTRICT ps)
+{
+    assert(0);
+    return 0;
 }
--- a/lib/xboxrt/libc_extensions/wchar.h
+++ b/lib/xboxrt/libc_extensions/wchar.h
@@ -28,6 +28,8 @@ typedef struct XBOXRT_mbstate {
 
 typedef XBOXRT_mbstate_t mbstate_t;
 
+struct tm;
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/lib/xboxrt/libc_extensions/wchar.h
+++ b/lib/xboxrt/libc_extensions/wchar.h
@@ -3,6 +3,7 @@
 
 #include <stdint.h>
 #include <stddef.h>
+#include <wctype.h>
 
 #ifdef __cplusplus
     #define XBOXRT_RESTRICT __restrict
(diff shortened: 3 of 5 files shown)


dracc:
I've tested this as part of the libcxx2 branch. Looks good to me. 👍

lib/xboxrt/libc_extensions/wchar.h
```diff
@@ -55,6 +58,21 @@ int mbsinit( const mbstate_t * ps );
...
+static long wcstol(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base) {assert(0); return 0.0;};
+static long long wcstoll(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base) {assert(0); return 0.0;};
+static unsigned long wcstoul(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base) {assert(0); return 0.0;};
+static unsigned long long wcstoull(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base) {assert(0); return 0.0;};
+static size_t mbrlen(const char *s, size_t n, mbstate_t *ps) {assert(0); return 0;};
+static wint_t btowc(int c) {assert(0); return 0;};
+static size_t mbsrtowcs(wchar_t *dest, const char **src, size_t len, mbstate_t *ps) {assert(0); return 0;};
+static int wctob(wint_t c) {assert(0); return 0;};
+static int _snwprintf(wchar_t *buffer, size_t count, const wchar_t *format, ...) {assert(0); return 0;};
+static int wcrtomb_s(size_t * XBOXRT_RESTRICT retval, char * XBOXRT_RESTRICT s, size_t ssz, wchar_t wc, mbstate_t * XBOXRT_RESTRICT ps) {assert(0); return 0;};
```
JayFoxRox: Why are some stubs in this header, while other stubs are in the .c file?
I'd probably prefer all stubs in the .c file, so we can avoid future back-and-forth changes in the header (and possible merge conflicts when touching nearby functions in different PRs).

lib/xboxrt/libc_extensions/wchar.h
```diff
@@ -55,6 +58,21 @@ int mbsinit( const mbstate_t * ps );
 size_t mbrtowc( wchar_t * XBOXRT_RESTRICT pwc, const char * XBOXRT_RESTRICT s, size_t n, mbstate_t * XBOXRT_RESTRICT ps );
 size_t wcrtomb( char * XBOXRT_RESTRICT s, wchar_t wc, mbstate_t * XBOXRT_RESTRICT ps );
 
+wchar_t *wmemset(wchar_t *wcs, wchar_t wc, size_t n);
```
JayFoxRox: Why is there a blank line before this?
Why did we already have prototypes for some functions, but not others?

lib/xboxrt/libc_extensions/wchar.h
thrimbor: All declarations I added had their definition included in `wchar.h` as well. I moved them over to `wchar.c` now.

For `strtof` / `strtod` there was already an upstream header, which is not the case here yet. I'm not sure whether there's upstream interest to already have that header (but without implementations), but it's a bit outside the scope of what I'm wanting to work on rn.

lib/xboxrt/libc_extensions/wchar.h
thrimbor: The blank line was left over from visually separating the new stuff back when I wrote this. It's gone now.
The previous prototypes were probably all added on an as-needed basis (which is mostly true for what I've added as well).

lib/xboxrt/libc_extensions/wchar.h
```diff
@@ -54,6 +57,20 @@ size_t wcslen( const wchar_t * s );
...
+long double wcstold(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr);
+long wcstol(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base);
+long long wcstoll(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base);
+unsigned long wcstoul(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base);
+unsigned long long wcstoull(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base);
+size_t mbrlen(const char *s, size_t n, mbstate_t *ps);
+wint_t btowc(int c);
+size_t mbsrtowcs(wchar_t *dest, const char **src, size_t len, mbstate_t *ps);
+int wctob(wint_t c);
+int _snwprintf(wchar_t *buffer, size_t count, const wchar_t *format, ...);
```
JayFoxRox: This looks like a Windows extension again. It should be separated from the rest, a comment should be added

lib/xboxrt/libc_extensions/wchar.h
JayFoxRox: Fair enough. I'm not sure if order is relevant for something (grouping of features / doxygen / ..) but we can worry about it later.

lib/xboxrt/libc_extensions/wchar.h
JayFoxRox: We should create an issue upstream and call out this discrepancy.

lib/xboxrt/libc_extensions/wchar.c
```diff
@@ -323,3 +323,103 @@ wchar_t * wmemmove( wchar_t * dest, const wchar_t * src, size_t n )
...
+unsigned long wcstoul(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base)
+{
+    assert(0);
+    return 0.0;
+}
+
+unsigned long long wcstoull(const wchar_t * XBOXRT_RESTRICT nptr, wchar_t ** XBOXRT_RESTRICT endptr, int base)
+{
+    assert(0);
+    return 0.0;
```
JayFoxRox: Returns `unsigned long long` but is stubbed with a non-integer value. Looks like a copy-paste mistake?

lib/xboxrt/libc_extensions/wchar.h
JayFoxRox: This is actually somewhat consistent with https://rootdirectory.ddns.net/dokuwiki/doku.php?id=pdclib:start as stdlib is considered "DONE" but wchar.h hasn't even been started yet.

lib/xboxrt/libc_extensions/wchar.h
JayFoxRox: Non-C features should not only be separated, but also moved to an extension header / C file instead (wchar_ext_.h / wchar_ext_.c). I didn't even realize that upstream was missing wchar.h entirely.

We should then mimic the extension hook by including the wchar_ext_.h from our own wchar.h (which, hopefully, will eventually replaced by the pdclib wchar.h).

thrimbor: I checked all declarations I added against the C standard, and as a result, moved `_snwprintf` into an extension header along with a comment.

lib/xboxrt/libc_extensions/wchar.c
thrimbor: Fixed all occurrences of this. Interesting that it didn't produce a warning before.

lib/xboxrt/libc_extensions/wchar.h
```diff
@@ -1,8 +1,11 @@
 #ifndef _XBOXRT_WCHAR_H
 #define _XBOXRT_WCHAR_H
 
+#include <assert.h>
 #include <stdint.h>
 #include <stddef.h>
+#include <time.h>
```
JayFoxRox: Why?

lib/xboxrt/libc_extensions/wchar.h
```diff
@@ -1,8 +1,11 @@
 #ifndef _XBOXRT_WCHAR_H
 #define _XBOXRT_WCHAR_H
 
+#include <assert.h>
```
JayFoxRox: Should move to source file - shouldn't be needed in header

lib/xboxrt/libc_extensions/wchar_ext_.h
```diff
@@ -0,0 +1,2 @@
+// Windows-specific extension required by libc++
+int _snwprintf(wchar_t *buffer, size_t count, const wchar_t *format, ...);
```
JayFoxRox: Wouldn't this need `#ifdef __cplusplus` ?

lib/xboxrt/libc_extensions/wchar_ext_.c
```diff
@@ -0,0 +1,7 @@
+#include <wchar.h>
+
+int _snwprintf(wchar_t *buffer, size_t count, const wchar_t *format, ...)
```
JayFoxRox: nit: `size_t` should require string.h or something? - same implict style in wchar.h

lib/xboxrt/Makefile
```diff
@@ -4,6 +4,7 @@ SRCS += \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/wchar.c \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/stdlib_ext_.c \
 	$(NXDK_DIR)/lib/xboxrt/libc_extensions/string_ext_.c \
+	$(NXDK_DIR)/lib/xboxrt/libc_extensions/wchar_ext_.c \
```
JayFoxRox: nit: I'd move this directly below wchar.c. So if wchar-related errors occur, they are always close to each other in the compilation logs

lib/xboxrt/libc_extensions/wchar.h
thrimbor: Because `wchar.h` needs to provide `tm`, which is declared in `time.h`. At the time I wrote this I thought this was the cleanest way to provide it (i.e. not having to deal with PDCLib internals here), however, reading the standard again, I realized that `wchar.h` only needs to provide an incomplete type, so I replaced this with `struct tm;`, for which I added a new commit.

lib/xboxrt/libc_extensions/wchar.h
thrimbor: Done.

lib/xboxrt/libc_extensions/wchar_ext_.h
thrimbor: Yes it does, fixed.

lib/xboxrt/libc_extensions/wchar_ext_.c
thrimbor: No, `size_t` is supposed to be provided by multiple headers, including `wchar.h`. It currently does this indirectly (and I didn't change this because I think we shouldn't touch PDCLib internals here). A proper upstream `wchar.h` will eventually resolve this.

lib/xboxrt/Makefile
thrimbor: Done.

lib/xboxrt/libc_extensions/wchar.h
JayFoxRox: I didn't check the standard, but I'll trust your expertise.
Thanks for explaining this - this is really surprising to me and wasn't obvious earlier.

nit: the new line could have mentioned that this is required by the spec in a comment. However, I feel that it is implicitly documented in `git blame` now by explicitly adding it (in its own commit).

PR #232, opened 2019-11-24
samples: Fix warnings about implicitly declaring debugPrint()

thrimbor:
The SDL samples weren't including `hal/debug.h` so the compiler warned about implicitly declaring `debugPrint`, this adds the include statement to fix that.


dracc:
Lgtm. 👍

PR #235, opened 2019-11-27
Fix some cxbe bugs

JayFoxRox:
This fixes some of the worst bugs in cxbe:

* Fixes dates, as cxbe uses `time_t` pointers to read from the XBE. `time_t` is larger than 4 bytes value on my platform, but the XBE only stored 4 byte timestamps. This meant we also read neighboring fields (and got bad timestamps).
* Fixes bug where our stubbed conversion from UCS-2 to ASCII left the string without zero-termination (causing buggy output on terminal). This is just a hotfix - we'll eventually have to fix the conversion routines and factor out common code (the code exists 2 times).
* Fixes bug where UCS-2 name was only zero-terminated in the first 20 symbols, after that, garbage could have followed, so any name above 20 symbols might have expanded to up to 40 symbols.
* Fixes bug where 256 symbols could be entered for XBE name, and it would overrun the 40 symbol space in the XBE certificate (and user got no warning).
* Changes behaviour so long names will be trimmed, rather than the XBE title-name being "Untitled". I assume we'll eventually cause hard-errors instead of warnings.
* Fixes bug where the generated output-filename for `main.exe` became `main.EXE.xbe`; now behaves as expected (`main.exe` &rarr; `main.xbe` / `main.bin` &rarr; `main.bin.xbe`).

I intend to do further refactors in cxbe afterwards (for eventually adding XDK DXT support).

### DIFF
--- a/tools/cxbe/Main.cpp
+++ b/tools/cxbe/Main.cpp
@@ -113,7 +113,7 @@ int main(int argc, char *argv[])
             }
             else if(strcmp(szOptionU, "TITLE") == 0)
             {
-                if(dwParamSize > 256)
+                if(dwParamSize > 40)
                     printf("WARNING: Title too long, using default title\n");
                 else
                     strcpy(szXbeTitle, szParam);
--- a/tools/cxbe/Main.cpp
+++ b/tools/cxbe/Main.cpp
@@ -166,15 +166,14 @@ int main(int argc, char *argv[])
 
             char *szWorking = szFilename;
 
-            strncpy(szWorkingU, szWorking, 265);
-
             for(int c=0;szFilename[c] != 0;c++)
                 if(szFilename[c] == '.')
                     szWorking = &szFilename[c];
 
-            MakeUpper(szWorking);
+            strncpy(szWorkingU, szWorking, 265);
+            MakeUpper(szWorkingU);
 
-            if(strcmp(szWorkingU, ".exe") == 0)
+            if(strcmp(szWorkingU, ".EXE") == 0)
                 strcpy(szWorking, ".xbe");
             else
                 strcat(szXbeFilename, ".xbe");
--- a/tools/cxbe/Main.cpp
+++ b/tools/cxbe/Main.cpp
@@ -55,7 +55,6 @@ int main(int argc, char *argv[])
     {
         char *szOption    = 0;
         char *szParam     = 0;
-        uint  dwParamSize = 0;
 
         // if this isn't an option, it must be the Exe file
         if(argv[v][0] != '-')
@@ -87,9 +86,6 @@ int main(int argc, char *argv[])
 
             szOption = &argv[v][1];
             szParam  = &argv[v][dwColon + 1];
-
-            while(szParam[dwParamSize] != 0)
-                dwParamSize++;
         }
 
         // interpret the current switch
@@ -113,10 +109,7 @@ int main(int argc, char *argv[])
             }
             else if(strcmp(szOptionU, "TITLE") == 0)
             {
-                if(dwParamSize > 40)
-                    printf("WARNING: Title too long, using default title\n");
-                else
-                    strcpy(szXbeTitle, szParam);
+                strcpy(szXbeTitle, szParam);
             }
             else if(strcmp(szOptionU, "MODE") == 0)
             {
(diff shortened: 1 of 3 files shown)


tools/cxbe/Main.cpp
```diff
@@ -113,10 +113,11 @@ int main(int argc, char *argv[])
             }
             else if(strcmp(szOptionU, "TITLE") == 0)
             {
-                if(dwParamSize > 256)
-                    printf("WARNING: Title too long, using default title\n");
-                else
-                    strcpy(szXbeTitle, szParam);
```
dracc: I fail to see where this strcpy ended up instead. Was it extraneous to begin with?

JayFoxRox: This was a bug; I have fixed it.

I had originally written this PR on top of a larger refactor (which touches the argument-parser); that refactored code read the argument directly into `szXbeTitle`.

I have added the `strcpy` back in.

I've also moved the trimming out of the argument-parser. This will help with the mentioned argument-parser refactor and also deal with a long default title (in case someone ever changes it).

tools/cxbe/Main.cpp
```diff
@@ -113,10 +113,7 @@ int main(int argc, char *argv[])
             }
             else if(strcmp(szOptionU, "TITLE") == 0)
             {
-                if(dwParamSize > 40)
```
JayFoxRox: Looks like this variable no longer needed after this change.
I will remove it; I'll leave a message and resolve this conversation when I've done that.

JayFoxRox: Fixed

tools/cxbe/Xbe.cpp
```diff
@@ -592,13 +592,14 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail)
...
         setlocale( LC_ALL, "English" );
-        ///wcstombs(m_szAsciiTitle, m_Certificate.wszTitleName, 40);
+        //wcstombs(m_szAsciiTitle, m_Certificate.wszTitleName, 40);
         char *c = m_szAsciiTitle;
         char *d = (char *) m_Certificate.wszTitleName;
         while (*d) {
            *c++ = *d++;
            d++;
         }
+        *c = '\0';
```
thrimbor: Same problem as above.

tools/cxbe/Xbe.cpp
```diff
@@ -131,16 +131,16 @@ Xbe::Xbe(const char *x_szFilename)
...
-
         //wcstombs(m_szAsciiTitle, m_Certificate.wszTitleName, 40);
         char *c = m_szAsciiTitle;
         char *d = (char *) m_Certificate.wszTitleName;
         while (*d) {
            *c++ = *d++;
            d++;
         }
-
+        *c = '\0';
```
thrimbor: This addition on it's own is fine, but the preceding loop is highly problematic. If the XBE contains 40 unicode characters for the title name, then there will be no zero-termination of `wszTitleName`. As the loop stubbornly tries to find a zero, it will end up looking at the first byte of the "alternate title IDs" table, which cxbe sets to zero, so it will terminate with `c` pointing at the byte after the buffer, which makes `*c = '\0';` an out of bounds write.

To fix this, the loop would need to check for the field length, too, and the size of `m_szAsciiTitle` needs to be increased to 41 bytes to include space for the zero-terminator.

JayFoxRox: I have added a hack to the original commit which *should* fix this, although I did not test it.

I tried to keep changes simple as I don't feel like adding much more code to this. There's *more* buffer overruns elsewhere and the code doesn't even do character conversion yet.

tools/cxbe/Main.cpp
```diff
@@ -140,6 +133,12 @@ int main(int argc, char *argv[])
         }
     }
 
+    if(strlen(szXbeTitle) > 40)
+    {
+        printf("WARNING: Title too long, trimming\n");
```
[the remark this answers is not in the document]
JayFoxRox: Should be done in follow-up. This PR is just preparing for refactors.

There's a lot of issues in cxbe. I think we'll eventually replace it with a simpler Python script, replace it with a new C program, or perform a major code refactor / rewrite.

tools/cxbe/Xbe.cpp
thrimbor: The hack is good enough for me.

PR #240, opened 2019-11-29
fp20compiler: Avoid overflow bug with duplicated constant-colors

### LINKED ISSUE #216: fp20compiler: Possible crash if constant colors are declared more than once
This shader will `assert(false)` (SIGABRT):
```c
!!RC1.0
const0 = (0,0,0,0);
{
  const1 = (0,0,0,0);
  const1 = (0,0,0,0);
  rgb { discard = zero; }
}
out.rgb=zero;
out.a=zero;
```

This shader will crash (SIGSEGV):
```c
!!RC1.0
const0 = (0,0,0,0);
const0 = (0,0,0,0);
{
  const1 = (0,0,0,0);
  const1 = (0,0,0,0);
  rgb { discard = zero; }
}
out.rgb=zero;
out.a=zero;
```

The issue stems from incorrect handling of colors that are declared twice (The same code also exists for the global constants):

https://github.com/XboxDev/nxdk/blob/1d0900b7de544113f41b48942d342a2f405eedd1/tools/fp20compiler/rc1.0_general.cpp#L104-L106

So even though the colors are declared twice, it keeps `numConsts` at 2.
Later, it tries to inherit locals from globals using this:

https://github.com/XboxDev/nxdk/blob/1d0900b7de544113f41b48942d342a2f405eedd1/tools/fp20compiler/rc1.0_general.cpp#L88-L99

So we have `2x const1` as local, but also inherit 1-2x `const0` from the global scope. This increments `numConsts` to 3 or 4.
However, the array to store them is only 2 elements long (1 slot for `const0`; 1 slot for `const1`):

https://github.com/XboxDev/nxdk/blob/1d0900b7de544113f41b48942d342a2f405eedd1/tools/fp20compiler/rc1.0_general.h#L82-L83

---

A solution is to just reduce `numConsts` for each constant which existed twice.
I have actually done this in https://github.com/JayFoxRox/nxdk/pull/41/commits/22595850e2175ef31caf952db3bac587438d0b57.

However, I believe it's better to move constant colors into a new struct (grammar rule) altogether. Currently, we have 2 duplicated copies of the color handling. For #197 we'd add a third, and also duplicate more code.

In the meantime, just avoid defining a color twice.

JayFoxRox:
Closes #216 

I was originally going to send another PR right now, but then I noticed that the planned PR hides the bug explained in #216 (but does not fix it).

This PR aims to fix #216 and adds more asserts to catch such errors in the future.

If a constant-color is set twice, the later one is used; the first color value is discarded.
*(Although the error message should discourage the user from setting colors twice anyway)*

### DIFF
--- a/tools/fp20compiler/rc1.0_combiners.cpp
+++ b/tools/fp20compiler/rc1.0_combiners.cpp
@@ -11,8 +11,11 @@
 void CombinersStruct::Validate()
 {
     if (2 == numConsts &&
-        cc[0].reg.bits.name == cc[1].reg.bits.name)
+        cc[0].reg.bits.name == cc[1].reg.bits.name) {
         errors.set("global constant set twice");
+        cc[0] = cc[1];
+        numConsts = 1;
+    }
 
     generals.Validate(numConsts, &cc[0]);
 
--- a/tools/fp20compiler/rc1.0_general.cpp
+++ b/tools/fp20compiler/rc1.0_general.cpp
@@ -27,8 +27,10 @@ void GeneralCombinersStruct::Validate(int numConsts, ConstColorStruct *pcc)
 
     localConsts = 0;
     int i;
-    for (i = 0; i < num; i++)
+    for (i = 0; i < num; i++) {
+        general[i].Validate(i);
         localConsts += general[i].numConsts;
+    }
 
     if (localConsts > 0)
         // if (NULL == glCombinerStageParameterfvNV)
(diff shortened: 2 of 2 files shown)


dracc:
Code changes look good to me.
I've tested compiling this branch, worked fine.
I have not tested any cases where these changes should matter as my shader writing skills are lacking at best.

dracc:
I've now tested fp20compiler with the shaders provided in #216 .
It no longer crashes with these changes in place. The compiler output looks reasonable too.

Issue #208, opened 2019-10-21
fp20compiler: Most texture-modes aren't supported yet


JayFoxRox:
nvparse supports a lot of texture modes, however, most of them have not been added to fp20compiler (despite being supported by hardware).

So far, we also do not support any texture-mode which has a dependency on a previous stage. While code for texture-dependencies was added in #195, it has never been tested with actual texture-modes which require it.

Sample code for these features (and support for those modes in fp20compiler, would be appreciated).

JayFoxRox:
There's actually a bug in the implementation that was merged by #195.
It assumes that the first argument in nvparse texture instructions is always a texture reference.

That breaks `cull_fragment` which has `(float)GL_LESS` or `(float)GL_GEQUAL` as first argument.
That isn't actually an issue, because that mode was never supported (it also has `assert(false)`, despite being handled by the switch-case).
All other modes should work, as they have a texture reference as first argument, or they don't take any arguments (which is an implicit 0, which is always a valid reference).

PR #241, opened 2019-11-30
fp20compiler: Do stricter error checking for texture-shaders

JayFoxRox:
This improves the error messages for texture-shaders:

- Line numbers are shown now. For technical-reasons, this is the line-number of the closing parentheses. This might change in the future.
- Fixes the bug described in https://github.com/XboxDev/nxdk/issues/208#issuecomment-546570695 (Issue 208 itself is unaffected).
- TS1.0 supports multi-stage instructions in the style `1_of_2` / `2_of_2`. This ensures all stages follow each other (thereby completing the instruction).
    
    The existing code in nvparse only checked if the previous stage existed, but not if the next stage existed. I'm not sure if there's any benefit of using partial instructions (there might be: some instructions output data before the last stage).
    However, due to the suffix naming I think it makes sense to force completeness. We can later turn this into a warning (we currently only support fatal errors in nvparse_errors) or introduce new instruction syntax for partial instructions (as there are more issues with TS1.0 syntax).
    
    This code uses some ugly interleaving of pre-processor / C-code. However, it's the best solution I could come up with.
- The existing code in nvparse caused an error message for invalid references, by not allowing to read from after the first stage of the current instruction.

    That is (by examples):
    - `<previous instruction>*_of_N` in `<current instruction>*_of_N`: allowed (reference of an earlier instruction)
    - `<current instruction>1_of_N` in `<current instruction>1_of_N`: not allowed (self-reference of current stage)
    - `<current instruction>2_of_N` in `<current instruction>1_of_N`: not allowed (reference to future stage)
    - `<next instruction>*_of_N` in `<current instruction>*_of_N`: not allowed (reference to future instruction)
    - *`<current instruction>1_of_N` in `<current instruction>2_of_N`: not allowed* (self-reference of earlier stage from instruction)

    As we didn't use error-checking in the past, we had an `assert` for texture-dependencies which enforced this rule less strictly: it disallows referencing anything starting with the current stage, but allows referencing within earlier stages of the current instruction.

    That is (by example):
    - Same as above, except..
    - .. *`<current instruction>1_of_N` in `<current instruction>2_of_N`: allowed* (self-reference of earlier stage from instruction)

    This assert is being removed because we have error messages now. The assert actually prevents the shader compiler from reaching error output, as the assert would trigger first.
    The assert was turned into a second error message now.
    I'm not sure how hardware behaves if these dependencies are invalid; we might want to turn one or both of the errors into warnings in the future.

---

Due to the incomplete state of texture-shader compilation, you have to disable this assert for testing:

https://github.com/XboxDev/nxdk/blob/c922276d10bd44c3a301e3e0b76fbddf01266017/tools/fp20compiler/ts1.0_inst_list.cpp#L81-L83

The generated code will be garbage, but the errors should be reported correctly.

You can find shaders for testing (and a tool to run them) in this branch: https://github.com/JayFoxRox/nxdk/pull/56
Some of those tests annotations might not mention some additional errors that are displayed by now. Just review the shaders yourself to check if the errors make sense.


dracc:
Code changes look good. Will test according to instructions before this week is gone.

JayFoxRox:
I don't think there'll be anybody with a good understanding of it.

Texture shaders was an nvidia specific extension (https://github.com/KhronosGroup/OpenGL-Registry/blob/master/extensions/NV/NV_texture_shader.txt) for OpenGL which only existed for a couple of years. They were a niche for a short period in time.

The OpenGL API exposes them rather powerful (see spec), however, nvparse (fp20compiler) abstracts this (and doesn't expose the actual instructions from the spec) into a "high-level" language called `!!TS1.0`.
So nvparse is actually very limited as it combines the micro-operations into larger `!!TS1.0` macro-operations (multiple stages like `usecase_1_of_3`, `usecase_2_of_3`, `usecase_3_of_3` as alias for 3 different micro-operations).

So by design texture_shaders (texture-shaders in OpenGL API) and `!!TS1.0` (texture-shaders in nvparse/fp20compiler) are very different.

This PR mostly checks for completeness of those multi-stage macro-operations in `!!TS1.0`. It doesn't have to do much with the hardware feature (which cares about the underlying micro-operations), but just how the `!!TS1.0` syntax is implied to be:

- You wouldn't expect `foo_1_of_2` followed by `bar_3_of_3` (even if the hardware supports the underlying micro-operations).
- You wouldn't expect `foo_1_of_2` followed by `bar_2_of_2` (even if the hardware supports the underlying micro-operations).
- You wouldn't expect `foo_2_of_2` followed by `foo_1_of_2` (even if the hardware supports the underlying micro-operations).
- You would only allow `foo_1_of_2` followed by `foo_2_of_2` (as implied by the naming).
- ...
- You wouldn't expect the user to use the output of a previous macro stage as input to a later stage (the macro-operation would have taken care of it internally).
- The output of macro stages is only implied for the last stage (`2_of_2` / `3_of_3`) of the macro, so you'd disallow referencing a stage mid-operation (`1_of_2` / `1_of_3` / `2_of_3`) as input for something else (even if the hardware supports the underlying micro-operations).

The original nvparse allows many of those things, but in practice they aren't useful due to the `!!TS1.0` language design - hence I explicitly check for such conditions to avoid typos or dirty hacks in shaders.
If we want to enable more powerful features we should expose micro-operations from `!!TS1.0` eventually (which I'm planning to do).

The only documentation about `!!TS1.0` that I'm aware of is probably this: https://www.nvidia.com/attach/6371  - and you'll see what I mean (they explain the underlying texture_shader features and how those micro-operations can be used, then introduce a macro-operation for their `!!TS1.0` language).

mborgerson:
Thanks!

PR #242, opened 2019-11-30
fp20compiler: Improve error messages for register-combiners

JayFoxRox:
There should be no changes in behaviour, but:

- Adds line numbers to error messages.
- Provides more explicit error messages.
- Refactors code so code duplication is minimized.

All of the refactoring is contained in the final-combiner commit. I hope I didn't break anything. The code should be reviewed to make sure.

*This is the PR that has been mentioned in #240:
If this PR would have come first, it somehow would have prevented the issue explained in #216 (no more assert or crash).
I'm not sure why that happened, but I assume the structs get packed differently, so that the colors actually overwrite the color count.*

---

You can find shaders for testing (and a tool to run them) in this branch: https://github.com/JayFoxRox/nxdk/pull/56
Some of those tests annotations might not mention some additional errors that are displayed by now. Just review the shaders yourself to check if the errors make sense.

### DIFF
--- a/tools/fp20compiler/rc1.0_final.cpp
+++ b/tools/fp20compiler/rc1.0_final.cpp
@@ -31,30 +31,42 @@ void FinalProductStruct::ZeroOut()
     e.Init(zero, MAP_UNSIGNED_IDENTITY);
     f.Init(zero, MAP_UNSIGNED_IDENTITY);
 }
+
+static void ValidateRgbInputRegister(RegisterEnum& reg, bool isEorF, bool hasProduct) {
+    if (RCP_NONE == reg.bits.channel)
+        reg.bits.channel = RCP_RGB;
+
+    if (REG_DISCARD == reg.bits.name)
+        errors.set("reading from discard", reg.line_number);
+    if (RCP_BLUE == reg.bits.channel)
+        errors.set("blue component can not be used in rgb portion", reg.line_number);
+    if (!isEorF) {
+        if (!hasProduct && (REG_E_TIMES_F == reg.bits.name))
+            errors.set("final_product used but not set", reg.line_number);
+    } else {
+        if (REG_E_TIMES_F == reg.bits.name)
+            errors.set("final_product can not reference itself", reg.line_number);
+        if (REG_SPARE0_PLUS_SECONDARY_COLOR == reg.bits.name)
+            errors.set("color_sum can not be used in final_product", reg.line_number);
+    }
+}
+
+static void ValidateAlphaInputRegister(RegisterEnum& reg) {
+    if (RCP_NONE == reg.bits.channel)
+        reg.bits.channel = RCP_ALPHA;
+
+    if (REG_E_TIMES_F == reg.bits.name)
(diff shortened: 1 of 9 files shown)


dracc:
I've read through the code changes. They look good to me.
I've tested compiling this branch with the `mesh` sample. Worked as expected.
I have not tested running anything where these changes should matter as my shader writing skills are lacking.

dracc:
#240 has been merged. Time to step out of the ~shadows~ draft status?

dracc:
Tested with the example shaders in #216. Works as expected with nice error output.
```
[...]
errors in shader "!!RC1.0" (line 1):
error: line 3: global constant set twice
error: line 6: local constant set twice
```
It bugs me a bit that the line numbers are shown in different ways depending on the context. OTOH that might be good for shader separation.

tools/fp20compiler/rc1.0_final.cpp
```diff
@@ -76,39 +88,20 @@ void FinalCombinerStruct::Validate()
...
-        alpha.g.reg.bits.channel = RCP_ALPHA;
+    if (REG_SPARE0_PLUS_SECONDARY_COLOR == rgb.a.reg.bits.name)
+        errors.set("color_sum can not be used in this term of the final-combiner rgb portion", rgb.a.reg.line_number);
+
+    ValidateRgbInputRegister(rgb.a.reg, false, hasProduct);
+    ValidateRgbInputRegister(rgb.b.reg, false, hasProduct);
+    ValidateRgbInputRegister(rgb.c.reg, false, hasProduct);
+    ValidateRgbInputRegister(rgb.d.reg, false, hasProduct);
+    if (hasProduct)
+    {
```
JayFoxRox: nit: Should be on same line as previous `if` I think; if not, then the newly added functions above are inconsistent.

tools/fp20compiler/rc1.0_final.cpp
```diff
@@ -76,39 +88,20 @@ void FinalCombinerStruct::Validate()
...
-    if (RCP_NONE == rgb.b.reg.bits.channel)
-        rgb.b.reg.bits.channel = RCP_RGB;
-    if (RCP_NONE == rgb.c.reg.bits.channel)
-        rgb.c.reg.bits.channel = RCP_RGB;
-    if (RCP_NONE == rgb.d.reg.bits.channel)
-        rgb.d.reg.bits.channel = RCP_RGB;
-    if (RCP_NONE == product.e.reg.bits.channel)
-        product.e.reg.bits.channel = RCP_RGB;
-    if (RCP_NONE == product.f.reg.bits.channel)
-        product.f.reg.bits.channel = RCP_RGB;
```
JayFoxRox: E and F channel initialization was previously unconditional, it is now conditional (as the call to `ValidateRgbInputRegister` which does it is conditional).
So, depending on the emitter it might cause issues later.

The condition should be moved into the function (or removed? Is it even necessary?).

thrimbor:
I reviewed the code as thorough as I could and compared the output of https://github.com/JayFoxRox/nxdk/pull/56/files from before and after these changes. The error reporting seemed to work fine and there were no regressions visible. Merging.

PR #244, opened 2019-12-04
lwip: Update submodule to STABLE-2_1_2_RELEASE

thrimbor:
This pretty much does what the title says - instead of using some random commit before the 2.0.0 release, this updates our lwip submodule to the latest stable release. Some changes were necessary for this, and I'll leave review comments with the details.

This PR intentionally doesn't touch any areas where we could optimize or clean up things or utilize features of newer releases - I focused only on updating the submodule without introducing regressions.

I tested this with our two http server samples on my 1.0 Xbox.

### DIFF
--- a/lib/net/nforceif/include/lwipopts.h
+++ b/lib/net/nforceif/include/lwipopts.h
@@ -43,7 +43,9 @@
 #define LWIP_IPV4 1
 #define LWIP_IPV6 1
 #define LWIP_DEBUG 1
-extern int errno;
+#define LWIP_ERRNO_INCLUDE <errno.h>
+#define LWIP_COMPAT_MUTEX_ALLOWED
+#define SYS_LIGHTWEIGHT_PROT 0
 
 /*
    -----------------------------------------------
--- a/lib/net/nforceif/src/driver.c
+++ b/lib/net/nforceif/src/driver.c
@@ -117,7 +117,7 @@ low_level_init(struct netif *netif)
   if (netif->mld_mac_filter != NULL) {
     ip6_addr_t ip6_allnodes_ll;
     ip6_addr_set_allnodes_linklocal(&ip6_allnodes_ll);
-    netif->mld_mac_filter(netif, &ip6_allnodes_ll, MLD6_ADD_MAC_FILTER);
+    netif->mld_mac_filter(netif, &ip6_allnodes_ll, NETIF_ADD_MAC_FILTER);
   }
 #endif /* LWIP_IPV6 && LWIP_IPV6_MLD */
 
--- a/lib/net/nforceif/src/sys_arch.c
+++ b/lib/net/nforceif/src/sys_arch.c
@@ -198,6 +198,12 @@ sys_mbox_trypost(struct sys_mbox **mb, void *msg)
   return ERR_OK;
 }
 
+err_t
+sys_mbox_trypost_fromisr(struct sys_mbox **mb, void *msg)
+{
+  return sys_mbox_trypost(mb, msg);
+}
+
 void
 sys_mbox_post(struct sys_mbox **mb, void *msg)
 {
--- a/samples/httpd/main.c
+++ b/samples/httpd/main.c
@@ -4,7 +4,7 @@
 #include "lwip/netif.h"
 #include "lwip/sys.h"
 #include "lwip/tcpip.h"
-#include "lwip/timers.h"
+#include "lwip/timeouts.h"
 #include "netif/etharp.h"
 #include "pktdrv.h"
 #include <hal/input.h>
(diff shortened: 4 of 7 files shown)


lib/net/nforceif/include/lwipopts.h
```diff
@@ -43,7 +43,9 @@
 #define LWIP_IPV4 1
 #define LWIP_IPV6 1
 #define LWIP_DEBUG 1
-extern int errno;
+#define LWIP_ERRNO_INCLUDE <errno.h>
+#define LWIP_COMPAT_MUTEX_ALLOWED
+#define SYS_LIGHTWEIGHT_PROT 0
```
thrimbor: http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=6e6ce4fb78a28244c3d6581ce75d3c7391d58150 made `SYS_LIGHTWEIGHT_PROT` enabled by default. This does require additional glue code however, so I disabled it. While this matches the behavior we've had so far, we should probably hook this up eventually, since it seems to be used to make functions like `netif_poll` thread-safe.

lib/net/nforceif/include/lwipopts.h
```diff
@@ -43,7 +43,9 @@
 #define LWIP_IPV4 1
 #define LWIP_IPV6 1
 #define LWIP_DEBUG 1
-extern int errno;
+#define LWIP_ERRNO_INCLUDE <errno.h>
+#define LWIP_COMPAT_MUTEX_ALLOWED
```
thrimbor: http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=42dfa71f978296edd79040641234d9496f1439a7 introduced a change that makes the build error out when the platform does not provide glue code for using mutexes. This should definitely get resolved by adding such code in the future, but for now, this keeps the current behavior.
Also see the comment below.

lib/net/nforceif/src/driver.c
```diff
@@ -117,7 +117,7 @@ low_level_init(struct netif *netif)
   if (netif->mld_mac_filter != NULL) {
     ip6_addr_t ip6_allnodes_ll;
     ip6_addr_set_allnodes_linklocal(&ip6_allnodes_ll);
-    netif->mld_mac_filter(netif, &ip6_allnodes_ll, MLD6_ADD_MAC_FILTER);
+    netif->mld_mac_filter(netif, &ip6_allnodes_ll, NETIF_ADD_MAC_FILTER);
```
thrimbor: http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=6b1e1af3d19c28f26519460531f28b23d4f6e1f0 combined the filter actions for IGMP and MLD6. This adapts the code to use the new constant name, no behavioral change should result from this.

lib/net/nforceif/src/sys_arch.c
```diff
@@ -198,6 +198,12 @@ sys_mbox_trypost(struct sys_mbox **mb, void *msg)
   return ERR_OK;
 }
 
+err_t
+sys_mbox_trypost_fromisr(struct sys_mbox **mb, void *msg)
+{
+  return sys_mbox_trypost(mb, msg);
+}
+
```
thrimbor: http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=8fc20142f7b1b18f24fc48187d390187d718ce30 introduced the requirement to provide this function to address an issue that seems specific to FreeRTOS. I didn't spot any actual use of this / the function calling this, but the linker still demands it, so this just wraps `sys_mbox_trypost`.

samples/httpd/main.c
```diff
@@ -4,7 +4,7 @@
 #include "lwip/netif.h"
 #include "lwip/sys.h"
 #include "lwip/tcpip.h"
-#include "lwip/timers.h"
+#include "lwip/timeouts.h"
```
thrimbor: http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=5d4c1432c2b223f2e5a2511416a2cc7fdff3a61c renamed this header to remove a naming clash on FreeRTOS.

samples/httpd/main.c
```diff
@@ -87,7 +87,7 @@ int main(void)
 
 #if USE_DHCP
 	debugPrint("Waiting for DHCP...\n");
-	while (g_pnetif->dhcp->state != DHCP_STATE_BOUND)
+	while (dhcp_supplied_address(g_pnetif) == 0)
```
thrimbor: http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=1db9631e7d5dfd0a566f188d0f17fc017bad5af9 changed the API for this. Instead of having to touch the inner details of the DHCP implementation, there is a new function called `dhcp_supplied_address`, which is supposed to return 1 when an address was retrieved via DHCP. The commit I linked caused the breakage, the new API was introduced in a different one.

(http://git.savannah.nongnu.org/cgit/lwip.git/commit/?id=dc7340bbd5209d6fc2a6df5c23d9cadc0d4244c8 is also related to this)

Ryzee119:
I built both http samples and ran successfully on my 1.6.
High level code reviews looks good to me.

mborgerson:
Great, this was long over due

lib/net/nforceif/include/lwipopts.h
JayFoxRox: Consider creating an issue about this, so we don't forget.

PR #246, opened 2019-12-05
Use GitHub actions for CI

thrimbor:
This continues #168. Changes are small, mostly just renaming the cloning step. I've decided against splitting into multiple workflows, because GH's UI doesn't group them then (makes it much more annoying to navigate them imho).

I also added a CI badge - it obviously won't show correctly yet, because it shows the status of the master branch, and the master branch of this repo doesn't have a workflow yet. But here's mine so you can see what it looks like:


README.md
```diff
@@ -6,6 +6,10 @@ Notable features:
 - No complicated cross-compiling or big library dependencies! Builds with `make` and just needs standard tools and llvm.
 - 3D graphics! nxdk includes and expands on pbkit, a library for interfacing with the Xbox GPU.
 
+Build Status
+------
+![](https://github.com/XboxDev/nxdk/workflows/Build%20Samples/badge.svg?branch=master)
```
JayFoxRox: I think https://github.community/t5/GitHub-Actions/Badge-is-always-showing-quot-failing-quot/m-p/34807#M2019 implies that we should filter by event, to not display PRs?

JayFoxRox: Please add a description in`[]`; `[CI Status for "Build Samples"]` or something

thrimbor: I believe filtering by branch should already be sufficient, but I added the event filter anyway.

thrimbor: Done.

GXTX:
I briefly chatted with @JayFoxRox about the fact that the MSYS2 environment is automatically 'upgraded' to the latest base packages by the install script supplied by chocolatey, it may be good for the case of consistency to have up-to-date libs and packages on all OSs. However I expect such a change would add a _considerable_ amount of time to the build runners.

thrimbor:
> I briefly chatted with @JayFoxRox about the fact that the MSYS2 environment is automatically 'upgraded' to the latest base packages by the install script supplied by chocolatey, it may be good for the case of consistency to have up-to-date libs and packages on all OSs. However I expect such a change would add a _considerable_ amount of time to the build runners.

I agree with the idea, but I tested this and unfortunately a full upgrade is just not practical on some platforms - the total time spent on Ubuntu increased by ~510% due to large complicated upgrades for mono. On macOS, only a single (non-related) package was upgraded, so no significant impact there - but also no effect, because the stuff we're touching is either part of the system (i.e. we can't upgrade it) or installed after the upgrade.

However, I don't think we even need to do a full upgrade - we only depend on very few packages, most of what Ubuntu has installed doesn't affect us. `apt-get install` will upgrade all packages we pass as parameters (if they're installed and not up to date ofc), so as long as we list all our dependencies excplicitly, we should be fine. Doing this deserves it's own PR though imho, we'd need to carefully check which packages we (indirectly) depend on. It would be great if you could create an issue about it though.

DiscoStarslayer:
@thrimbor smoke testing against latest packages is generally a good idea, but something that should be done in a controlled manner. Our CI shouldn't magically update dependencies, if we re-run the same job multiple times we should expect the same result, otherwise we risk merging PR's which validate as a branch and then fail once merged to master. 

I'd say the chocolatey upgrade is actually the bad practice.

I think the ideal is to produce a docker image for each environment separately with all the pre-requisites installed. Then we can version these images as release artifacts and ensure this CI only uses a specific release version. This way we guarantee jobs even a year+ old still re-run fine and also control when we decide to bump up the versioning of our build environment packages explicitly. With this flow upgrading the build environment requires a PR meaning it needs explicit sign-off which I think is advantageous. 

That said I think this PR is a huge improvement, no more mixing CI vendors and the integration is clean. My above suggestion is way out of scope of this PR.

mborgerson:
Thanks @GXTX and @thrimbor

PR #250, opened 2019-12-09
cxbe: Refactor argument parser

JayFoxRox:
This moves the argument parser to Common.cpp and makes it more re-usable.
Intended to ease development of cdxt (a tool similar to cxbe); cdxt will re-use the frontend that is being created here.

Does some minor refactors first.


tools/cxbe/Error.cpp
```diff
@@ -54,9 +55,9 @@ bool Error::ClearError()
 void Error::SetError(const char *x_szError, bool x_bFatal)
 {
     if(m_szError == 0)
-        m_szError = new char[256];
+        m_szError = new char[ERROR_LEN+1];
```
JayFoxRox: nit: I just realized that I should initialize `m_szError[ERROR_LEN] = '\0'`; otherwise a long `x_szError` will not be zero-terminated in `strncpy` below (as it's not initialized by `new char[..]`).

Other instances of such `strncpy` might exist.
I don't consider this a blocking issue as the problem had already existed anyway.

tools/cxbe/Exe.cpp
```diff
@@ -45,11 +45,11 @@ Exe::Exe(const char *x_szFilename)
...
 
-    // verify Exe file was opened
-    if(ExeFile == 0)
+    // verify Exe file was opened successfully
+    if(ExeFile == NULL)
     {
-        SetError("Could not open Exe file.", true);
-        return;
+        SetError("Could not open Exe file", true);
+        goto cleanup;
```
GXTX: nit: Using `goto`'s.

JayFoxRox: I assume this is aimed at the whole "do not use `goto`" debate? Most of that `goto`-criticsm is very old and was meant to advocate the use of structured code (`while` and `for`) which were being introduced around that time.
`goto` is still useful for situations like this. - the readability would only get worse if code was encapsulated in tiny helper functions, or even in a `while` loop using `break` statements.

C++ also introduced exceptions to avoid `goto`, but this codebase does not use exceptions. In fact, a lot of organizations do not use exceptions for different reasons.

This codebase already uses `goto` in many places.
The error that is being fixed here was caused by erroneous use of `return` - all other instances in this function use `goto` (which close the open file descriptor, and print the error message).

If you think we should avoid `goto` in this repository, I'd suggest to open an issue about it.
Personally, I think it's fine to use it in cases like this.

---

However, there is actually a bug with this, because the `cleanup` does unconditionally close the file descriptor. I'll fix this by adding a check so it's not closed if it's already `NULL`.

tools/cxbe/Error.cpp
JayFoxRox: This should be fixed now.

dracc:
Tested compiling and using for putting together the mesh sample. Works as expected. :+1:

tools/cxbe/Xbe.cpp
```diff
@@ -66,7 +66,13 @@ Xbe::Xbe(const char *x_szFilename)
...
         printf("Xbe::Xbe: Storing Xbe Path...");
 
-        strcpy(m_szPath, x_szFilename);
+        if(strlen(x_szFilename) > sizeof(m_szPath)-1)
+        {
+            SetError("Xbe Path too long", true);
+            goto cleanup;
+        }
+
+        strncpy(m_szPath, x_szFilename, sizeof(m_szPath)-1);
```
thrimbor: If `strlen(x_szFilename) == sizeof(m_szPath)-1)` then this might leave `m_szPath` without a null-terminator.

tools/cxbe/Common.cpp
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        strncat(szNewPath, szNewExtension, freeLength);
+    }
+
+    return 0;
+}
+
+// case-insensitive string compare
+bool CompareString(const char *szA, const char *szB)
```
thrimbor: There's a POSIX function doing the same thing, `strcasecmp`. Idk whether it's available on all platforms we target.

JayFoxRox: I'm aware.

I believe it's not available. [MS traditionally uses `stricmp` instead](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stricmp-wcsicmp-mbsicmp-stricmp-l-wcsicmp-l-mbsicmp-l?view=vs-2019).
So we'd probably need some preprocessor stuff which I wanted to avoid here.

thrimbor: I kinda suspected that much. I'm fine with having a custom function then.

tools/cxbe/Xbe.cpp
JayFoxRox: I added a new commit at the beginning which removes this variable; it was unused and also uninitialized in other constructors.
If anyone wants to reintroduce it eventually we'd probably just use `strdup` instead.

thrimbor:
Final round of testing went well, and review was taken care of. Merging

PR #254, opened 2019-12-19
build: Fix typo breaking make clean

thrimbor:
During the refactoring of #230 I accidentally introduced a typo in one of the cleaning rules, which broke `make clean`. This fixes the typo in the rule name.


JayFoxRox:
Oops! My mistake.

We should probably be adding `make clean` to CI (possibly even making sure that it removed *all* auto-generated files); this could have been avoidable.

PR #255, opened 2019-12-19
libc++

thrimbor:
This adds libc++ (along with a few more required stubs). The libc++ branch for nxdk currently lives under https://github.com/thrimbor/libcxx.git, and should obviously get moved before merge.

libc++ brings quite a lot of functionality, I don't expect everything to be bug-free and functional already. Simple stuff (`std::string`, `std::vector`) should work already though, and make it much easier to port or write new stuff in a safer way. I do plan to add C++11 thread support soon, I already have a branch where I implemented it.

For now, libc++ has to be enabled via `NXDK_CXX` (similar to how we handle SDL atm). libc++ supports a special pragma that causes the linker to automatically link it, but I disabled it for now because our build system can't track that dependency (the ELF-variant of lld supports emitting a Make-style dependency file, but the PE-variant doesn't).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,3 +15,7 @@
 [submodule "lib/sdl/SDL_ttf"]
 	path = lib/sdl/SDL_ttf
 	url = https://github.com/SDL-mirror/SDL_ttf
+[submodule "lib/libcxx"]
+	path = lib/libcxx
+	url = https://github.com/XboxDev/nxdk-libcxx.git
+	branch = nxdk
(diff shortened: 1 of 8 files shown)


.gitmodules
```diff
@@ -15,3 +15,6 @@
 [submodule "lib/sdl/SDL_ttf"]
 	path = lib/sdl/SDL_ttf
 	url = https://github.com/SDL-mirror/SDL_ttf
+[submodule "lib/libcxx"]
+	path = lib/libcxx
+	url = https://github.com/thrimbor/nxdk-libcxx.git
```
JayFoxRox: This should move to XboxDev first (ideally without the "threads" branch). The nxdk branch should be made the default then.

mborgerson: Agree. Any update here?

thrimbor: Done.

mborgerson:
Great stuff. Thanks!

Issue #20, opened 2017-07-11
Audio API discussion


JayFoxRox:
*Consider this a room for discussion as we were starting to derail #13 with audio talk.*
*(Feel free to close this once the discussion has ended)*

Currently, nxdk supports XAudio which was originally from OpenXDK.

However, we don't want that. Trust me.

You'll want to use the APU for audio *and not just ACI (AC97)*. I'm currently [documenting the APU VP](http://xboxdevwiki.net/APU) and [adding emulation to XQEMU](https://github.com/JayFoxRox/xqemu/pull/24).

XAudio uses AC97 only. The ACI / AC97 is very limited: You allocate a buffer, you put 16 bit signed PCM audio in it and tell the AC97 to play it back at 48kHz. You can't modify the data format, you can't control the samplearate and the CPU has to do lots of work.

The APU on the other hand, manages up to 256 audio voices and you can control the pitch and the data format (8 bit, 16 bit, 24 bit and ADPCM compressed audio). It not only supports buffered audio, but also streamed audio, hardware-audio filters, envelopes etc. - it outputs that audio as (I think) 24 bit, always at 48kHz, 32 samples grouped at a time - no CPU intervention.
The audio is then routed to the GP which is a programmable DSP (See #12 ) and then forwarded to the EP (also programmable DSP) by programming a FIFO. The EP then has a FIFO to a single page of memory which is used by the ACI / AC97 as playback source.
This way there is no CPU intervention anywhere + we gain a ton of features and avoid stuff like software-resampling or filters. We only need some code for the GP and the EP. If we don't want dolby digital, we can just pass-through the EP. If we want dolby digital we'd have to write our own AC3 encoder or dump the Microsoft / Dolby one (which is easy, but we couldn't legaly include it with nxdk).

Writing a pass-through GP / EP driver is easy.
Programming the VP takes some work, but is considerably easier than writing a software mixer and more CPU friendly than using an existing software mixer.

---

@thrimbor suggested an OpenAL-implementation. To me, writing these kind of libs is not a priority to be honest, but be my guest.
The truth is that most users of nxdk don't intend to write / port much homebrew.
Focus is very much on having something to write smaller utilities or unit tests for the emulators.
However, I personally chose to embrace [nxdk-rdt](https://github.com/mborgerson/nxdk-rdt) which is a homebrew app compiled using nxdk. It's similar to the XDK xbdm.dll concept and allows remote control of the Xbox.
[...]

PR #257, opened 2019-12-27
Simple nxdk audio sample

mborgerson:
We should have *some* demo to show how to output basic audio with nxdk. If a ~~better~~ more portable/conventional API (e.g. SDL2) is available in the future, ideally one actually providing a convenient interface to the APU, then we can replace this. Until then, this sample provides a nice and simple means to get some audio playback.

Audio comes courtesy my dog, Murdock. CC0


samples/xaudio/Makefile
```diff
@@ -0,0 +1,12 @@
+XBE_TITLE=hello
```
JayFoxRox: Wrong title

samples/xaudio/main.c
```diff
@@ -0,0 +1,79 @@
...
+    current_buf = (current_buf+1) % NUM_BUFFERS;
+    callback_count++;
+}
+
+int main(void)
+{
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+
+    for (int i = 0; i < NUM_BUFFERS; i++) {
+        buffers[i] = MmAllocateContiguousMemoryEx(buffer_size,0,0x3ffb000,0,0x404);
```
JayFoxRox: Use `PAGE_READWRITE | PAGE_WRITECOMBINE` (or something?) instead of the 0x404.

JayFoxRox: nit: We use something like `#define MAX_RAM 0x3ffb000` for another sample.

samples/xaudio/main.c
```diff
@@ -0,0 +1,79 @@
...
+unsigned char *voice_data;
+unsigned int voice_len;
+unsigned int voice_pos;
+
+#define NUM_BUFFERS 2
+const unsigned short buffer_size = 48*1024;
+unsigned char *buffers[NUM_BUFFERS];
+unsigned int current_buf;
+
+static void provide_samples_callback(void *pac97Device, void *data)
```
JayFoxRox: How is this called? I could imagine that this doesn't allow float as it is an interrupt (do DPCs allow float?)

samples/xaudio/main.c
mborgerson: All of our samples allocate memory this way, but I agree we shouldn't use these constants

mborgerson: Same for your other comment for this line above

samples/xaudio/Makefile
mborgerson: Copy-paste bug, will fix

samples/xaudio/main.c
mborgerson: It's called from a DPC. No floats are used here, but if someone uses floats here they will need to save and restore the FPU state. I added a comment to warn people.

samples/xaudio/Makefile
mborgerson: Done

samples/xaudio/main.c
mborgerson: Done

mborgerson: Done with a FIXME to move this to another file. There's no common spot for it yet. We should refactor all samples that use this constant similarly.

JayFoxRox:
For consistency, please use prefix style for commit title (`samples: ` in this case).

I have skimmed over the code earlier. I'll probably try running the sample and also listen to the audio file before giving my final okay.

mborgerson:
> For consistency, please use prefix style for commit title (`samples: ` in this case).

Ok

JayFoxRox:
I have skimmed over the code and tested the sample.

Merged.

---

Some final notes about things which I would like to see improved in the future:

- nxdk.wav is a bad filename, woof.wav, dog.wav, bark.wav or something would have been more suitable.
- Another audio file which is more recognizable would have been nice, ideally human voice or a simple melody (to show pitch correctness, volume and playback speed).
- Direct loading of the WAV would have been nice, but I can also understand why it wasn't done.
- The callback `data` argument is a bit confusing. It could be documented that it comes from the callback setter.
- The callback ~`device` parameter is a bit confusing, and it being `NULL` in the initial samples might lead to bugs. There could have been a second function without arguments to avoid this problem.
- The `debugPrint` cursor position could have been controlled, so the counter could be updated more often (as it currently skips a couple of numbers).
- As already suggested in the PR description, we should try to get rid of this API, so we should replace it with an SDL audio or OpenAL-like (see #20) sample instead.
- "Dead code" in Makefile for format conversion could have had a comment.
- Buffer size in `unsigned short` is very tight.
- Buffer size follows no logical formula (is `48*1024` but could have been `12288 * NUM_CHANNELS * sizeof(int16_t) // ~250ms of audio (12288 samples at 48kHz)`).

mborgerson:
> nxdk.wav is a bad filename, woof.wav, dog.wav, bark.wav or something would have been more suitable.

I'm not sure that it is a "bad" filename. It's a placeholder file, so nxdk.wav is acceptable in my opinion. People are expected to change it.

> Another audio file which is more recognizable would have been nice, ideally human voice or a simple melody (to show pitch correctness, volume and playback speed).

That would be fine, as long as the file remains relatively small. If you do this, please find or create something that is short and CC0 for redistribution.

> Direct loading of the WAV would have been nice, but I can also understand why it wasn't done.

I'm not sure that's necessary for this small sample, but would like to see someone add WAV loader. I consider that to be outside the scope of this PR.

> The callback data argument is a bit confusing. It could be documented that it comes from the callback setter. The callback ~device parameter is a bit confusing, and it being NULL in the initial samples might lead to bugs. There could have been a second function without arguments to avoid this problem.

The prototype for the callback is copy-pasted from the xaudio API. It's good to stick with the names used by the API as I have done here. A follow-up PR could rename the parameters throughout the code base if you can think of better names.

Consumers of this particular sample code should be expected to understand the nature of a callback function and its typical arguments like application-supplied pointer and device pointer, and refer to the xaudio API if they have questions (which could be better documented, to be fair).

> The debugPrint cursor position could have been controlled, so the counter could be updated more often (as it currently skips a couple of numbers).

That's not a bad idea, might be easier to read too. Come to think of it, I did exactly that for my timer sample.

> "Dead code" in Makefile for format conversion could have had a comment.

I think it's pretty clear, but it wouldn't hurt I suppose.

> Buffer size in unsigned short is very tight.

Buffer size is defined by the API as an `unsigned short`, that's why it's used here. Using a wider type could lead to bugs if the API is not updated. Though, what is passed into the ac97 descriptor is an unsigned short of number of samples so this could actually support higher buffer sizes, but we need to change the API. I've left that for future work. Also, the 48 KiB buffer is plenty in this case--one can get away with much, much smaller buffers, at the cost of more frequent interrupts.

> Buffer size follows no logical formula (is 48*1024 but could have been 12288 * NUM_CHANNELS * sizeof(int16_t) // ~250ms of audio (12288 samples at 48kHz)).

Buffer size was selected to be "sufficiently large." If you want to PR a formula to calculate buffer size based on duration of audio content, that would be nice to have.

PR #258, opened 2019-12-27
build: Find nxdk includes last

JayFoxRox:
This changes all includes to `-isystem` so that they are searched later.

See https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html for the search order.

---

I'm not sure about this change myself:

- This treats headers like this https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html which might not be what we want (we want to improve our headers probably).
- It is probably incomplete (SDL include paths are added elsewhere?).
- #85 (by @thrimbor) did add an `-isystem` include but also an `-I` include, so there appear to be reasons for differentiating?

The motivation is to fix an issue in Neverball. Neverball has a file called `share/video.h` which is included by just using `#include "video.h"`; nxdk master will find `hal/video.h` first instead, which leads to undeclared functions, which then breaks behaviour at runtime.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,12 +51,13 @@ EXTRACT_XISO = $(NXDK_DIR)/tools/extract-xiso/build/extract-xiso
 TOOLS        = cxbe vp20compiler fp20compiler extract-xiso
 NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
                -ffreestanding -nostdlib -fno-builtin -fno-exceptions \
-               -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt/libc_extensions \
-               -I$(NXDK_DIR)/lib/hal \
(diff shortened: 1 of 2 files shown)


JayFoxRox:
CI fails as we'll have to update the submodules to include the proper path for debug.h:

```
$ grep -R \<debug\.h\> | grep incl
sdl/SDL2/src/video/xbox/SDL_xbframebuffer.c:#include <debug.h>
pdclib/platform/xbox/functions/assert/assert.c:#include <debug.h>
```

JayFoxRox:
Closing this PR as it's probably not what we want (we want warnings for our headers) and it also doesn't work on Windows for some reason (keeps failing CI; maybe bug in Clang/LLVM?).

A better solution would probably be to remove the `hal/` include directory and to re-order our parameters on the command line.

PR #259, opened 2019-12-28
ci: Utilize all cores for building

thrimbor:
Simple addition to our CI build script to let `make` use parallel jobs (where the number of jobs equals the number of cores). Roughly halves the build time (not the total job time, though).


JayFoxRox:
Fine with me; if we are lucky this will also help to catch some dependency errors in our Makefiles.

PR #260, opened 2019-12-28
ci: Use cached msys2 environment

thrimbor:
Pretty much the same thing @mborgerson set up for xqemu, I copied the config changes from him. The msys2 environment was adjusted to contain the packages nxdk needs, and while it's much larger than the xqemu version, it still reduces the time required a lot while also working around chocolatey reliability issues.

### DIFF
--- a/.github/workflows/build_samples.yml
+++ b/.github/workflows/build_samples.yml
@@ -13,13 +13,17 @@ jobs:
           submodules: recursive
       - name: Install MSYS2 & Dependencies
         run: |
-          choco install msys2 --no-progress
-          C:\tools\msys64\usr\bin\bash.exe -lc "pacman --needed --noconfirm -S make bison flex mingw-w64-x86_64-llvm mingw-w64-x86_64-clang mingw-w64-x86_64-lld cmake"
+          echo "Downloading MSYS2 environment..."
+          Invoke-WebRequest -Uri "https://github.com/XboxDev/nxdk-ci-environment-msys2/releases/latest/download/msys64.7z" -OutFile "msys64.7z"
+          echo "Extracting MSYS2 environment..."
+          7z x -y msys64.7z "-oC:\"
+          echo "Updating MSYS2 environment..."
+          C:\msys64\usr\bin\bash.exe -lc "pacman -Syu --noconfirm"
       - name: Build
         env:
           MSYS2_ARCH: x86_64
           MSYSTEM: MINGW64
-        run: C:\tools\msys64\usr\bin\bash.exe -lc "cd $(cygpath $env:GITHUB_WORKSPACE) && ./.ci_build_samples.sh"
+        run: C:\msys64\usr\bin\bash.exe -lc "cd $(cygpath $env:GITHUB_WORKSPACE) && ./.ci_build_samples.sh"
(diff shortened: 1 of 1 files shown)


.github/workflows/build_samples.yml
```diff
@@ -13,13 +13,17 @@ jobs:
           submodules: recursive
       - name: Install MSYS2 & Dependencies
         run: |
-          choco install msys2 --no-progress
-          C:\tools\msys64\usr\bin\bash.exe -lc "pacman --needed --noconfirm -S make bison flex mingw-w64-x86_64-llvm mingw-w64-x86_64-clang mingw-w64-x86_64-lld cmake"
+          echo "Downloading MSYS2 environment..."
+          Invoke-WebRequest -Uri "https://github.com/XboxDev/ci-environment-msys2/releases/latest/download/msys64.7z" -OutFile "msys64.7z"
```
JayFoxRox: I'm not sure about using one CI environment for all projects. I feel like an nxdk specific environment might be more suitable. This could be either in a branch, or in a well-named repository.

- If we have multiple environments it probably makes sense to have a base-environment so the only per-project change is the pacman line.
- If we keep one CI environment for all projects we should at least clearly document which packages are installed for each user, so we don't accidentally remove them in the future.

*Ideally this whole msys package issue wouldn't even exist, and GitHub Actions would be smarter about it; caching msys2, or at the very least chocolatey requests (I assume we aren't the only project with this issue). We should probably report such trouble on the Actions forums.*

mborgerson: > I'm not sure about using one CI environment for all projects. I feel like an nxdk specific environment might be more suitable

It would be beneficial to keep things simple for now with a single toolchain environment. nxdk is the only user at this point, if we want to add other users down tho road maybe we can come back to this if it becomes a bottleneck.

> If we keep one CI environment for all projects we should at least clearly document which packages are installed for each user

That's a good idea. Let's update the environment workflow file when we add additional packages.

> Ideally this whole msys package issue wouldn't even exist, and GitHub Actions would be smarter about it; caching msys2, or at the very least chocolatey requests

It would be neat if GitHub supported VM snapshots or something. On Ubuntu/macOS a significant amount of time is spent just installing deps. We will see what they come up with I guess.

mborgerson:
Approved. @JayFoxRox can merge when he sees my feedback above, or I will in the next day or so.

.github/workflows/build_samples.yml
JayFoxRox: Just to be clear: I was implying we should rename the repository to `nxdk-ci-...` if we only use it for nxdk, or we should change the environment file rather sooner than later to avoid any trouble (ideally before a merge, so it's always documented).

thrimbor: I'd do the renaming, but after moving the repo to XboxDev I don't have the permissions necessary.

mborgerson: That name sounds good to me. @JayFoxRox if you want to rename it and merge

JayFoxRox: I renamed the repository (and disabled some unused features), then force-pushed a the URL update to this branch.

PR #261, opened 2019-12-29
Xboxkrnl additions

thrimbor:
I'm trying to clean up my branches and noticed that I didn't upstream those changes yet. They're only documentation changes and added NTSTATUS-constants (I believe they come from my work on exceptions), so no behavioral changes are expected.


lib/xboxkrnl/xboxkrnl.h
```diff
@@ -4001,6 +4023,12 @@ XBAPI VOID DECLSPEC_NORETURN NTAPI HalReturnToFirmware
     IN FIRMWARE_REENTRY Routine
 );
 
+/**
+ * Registers or unregisters a shutdown notification which contains a function
+ * pointer that will be called on shutdown.
```
JayFoxRox: nit: This is the only description comment which ends in a period.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -3725,6 +3743,10 @@ XBAPI NTSTATUS NTAPI KeDelayExecutionThread
     IN PLARGE_INTEGER Interval
 );
 
+/**
+ * Connects an interrupt object, allowing it to receive interrupts
+ * @return FALSE if the interrupt is already connected or cannot be connected, TRUE if it was conected successfully.
```
JayFoxRox: nit: This is the only parameter comment which ends in a period.

lib/xboxkrnl/xboxkrnl.h
thrimbor: Actually I seem to have used periods a few times in the past, but the non-period variant is more common. Changed.

lib/xboxkrnl/xboxkrnl.h
thrimbor: Same as above, changed.

PR #264, opened 2020-01-02
xboxkrnl: Remove "volatile" keyword from declarations not requiring it

### LINKED ISSUE #263: Variables exported from xboxkrnl are marked `volatile`
This was discussed in https://github.com/XboxDev/nxdk/pull/229#discussion_r351749172

We currently get some warnings which means the `volatile` will quickly propagate into other parts of the codebase if the user respects them. Removing those `volatile` keywords would prevent that, at the risk of breaking user code which depends on kernel modifications to variables. The `volatile` might also disallow certain optimizations, even if the kernel variables are constant (or can be assumed as such) - so this issue also affects performance.

However, we aren't sure how the LLVM compiler/linker would behave when removing `volatile` (is it implied for library imports possibly?).

thrimbor:
Since #229 was merged, I've done a bit more reading on the volatile keyword and how it might impact our use-case (such as [this](https://docs.microsoft.com/en-us/cpp/cpp/volatile-cpp?redirectedfrom=MSDN&view=vs-2019), [this](https://docs.microsoft.com/en-us/cpp/cpp/const-and-volatile-pointers?view=vs-2019) and a couple of more random stuff).

Since we're using exported kernel symbols, clang cannot assume that pointers or values retrieved from the kernel or passed to kernel functions won't change during kernel calls. This is not explicitly documented somewhere, but not having this guarantee would completely break DLLs (or even normal linking), so I think this is safe to assume. That means we only have to worry about kernel data that could change during execution of XBE code, such as variables touched in IRQ handlers.

First of all, we have a couple of kernel exports that won't change during runtime (of the XBE), such as `HalDiskSerialNumber` etc. or the `*ObjectType` variables, which are only used to pass pointers to them to kernel functions. These obviously don't require `volatile`.

`KeTickCount` gets to keep its `volatile`. While I did successfully test the samples using it without `volatile` (it's mainly required in the kernel to stop compilers from optimizing away the seemingly unnecessary write), we want to make sure its content isn't cached when it's read to prevent timing issues, especially since its content is going to be updated in IRQ handlers, which could interrupt our code at any point. Since we provide a proper abstraction (`GetTickCount()`) and users will probably use their own local variables to store the retrieved value, I don't expect this to have any measurable performance impact.

The same goes for `KeSystemTime` and `KeInterruptTime` (though we don't provide abstractions yet afaik) - like with `KeTickCount`, we could probably remove `volatile` here and never experience any issues, but it doesn't hurt to be safe here.

TL;DR: I don't think we really require `volatile` anywhere, but I'm keeping it for `KeTickCount`, `KeSystemTime` and `KeInterruptTime` to be safe because they're timing sensitive, and dropping it from the rest.

Fixes #263.

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -2399,7 +2399,7 @@ XBAPI VOID NTAPI READ_PORT_BUFFER_UCHAR
     IN ULONG Count
 );
 
-XBAPI volatile OBJECT_TYPE PsThreadObjectType;
+XBAPI OBJECT_TYPE PsThreadObjectType;
 
 XBAPI VOID NTAPI __attribute__ ((noreturn)) PsTerminateSystemThread
 (
@@ -2462,8 +2462,8 @@ XBAPI DWORD NTAPI PhyGetLinkState
     BOOLEAN update
 );
 
-XBAPI volatile OBJECT_HANDLE_TABLE ObpObjectHandleTable;
-XBAPI volatile OBJECT_TYPE ObSymbolicLinkObjectType;
+XBAPI OBJECT_HANDLE_TABLE ObpObjectHandleTable;
+XBAPI OBJECT_TYPE ObSymbolicLinkObjectType;
 
 XBAPI NTSTATUS NTAPI ObReferenceObjectByPointer
 (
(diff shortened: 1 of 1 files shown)


dracc:
Looks good to me! :+1:
I haven't written any tests for these changes in particular but NevolutionX still runs like before.

JayFoxRox:
Merging untested.

Willing to take a risk as we have too many PRs piling up.
If there are any problems, create a new issue please.

PR #265, opened 2020-01-03
Gamma sample

JayFoxRox:
This is something I've played around with recently; as also described in the README, this isn't exactly useful on Xbox, but might still be useful in niche situations or for "free" effects (fades, color filters, flashes, ..).

As this is adding to the hal/video API I wasn't too focused on making this clean.
I chose a `memcpy` style API for now.

The sample has some ugly bits which will need to be improved in the future (the sample image is poorly chosen for a gamma test *and* tone-mapping). Ideally we'll be loading a couple of different images from memory.

In the XDK, this is exposed by D3D (presumably). However, it makes sense to expose this together with video-encoder features as it's not part of the rendering, but display routines.

---

I have not tested this with 16bpp modes yet. I'm not sure if the ramp is after conversion to 8-bit.
This should be checked before merge.

*(This feature is not emulated in XQEMU yet, so this must be tested on physical hardware)*

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -22,6 +22,9 @@
 #define NV_PMC_INTR_EN_0_INTA_HARDWARE 0x00000001
 #define NV_PMC_INTR_EN_0_INTA_SOFTWARE 0x00000002
 
+#define NV_USER_DAC_WRITE_MODE_ADDRESS 0x006813C8
+#define NV_USER_DAC_PALETTE_DATA 0x006813C9
+
 // Defines for video regions
 #define VIDEO_REGION_NTSCM			0x00000100
 #define VIDEO_REGION_NTSCJ			0x00000200
@@ -40,6 +43,7 @@ int			flickerLevel		= 5;
 BOOL			flickerSet		= FALSE;
 BOOL			softenFilter		= TRUE;
 BOOL			softenSet		= FALSE;
+GAMMA_RAMP_ENTRY	gammaRampEntries[256];
 
 static KINTERRUPT InterruptObject;
 static KDPC DPCObject;
@@ -294,6 +298,8 @@ VIDEO_MODE XVideoGetMode(void)
 
 void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 {
+	int i;
+	GAMMA_RAMP_ENTRY defaultGammaRampEntries[256];
 	ULONG Step = 0;
 	DWORD dwFormat = 0;
 	int bytesPerPixel = (bpp+7)/8;
@@ -325,6 +331,13 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 
 	XVideoSetFlickerFilter(5);
 	XVideoSetSoftenFilter(TRUE);
+
+	for(i = 0; i < 256; i++) {
+		defaultGammaRampEntries[i].red = i;
+		defaultGammaRampEntries[i].green = i;
+		defaultGammaRampEntries[i].blue = i;
+	}
+	XVideoSetGammaRamp(0, defaultGammaRampEntries, 256);
 
 	XVideoSetVideoEnable(TRUE);
 
@@ -384,6 +397,25 @@ void XVideoSetSoftenFilter(BOOL enable)
 
 		softenSet = TRUE;
 		softenFilter = enable;
+	}
+}
+
+void XVideoSetGammaRamp(unsigned int offset, const GAMMA_RAMP_ENTRY *entries, unsigned int count)
+{
+	int i;
+
+	assert((offset + count) <= 256);
+	memcpy(&gammaRampEntries[offset], entries, count * sizeof(GAMMA_RAMP_ENTRY));
+
+	/* Partial setting of `count` entries at `offset` would be nice, but
+	   NV2A doesn't seem to support this: Changing the write address will
+	   sometimes mess up the existing table.
+	   To avoid this, we update it all from a cached state. */
+	VIDEOREG8(NV_USER_DAC_WRITE_MODE_ADDRESS) = 0;
+	for(i = 0; i < 256; i++) {
+		VIDEOREG8(NV_USER_DAC_PALETTE_DATA) = gammaRampEntries[i].red;
+		VIDEOREG8(NV_USER_DAC_PALETTE_DATA) = gammaRampEntries[i].green;
+		VIDEOREG8(NV_USER_DAC_PALETTE_DATA) = gammaRampEntries[i].blue;
 	}
 }
 
--- a/samples/gamma/README.md
+++ b/samples/gamma/README.md
@@ -1,0 +1,11 @@
+gamma
+=====
+
+Example of using GPU gamma ramps for tonemapping and effects.
+
+By not setting a linear ramp, you are typically reducing the amount of colors that you can output to the display.
+It might still be beneficial in some situations (correcting for bad display).
+
+In practice, tonemapping is not useful on original Xbox.
+The GPU does not support higher framebuffer precision than 8-bit integers per channel.
+So when working in linear space, it will quickly lead to quantization artifacts.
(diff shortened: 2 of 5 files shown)


dracc:
Looks good to me, looking forward to testing!

lib/hal/video.c
```diff
@@ -326,6 +331,14 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 	XVideoSetFlickerFilter(5);
 	XVideoSetSoftenFilter(TRUE);
 
+	GAMMA_RAMP_ENTRY default_gamma_ramp_entries[256];
+	for(i = 0; i < 256; i++) {
```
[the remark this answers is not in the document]
JayFoxRox: Why? `var++` is way more common than `++var` for C for-loops.

Other loop in this code also uses `position++`, too.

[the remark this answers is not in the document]
dracc: According to what I've been taught preincrement is more efficient than postincrement. I'm sure a good compiler optimizes away the difference in this case, but still.

samples/gamma/README.md
```diff
@@ -0,0 +1,13 @@
+gamma
+=====
+
+Example of using GPU gamma ramps for tonemapping and effects.
+
+By not setting a linear ramp, you are typically reducing the amount of colors
+that you can output to the display.
```
mborgerson: nit: wrapping

It might be nice to include your suggestion of using this to produce fade effects, and maybe some resources here on how to go about calculating these values (or a note about the links you've provided in the sample code).

samples/gamma/main.c
```diff
@@ -0,0 +1,159 @@
...
+    nextRow = 25;
+    debugPrint("%s", title);
+    for(int i = strlen(title); i < 60; i++) {
+        debugPrint(" ");
+    }
+}
+
+int main(void)
+{
+
```
mborgerson: nit: our other samples don't have a blank line here if no local args are declared..

samples/gamma/main.c
```diff
@@ -0,0 +1,159 @@
...
+#include <windows.h>
+
+static void draw_title(const char* title)
+{
+    extern int nextCol;
+    extern int nextRow;
+    nextCol = 25;
+    nextRow = 25;
+    debugPrint("%s", title);
+    for(int i = strlen(title); i < 60; i++) {
```
mborgerson: nit: our other samples use space between keyword and open paren in control structures

lib/hal/video.c
```diff
@@ -22,6 +22,9 @@
 #define NV_PMC_INTR_EN_0_INTA_HARDWARE 0x00000001
 #define NV_PMC_INTR_EN_0_INTA_SOFTWARE 0x00000002
 
+#define NV_USER_DAC_WRITE_MODE_ADDRESS 0x006813C8
```
mborgerson: Ideally these go in some `.h` file, but ok for now.

lib/hal/video.c
```diff
@@ -387,6 +400,24 @@ void XVideoSetSoftenFilter(BOOL enable)
 	}
 }
 
+void XVideoSetGammaRamp(unsigned int offset, const GAMMA_RAMP_ENTRY *entries, unsigned int count) {
+	int i;
+
+	assert((offset + count) <= 256);
+	memcpy(&gamma_ramp_entries[offset], entries, count * sizeof(GAMMA_RAMP_ENTRY));
```
mborgerson: Since you've cached the gamma ramp entries, should we enforce people access only through a corresponding `XVideoSetGammaRamp` function?

lib/hal/video.c
```diff
@@ -40,6 +43,7 @@ int			flickerLevel		= 5;
 BOOL			flickerSet		= FALSE;
 BOOL			softenFilter		= TRUE;
 BOOL			softenSet		= FALSE;
+GAMMA_RAMP_ENTRY	gamma_ramp_entries[256];
```
mborgerson: nit: Style in file uses camel case

lib/hal/video.c
```diff
@@ -387,6 +400,24 @@ void XVideoSetSoftenFilter(BOOL enable)
 	}
 }
 
+void XVideoSetGammaRamp(unsigned int offset, const GAMMA_RAMP_ENTRY *entries, unsigned int count) {
```
mborgerson: nit: newline before brace

samples/gamma/README.md
JayFoxRox: How would you like to have this wrapped? I did manually wrap at 80 columns.

samples/gamma/main.c
JayFoxRox: This is actually inconsistent in the existing samples.
I did look at mesh, triangle and led - they all seem to mix these.

Only the new audio sample seems to use spaces consistently.

I personally prefer the non-space style, but am willing to change this if you want me to (waiting for response).

Ideally we'll take care of this with clang-format rules in the future.

lib/hal/video.c
JayFoxRox: Agreed, but this is how hal/video.c does it.

lib/hal/video.c
JayFoxRox: > should we make `gamma_ramp_entries` static

No, because this also applies to other code in hal/video.c.
We'll have to refactor this. I'll create an issue later if none exists yet.

*(It used to be static before I made style consistent in preparation for this PR)*

>  if necessary later, a corresponding `XVideoGetGammaRamp` function?

This branch used to have this function, but it was a last-minute removal before sending the PR.

I decided to remove it because there's no way to read the initial system state (there might be a way, but I don't know how yet).
So this can leave us in a situation where we are unable to use this function because the gamma ramp was never set.

Particularily, this will become a problem when introducing another video API to replace hal/video, which doesn't automatically set the state (to be able to take over rendering from another XBE; e.g.: menu.xbe &rarr; ingame.xbe where ingame.xbe should do a lightweight video initialization without resetting gamma ramps etc.)

So I'm being a bit cautious with adding an API yet.

lib/hal/video.c
JayFoxRox: Fixed.

lib/hal/video.c
JayFoxRox: I feel stupid now. Fixed.

samples/gamma/main.c
JayFoxRox: Fixed.

lib/hal/video.c
mborgerson: Fair enough

samples/gamma/README.md
mborgerson: > How would you like to have this wrapped? I did manually wrap at 80 columns.

Looks like you fixed it. The version I reviewed had a sentence wrap randomly at 36 chars.

> I think that people will figure this out themselves.

Fair enough I suppose. Would like to provide more context for newbies, but I guess we don't really do it with other samples so we shouldn't have to start now.

> I'm unable to take a screenshot due to lack of emulation in XQEMU. With this effect not being part of framebuffer I also can't easily dump it from a physical Xbox. A photo of a display isn't nice either, and a photoshopped screenshot also isn't very good.

Let's revisit once we emulate this in XQEMU properly, or if someone can grab a screenshot through a capture card that would be nice too (@ryze119 perhaps?)

> We can probably update the description with more information in the future, when we improve / split the sample (like providing an actual gamma slider for compensating for the users display). For now, this is probably "good enough".

Agreed, don't feel like bikesheding

samples/gamma/main.c
mborgerson: Hmm.. this is weird. You're right, looks like we mix styles annoyingly. Let's not worry for now and clang-format it in the future.

samples/gamma/README.md
JayFoxRox: > Looks like you fixed it. The version I reviewed had a sentence wrap randomly at 36 chars.

The old one was wrapped at 80 columns (not automatically, but manually).
I changed it because manually breaking at 80 columns would be annoying for my follow up samples (noticed while working on gamma-fade PR 268).
I just use shorter sentences now and have one sentence per line.

PR #266, opened 2020-01-04
Mounting fixes

thrimbor:
Small but crucial fixes that became an issue when using the mounting functionality from C++ code.


JayFoxRox:
Merging without test; these things look right.

Issue #169, opened 2019-09-08
`XVideoInit` never allocated the framebuffer memory that it advertises via `XVideoGetFB`


JayFoxRox:
The address returned from `XVideoGetFB` is usually never claimed by nxdk programs; so it's unallocated memory.

The address returned by `XVideoGetFB` is initially retrieved in `XVideoInit`  by asking the GPU for the current `PCRTC_START` (hardware framebuffer output) address.
I'm not sure if the previously called `AvSetDisplayMode` controls this, so `PCRTC_START` is either the address that just happend to be framebuffer by the application that had launched nxdk, or, it's `#define VIDEO_FRAMEBUFFER			0x03c00000` (as set by `AvSetDisplayMode`).

I don't think the memory allocator persists the previous framebuffer, although it would be nice if someone could check the allocation status.

So I don't think either of these addresses is allocated; meaning if you allocate memory using the kernel, you can get unlucky and your memory will overlap the framebuffer. Any drawing will lead to undefined behaviour.

A cheap fix would be to just prevent any advertisement of memory by `XVideoGetFB` until it is allocated by the running program.

PR #268, opened 2020-01-05
samples: Add gamma-fade sample

JayFoxRox:
This adds another gamma sample which shows how to abuse the gamma ramps for fade-effects.

This was written independently from #265 (before review had finished) but I tried to adapt the requested style changes.

The README.md needs someone to spell-check.

The image is as unsuitable as the one in #265. I might improve it in the future maybe.
I already have a prototype for loading BMPs with SDL (in https://github.com/JayFoxRox/nxdk/pull/66), but that has some drawbacks, too (so I'm not going to submit it).

### DIFF
--- a/samples/gamma-fade/README.md
+++ b/samples/gamma-fade/README.md
@@ -1,0 +1,39 @@
+gamma-fade
+==========
+
+Example of using GPU gamma ramps for fade effects.
+
+The benefit of this technique is that it doesn't require any memory access.
+So if your scene already has a lot of overdraw, this might be a suitable solution.
+It only costs a bit of CPU time (while setting gamma ramps for GPU).
+
+However, this independence from memory can also be a drawback.
+As it is applied during video-output, it is a global effect.
+So it will also always affect your HUD or other elements which might not be supposed to be part of the fade.
+
+Also, because it is controlled by the CPU, it requires some form of synchronization.
+You should try to do it during vertical-blanking to avoid tearing.
+
+This sample shows 2 common styles of fade effects:
+
+* Mix:
+    This linearly interpolates between a constant color and the intended color.
+
+* Cinematic:
+    Similar to how a fade looks when done in film.
+    This is closer to an add (fade-to-white) or subtract (fade-to-black).
+
+In practice, you could extend this effect by fine-tuning it per component.
+You can also set custom curves to suit your application.
+In a real application you might also want to convolve the fade with the actual gamma curve.
+
+Each style is shown with 2 interpolations:
+
+* Linear:
+    This animates the fade with constant "speed".
+
+* Smooth:
+    This controls the animation speed so it performs ease-in and ease-out.
+
+In a real application you'd likely use a custom interpolator (a custom curve).
+You'd also vary the duration of the animation.
--- a/samples/gamma-fade/main.c
+++ b/samples/gamma-fade/main.c
@@ -1,0 +1,255 @@
+#include <stdint.h>
+#include <stdbool.h>
+#include <string.h>
+#include <math.h>
+#include <hal/video.h>
+#include <hal/debug.h>
+#include <windows.h>
+
+static float remap(float in_low,  float in_high,
+                   float out_low, float out_high,
+                   float value)
+{
+                                 /* [in_low; in_high]       */
+    value -= in_low;             /* [0.0; in_high-in_low]   */
+    value /= in_high - in_low;   /* [0.0; 1.0]              */
+    value *= out_high - out_low; /* [0.0; out_high-out_low] */
+    value += out_low;            /* [out_low; out_high]     */
+    return value;
+}
+
+/* This is also known as `lerp(a, b, t)` */
+static float mix(float a, float b, float t)
+{
+    return a * (1.0f - t) + b * t;
+}
+
+/* This is also known as `mix(0.0, 1.0, x)` */
+static float linear(float x)
+{
+    return x;
+}
+
+/* This is also known as `smoothstep(0.0, 1.0, x)` */
+static float smooth(float x)
+{
+    return x * x * (3.0f - 2.0f * x);
+}
+
+static void set_mix_fade(float red, float green, float blue,
+                         float t)
+{
+    GAMMA_RAMP_ENTRY entries[256];
+    for(int i = 0; i < 256; i++) {
+        float f = i / (float)0xFF;
+
+        entries[i].red   = 0xFF * mix(f, red,   t);
+        entries[i].green = 0xFF * mix(f, green, t);
+        entries[i].blue  = 0xFF * mix(f, blue,  t);
+    }
+    XVideoSetGammaRamp(0, entries, 256);
+}
+
+static void set_cinematic_fade(float a_black_input,  float a_white_input,
+                               float a_black_output, float a_white_output,
+                               float b_black_input,  float b_white_input,
+                               float b_black_output, float b_white_output,
+                               float t)
+{
+    GAMMA_RAMP_ENTRY entries[256];
+    for(int i = 0; i < 256; i++) {
+        float f = i / (float)0xFF;
+
+        /* Interpolate the input and output levels */
+        float black_input  = mix(a_black_input,  b_black_input,  t);
+        float white_input  = mix(a_white_input,  b_white_input,  t);
+        float black_output = mix(a_black_output, b_black_output, t);
+        float white_output = mix(a_white_output, b_white_output, t);
+
+        /* Remap our brightness */
+        if (f < black_input) { f = black_input; }
+        if (f > white_input) { f = white_input; }
+        f = remap(black_input, white_input, black_output, white_output, f);
+        if (f < black_output) { f = black_output; }
+        if (f > white_output) { f = white_output; }
+
+        entries[i].red   = 0xFF * f;
+        entries[i].green = 0xFF * f;
+        entries[i].blue  = 0xFF * f;
+    }
+    XVideoSetGammaRamp(0, entries, 256);
+}
+
+static float animate_frame(DWORD ts_start, DWORD ts, float duration)
+{
+    /* Wait for the next frame */
+    XVideoWaitForVBlank();
+
+    /* Update timestamp */
+    DWORD ts_delta = ts - ts_start;
+    float t = remap(0.0f, duration, 0.0f, 1.0f, ts_delta);
+
+    /* Signal completion */
+    if (t >= 1.0f) {
+        return 1.0f;
+    }
+
+    /* Return time in range [0.0; 1.0[ */
+    return t;
+}
+
+static void do_mix_fade(float red, float green, float blue,
+                        float(*interpolator)(float), float duration)
+{
+    DWORD ts_start;
+    float t;
+
+    /* Fade to color */
+    ts_start = GetTickCount();
+    do {
+        t = animate_frame(ts_start, GetTickCount(), duration / 2.0f);
+        set_mix_fade(red, green, blue,
+                     interpolator(t));
+    } while(t < 1.0f);
+
+    /* Fade to image */
+    ts_start = GetTickCount();
+    do {
+        t = animate_frame(ts_start, GetTickCount(), duration / 2.0f);
+        set_mix_fade(red, green, blue,
+                     interpolator(1.0f - t));
+    } while(t < 1.0f);
+}
+
+static void do_cinematic_fade(float a_black_input,  float a_white_input,
+                              float a_black_output, float a_white_output,
+                              float b_black_input,  float b_white_input,
+                              float b_black_output, float b_white_output,
+                              float(*interpolator)(float), float duration)
+{
+    DWORD ts_start;
+    float t;
+
+    /* Fade to color */
+    ts_start = GetTickCount();
+    do {
+        t = animate_frame(ts_start, GetTickCount(), duration / 2.0f);
+        set_cinematic_fade(a_black_input,  a_white_input,
+                           a_black_output, a_white_output,
+                           b_black_input,  b_white_input,
+                           b_black_output, b_white_output,
+                           interpolator(t));
+    } while(t < 1.0f);
+
+    /* Fade to image */
+    ts_start = GetTickCount();
+    do {
+        t = animate_frame(ts_start, GetTickCount(), duration / 2.0f);
+        set_cinematic_fade(a_black_input,  a_white_input,
+                           a_black_output, a_white_output,
+                           b_black_input,  b_white_input,
+                           b_black_output, b_white_output,
+                           interpolator(1.0f - t));
+    } while(t < 1.0f);
+}
+
+static void draw_title(const char* title)
+{
+    extern int nextCol;
+    extern int nextRow;
+    nextCol = 25;
+    nextRow = 25;
+    debugPrint("%s", title);
+    for(int i = strlen(title); i < 60; i++) {
+        debugPrint(" ");
+    }
+}
+
+int main(void)
+{
+    float duration = 2000.0f;
+
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+
+    /* Create test image in framebuffer */
+    volatile uint32_t *rgbx = (uint32_t*)XVideoGetFB();
+    for(unsigned int y = 0; y < 480; y++) {
+        unsigned int stripe = y / (480/3);
+
+        /* Border left */
+        for(unsigned int x = 0; x < (640-512)/2; x++) {
+            *rgbx++ = 0x808080;
+        }
+        /* 512 pixels gradient */
+        for(unsigned int x = 0; x < 256; x++) {
+            *rgbx++ = x << (16 - stripe * 8);
+            *rgbx++ = x << (16 - stripe * 8);
+        }
+        /* Border right */
+        for(unsigned int x = 0; x < (640-512)/2; x++) {
+            *rgbx++ = 0x808080;
+        }
+    }
+
+    while(1) {
+
+        draw_title("Mix fade to black (Linear)");
+        do_mix_fade(0.0f, 0.0f, 0.0f,
+                    linear, duration);
+        Sleep(1000);
+
+        draw_title("Mix fade to black (Smooth)");
+        do_mix_fade(0.0f, 0.0f, 0.0f,
+                    smooth, duration);
+        Sleep(1000);
+
+        draw_title("Mix fade to white (Linear)");
+        do_mix_fade(1.0f, 1.0f, 1.0f,
+                    linear, duration);
+        Sleep(1000);
+
+        draw_title("Mix fade to white (Smooth)");
+        do_mix_fade(1.0f, 1.0f, 1.0f,
+                    smooth, duration);
+        Sleep(1000);
+
+        draw_title("Mix fade to pink (Linear)");
+        do_mix_fade(1.0f, 0.0f, 1.0f,
+                    linear, duration);
+        Sleep(1000);
+
+        draw_title("Mix fade to pink (Smooth)");
+        do_mix_fade(1.0f, 0.0f, 1.0f,
+                    smooth, duration);
+        Sleep(1000);
+
+        draw_title("Cinematic fade to black (Linear)");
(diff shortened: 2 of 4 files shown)


samples/gamma-fade/README.md
```diff
@@ -0,0 +1,39 @@
...
+You can also set custom curves to suit your application.
+In a real application you might also want to convolve the fade with the actual gamma curve.
+
+Each style is shown with 2 interpolations:
+
+* Linear:
+    This animates the fade with constant "speed".
+
+* Smooth:
+    This controls the animation speed so it behaves like a bezier curve to ease-in and ease-out.
```
JayFoxRox: Bézier

samples/gamma-fade/main.c
```diff
@@ -0,0 +1,268 @@
...
+    return x * x * (3.0f - 2.0f * x);
+}
+
+/* This is also known as `lerp(a, b, t)` */
+float mix(float a, float b, float t)
+{
+    return a * (1.0f - t) + b * t;
+}
+
+static void reset_gamma(void)
```
mborgerson: This should probably be a standard function

samples/gamma-fade/main.c
```diff
@@ -0,0 +1,268 @@
...
+    for(int i = 0; i < 256; i++) {
+        entries[i].red   = i;
+        entries[i].green = i;
+        entries[i].blue  = i;
+    }
+    XVideoSetGammaRamp(0, entries, 256);
+}
+
+static void mix_fade(float red, float green, float blue,
+                     float t)
```
mborgerson: Why the wrap? I'm not sure it helps with readability?

samples/gamma-fade/main.c
```diff
@@ -0,0 +1,268 @@
...
+    /* Wait for the next frame */
+    XVideoWaitForVBlank();
+
+    /* Update timestamp */
+    DWORD ts_delta = ts - ts_start;
+    float t = remap(0.0f, duration, 0.0f, 1.0f, ts_delta);
+
+    /* Signal completion */
+    if (t >= 1.0f) {
+      return false;
```
mborgerson: nit: indentation

samples/gamma-fade/main.c
```diff
@@ -0,0 +1,268 @@
...
+}
+
+int main(void)
+{
+    float duration = 2000.0f;
+
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+
+    /* Create test image in framebuffer */
+    uint32_t *rgbx = (uint32_t*)XVideoGetFB();
```
mborgerson: Probably want to make this `volatile`

samples/gamma-fade/main.c
```diff
@@ -0,0 +1,268 @@
+#include <stdint.h>
+#include <stdbool.h>
+#include <string.h>
+#include <math.h>
+#include <hal/video.h>
+#include <hal/debug.h>
+#include <windows.h>
+
+float remap(float in_low,  float in_high,
```
mborgerson: Why are some of these functions `static` and others are not? They are only used here, let's just make them `static`

samples/gamma-fade/main.c
JayFoxRox: I disagree. 

In a real application the user would use their existing gamma control functions.
This could have also been implemented like either of those (to avoid any discontinuity, the real parameters in the caller would be used, with only `t=0.0f`):

```c
// 0% fade to black
mix_fade(0.0f, 0.0f, 0.0f
         0.0f);
```
```c
// 0% fade of linear ramp
cinematic_fade(0.0f, 1.0f, 0.0f, 1.0f,
               0.0f, 1.0f, 0.0f, 1.0f,
               0.0f);
```

I split this because it was more self-documentation, but I can also see how users might not be able to tell what the intention was.


---

I did actually consider this when I wrote the original gamma sample, but here are some arguments:

- If the user doesn't use gamma ramps, then hal/video does this during initialization; so the user doesn't have to worry about their gamma ramps at all. This mimics other features of XVideo like soften or flicker filter.
- If the user does use gamma ramps, then they'll probably have a simple `set_gamma_correction(float gamma)` or they'll abuse it for effects and have a more complicated curve setup to convolve different things `set_gamma_corrected_fade(float gamma, float fade)`.

Providing a ~`XVideoSetGammaCorrection` is pointless and not flexible enough for many users, and we'd have 2 conflicting APIs (low level ramp entry setup, and high level scalar). Auto-convolving those 2 would be confusing and depend on FPU in code that is likely to be ran from DPCs (when handling the gamma ramp update from pbkit in a GPU callback... when we implement those).

Providing ~`XVideoResetGammaRamps` also might lead to accidental calls which break continuity over an animation. This happens if the users gamma function is not close to the linear identity ramp (for example, if their curve generator does primarily do a fade, but also applies gamma correction).

Keeping the XVideo API non-ambigious and lower-level was the best choice (and also mimics how D3D does it).

samples/gamma-fade/main.c
JayFoxRox: This mimics the `cinematic_fade` function so the signatures and callers have the same pattern:

```c
void function([argumentGroup1,] // Fade parameters
              [argumentGroup2,] // Fade parameters
              [interpolator,] float t) // Animation control
```

*(I did have to separate the groups in the `cinematic_fade` function signature but do not separate them in the caller, to avoid vertical code, so I can see that this is not ideal either)*

samples/gamma-fade/main.c
JayFoxRox: Why? This is also how it's done in the other gamma sample. The framebuffer memory is just standard RAM and the compiler shouldn't be able to skip memory writes (at least not until I hammer the same address with the same value).

samples/gamma-fade/main.c
JayFoxRox: Bug, should be fixed.

samples/gamma-fade/main.c
JayFoxRox: I refactored the code to get rid of this (by ensuring that the function runs for `t=1.0`).

This also avoids an issue where the fade might only have been 99% complete (thereby never reaching full opaqueness). I had previously used `reset_gamma` for the 1% case after the fade, but not for the 99% case between fade-to-color and fade-to-image.

samples/gamma-fade/main.c
JayFoxRox: Fixed.

JayFoxRox:
- I addressed all feedback.
- I changed the animation loops to perform at least one run at `t = 1.0f` thereby guaranteeing that the fade completes (uses `do while` loops which I don't really like either); this is related to the `reset_gamma` discussion above.
- I re-arranged the code so `remap` and `mix` are near the top, I assume we'll eventually move those to some common sample code or use code from GLM or something.
- I prefixed the fade functions which set the gamma ramp with `set_` to hint that they changes which stick.
- Changed ranges from `[a,b]` to be `[a; b]` in comments.

samples/gamma-fade/main.c
mborgerson: I understand the pattern, but disagree that it has readability benefits, however I don't feel strongly about it

samples/gamma-fade/main.c
mborgerson: > Why? This is also how it's done in the other gamma sample.

I should have mentioned it when reviewing the other sample.

> The framebuffer memory is just standard RAM and the compiler shouldn't be able to skip memory writes (at least not until I hammer the same address with the same value)

Compilers can (and do) skip memory writes. Although we don't currently suffer this problem, in the future a sufficiently aggressive compiler optimization pass may chose to eliminate what it believes are dead writes because it lacks the knowledge that this memory is being sampled by the video encoder.

As an example, I'll turn on link-time optimization:
```c
int x[10];

int *get_fb()
{
    return x;
}

void draw()
{
    int *y = get_fb(); // a
    y[0] = 10;
    y[1] = 100;
}

int main(int argc, char *argv[])
{
    draw();
    return 0x1337;
}
```
``` bash
$ gcc -o example -O3 -flto example.c
$ objdump -dMintel example | grep -A4 main\>:
00000000000004f0 <main>:
 4f0:	b8 37 13 00 00       	mov    eax,0x1337
 4f5:	c3                   	ret    
 4f6:	66 2e 0f 1f 84 00 00 	nop    WORD PTR cs:[rax+rax*1+0x0]
 4fd:	00 00 00 
```

Changing `(a)` to be `volatile int *y = get_fb();`, we get instead what we expected:

```bash
$ objdump -dMintel example | grep -A6 main\>:
00000000000004f0 <main>:
 4f0:	c7 05 46 0b 20 00 0a 	mov    DWORD PTR [rip+0x200b46],0xa        # 201040 <x>
 4f7:	00 00 00 
 4fa:	b8 37 13 00 00       	mov    eax,0x1337
 4ff:	c7 05 3b 0b 20 00 64 	mov    DWORD PTR [rip+0x200b3b],0x64        # 201044 <x+0x4>
 506:	00 00 00 
 509:	c3                   	ret
```

JayFoxRox: This isn't actually what's happening in this code though. The optimization is only allowed because the compiler can check if `x[0..9]` is accessed elsewhere (it isn't).

In our use-case, the compiler doesn't know what memory we are accessing (it's not managed by the compiler), so it can't check if anything uses it, and it won't make any assumptions.

So, for simulating what we do here, replace `int x[10];` by `int* x = (int*)0x100;`:

```
$ objdump -dMintel example | grep -A6 main\>
0000000000001020 <main>:
    1020:	48 b8 0a 00 00 00 64 	movabs rax,0x640000000a
    1027:	00 00 00 
    102a:	48 89 04 25 00 01 00 	mov    QWORD PTR ds:0x100,rax
    1031:	00 
    1032:	b8 37 13 00 00       	mov    eax,0x1337
    1037:	c3                   	ret    
```

*(Interestingly `int m[10]; int* x = m;` is also enough to break this optimization - the write still happens)*

I'd have to do more reading, but I'm confident that this should not require `volatile`. If it did, it would break thousands of applications which map device memory into the host. [D3D](https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dlocked-rect), [SDL2](https://hg.libsdl.org/SDL/file/default/include/SDL_surface.h#l76) or [GL](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glMapBuffer.xml) don't use `volatile` in these situations either.

I do see the validity of your comment though. It *does* get interesting when adding a loop around those writes.

- `int x = 5; while(x--)` will only lead to a single write (which is somewhat expected and wanted), however, when adding some call or other action into the loop, it will do the writes again (I tested with `usleep(100)`).
- `while(1)` will lead to what I believe is a bug/oversight in gcc. Because gcc will then skip the writes - which breaks the assumption I (and also those other APIs) just did. However, clang actually keeps a single write before entering the endless loop.

> We should enforce this by also updating XVideoGetFB in a follow-up PR to return a volatile pointer.

I think we should be careful with `volatile` pointers in APIs, and expect the user to make logical decisions when accessing memory. `volatile` would still cause a lot of unintended sideffects (performance + incomplete cache flush), because the framebuffer will probably be write-combined in the future.

At the moment the discussion is also severely affected by an orthogonal issue, because the buffer is not even owned (#169). That is much worse than anything a lack of `volatile` could cause, and it will also require refactors in application code probably.

Most users will (or should) probably just use pbkit or SDL (or the ugly hal/debug.c) to access the framebuffer anyway.

---

For this PR, I feel like we should keep it consistent with the other gamma sample.

We should potentially have a follow-up to add the `volatile` or to replace such accesses by SDL (which can be assumed to be well-defined API which does not require `volatile`). Using SDL would also allow support for 16-bit modes out-of-the-box. I just wanted to avoid the additional dependency.

The pattern shown by the gamma samples should also be improved.

samples/gamma-fade/main.c
JayFoxRox: I also don't feel strongly; but I'd like to keep it the way it is.

samples/gamma-fade/main.c
mborgerson: > This isn't actually what's happening in this code though. [...]

I think this is a fair analog to what could happen in the future, even if it's not happening now with a pointer. In the past I've had compilers silently eliminate MMIO writes. While it may work now, I wouldn't be surprised to see this break in the future if (when) we apply whole-program optimization.

> If it did, it would break thousands of applications which map device memory into the host. D3D, SDL2 or GL don't use volatile in these situations either.

This is an interesting point, and I'd like to review the C memory rules more carefully--I believe the reason it works in the cases you've mentioned is due to these calls being made into shared libraries, the compiler obviously cannot optimize these writes away because this code is not loaded until runtime.

> I think we should be careful with volatile pointers in APIs, and expect the user to make logical decisions when accessing memory. volatile would still cause a lot of unintended sideffects (performance + incomplete cache flush), because the framebuffer will probably be write-combined in the future (or otherwise cached; volatile does not take care of flushes).

I agree, we *should* be careful. I'm not sure what specific side effects you're referring to in this case. The compiler is not responsible for flushing the cache--the processor itself (or OS) will handle that. The memory allocated for the frame buffer *should* be configured as write-combining, which will combine writes made by software, but the software itself (at least *this* software, OS memory management/context switching is different) doesn't handle flushing the cache.

> At the moment the discussion is also severely affected by an orthogonal issue, because the buffer is not even owned (#169). That is much worse than anything that a lack of volatile could cause, and it will also require refactors in application code probably.

I don't think this will require significant refactors to be honest. But you are correct, we should allocate the memory during `XVideoInit` (with appropriate flags) and maintain a reference to it.

As we are talking about changing to use allocated memory, here's another sample of dead write elimination that I found interesting:

```c
#include <stdlib.h>
#include <string.h>

int *get_fb()
{
    return malloc(8);
}

void draw()
{
    int *fb = get_fb();
    memset(fb, 0, 8);
}

int main(int argc, char **argv[])
{
    draw();
    return 0x1337;
}
```
```
00000000000004f0 <main>:
 4f0:	b8 37 13 00 00       	mov    eax,0x1337
 4f5:	c3                   	ret    
```



> Most users will (or should) probably just use pbkit or SDL (or the ugly hal/debug.c) to access the framebuffer anyway.

Possibly? Possibly not. This is the API we have today, and we should improve it or get rid of it.

> For this PR, I feel like we should keep it consistent with the other gamma sample.

I disagree. I think we should fix it here now, and go back and fix it in the other sample as well. This sample may work today with the compiler and settings we are using, but I don't expect it to work in the future. It's a simple change to guarantee that the compiler preserves these writes to the framebuffer and we should do it.

mborgerson: > This isn't actually what's happening in this code though. [...]

I think this is a fair analog to what could happen in the future, even if it's not happening now with a pointer. In the past I've had compilers silently eliminate MMIO writes (this will typically happen if you write to a location twice, the compiler just writes once with the second value--but if I recall correctly, I've also had it outright just drop a single write). While it may work now, I wouldn't be surprised to see this break in the future if (when) we apply whole-program optimization.

> If it did, it would break thousands of applications which map device memory into the host. D3D, SDL2 or GL don't use volatile in these situations either.

This is an interesting point, and I'd like to review the C memory rules more carefully--I believe the reason it works in the cases you've mentioned is due to these calls being made into shared libraries, the compiler obviously cannot optimize these writes away because this code is not loaded until runtime. I also think if memory is mapped into the host address space, code should be very cautious about compilers optimizing away writes to shared buffers, which may also include use of the `volatile` keyword. Frankly, I need to review the C spec to better understand the semantics of these dead-write elimination optimizations.

> I think we should be careful with volatile pointers in APIs, and expect the user to make logical decisions when accessing memory. volatile would still cause a lot of unintended sideffects (performance + incomplete cache flush), because the framebuffer will probably be write-combined in the future (or otherwise cached; volatile does not take care of flushes).

I agree, we *should* be careful. I'm not sure what specific side effects you're referring to in this case. The compiler is not responsible for flushing the cache--the processor itself (or OS) will handle that. The caching policy for the memory allocated for the frame buffer *should* be configured as write-combining, which will combine writes made by software, but the software itself (at least *this* software, OS memory management/context switching is different) doesn't typically handle flushing the cache, although there certainly may be some times when we probably want our "driver level" software to do that.

> At the moment the discussion is also severely affected by an orthogonal issue, because the buffer is not even owned (#169). That is much worse than anything that a lack of volatile could cause, and it will also require refactors in application code probably.

I don't think this will require significant refactors to be honest. But you are correct, we should allocate the memory during `XVideoInit` (with appropriate flags)--or something along those lines--and maintain a reference to it.

As we are talking about changing to use allocated memory, here's another sample of dead write elimination that I found interesting:

```c
#include <stdlib.h>
#include <string.h>

int *get_fb()
{
    return malloc(8);
}

void draw()
{
    int *fb = get_fb();
    memset(fb, 0, 8);
}

int main(int argc, char *argv[])
{
    draw();
    return 0x1337;
}
```
```
00000000000004f0 <main>:
 4f0:	b8 37 13 00 00       	mov    eax,0x1337
 4f5:	c3                   	ret    
```

Here you can see I have switched it over to using a pointer, and the compiler realized that these writes are not useful and rightly eliminated the allocation and write. You can add sufficient complexity around here to get the compiler to not do this, but eventually as compilers get even better, I think such elimination will still be done. Perhaps surprisingly in this case, *even using `volatile`* doesn't fix this elimination (the compiler does rightfully complain when passing this pointer to `memset` discards the `volatile` keyword; it ignores the `volatile` keyword in this case and respects the non-volatile definition and then proceeds to eliminate the writes).

If I change the code to not use `memset`, but just dereference and assign to the memory, then `volatile` works as expected, and the `volatile` semantics are preserved:

```c
#include <stdlib.h>
#include <string.h>

int *get_fb()
{
    return malloc(8);
}

void draw()
{
    int *fb = get_fb();
    *fb = 20;
}

int main(int argc, char *argv[])
{
    draw();
    return 0x1337;
}
```

Without `volatile`:
```
0000000000000540 <main>:
 540:	b8 37 13 00 00       	mov    eax,0x1337
 545:	c3                   	ret    
```

and with:
```
0000000000000540 <main>:
 540:	48 83 ec 08          	sub    rsp,0x8
 544:	bf 08 00 00 00       	mov    edi,0x8
 549:	e8 d2 ff ff ff       	call   520 <malloc@plt>
 54e:	c7 00 14 00 00 00    	mov    DWORD PTR [rax],0x14
 554:	b8 37 13 00 00       	mov    eax,0x1337
 559:	48 83 c4 08          	add    rsp,0x8
 55d:	c3                   	ret
```

> Most users will (or should) probably just use pbkit or SDL (or the ugly hal/debug.c) to access the framebuffer anyway.

Possibly? Possibly not. This is the API we have today, and we should improve it or get rid of it.

> For this PR, I feel like we should keep it consistent with the other gamma sample.

I disagree. I think we should fix it here now, and go back and fix it in the other sample as well. This sample may work today with the compiler and settings we are using, but I don't expect it to work in the future. It's a simple change to guarantee that the compiler preserves these writes to the framebuffer and we *should* do it.

JayFoxRox: I added a new commit to make the gamma sample accessor `volatile` and also made it `volatile` here.

samples/gamma-fade/main.c
```diff
@@ -0,0 +1,255 @@
...
+    /* Update timestamp */
+    DWORD ts_delta = ts - ts_start;
+    float t = remap(0.0f, duration, 0.0f, 1.0f, ts_delta);
+
+    /* Signal completion */
+    if (t >= 1.0f) {
+        return 1.0f;
+    }
+
+    /* Return time in range [0.0; 1.0[ */
```
mborgerson: Typo here

samples/gamma-fade/main.c
mborgerson: Thanks

samples/gamma-fade/main.c
JayFoxRox: I don't see it?

In case you mean the closing "[": The range has 1.0 excluded (it was handled above). [See this reference](https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints ).

mborgerson: Oh I see what you mean here. In this case I would have used: `[0, 1)` but it appears `[0, 1[` is also some convention..

mborgerson:
Cool, thanks

Issue #271, opened 2020-01-05
pbkit does not fully "initialize" the pipeline path in `pb_init`


JayFoxRox:
*This is an extension of #270 (which was about conditional parameters) and an actual problem (not just a documentation issue). This issue is about configuring which parameters get used (path through the graphics pipeline).*

There might be features that `pb_init` never even initializes to "on" or "off", that will affect which parameters get used.
So not only are the conditional parameters poorly configured (if you were to turn on a feature), but also wether the feature is turned on after `pb_init` is still "random" (dependent on the previous application).

So which hardware features are used in the graphics pipeline after `pb_init` is not well-defined.

I propose that we safely initialize all conditional features to "on" or "off", including the parameters they use in the chosen configuration (affects #270).
However, parameters that won't be used in the default configuration do not have to be initialized. Any higher level API with state-tracking should simply have it's own initialization routine to take control over the parameters. Setting unused parameters in pbkit would only cost memory and setup time, for very little gain.

PR #272, opened 2020-01-05
Improve `pb_init`

### LINKED ISSUE #41: 'triangle' sample gives GPU Error at current state

JayFoxRox:
Closes #41.

See https://github.com/XboxDev/nxdk/issues/41#issuecomment-570888513 for my theory as to why that is.

This was debugged using https://github.com/JayFoxRox/nxdk/pull/65.

There are actually a handful of registers which should probably be set, but aren't. I only focused on those which "scared" me the most. So this is merely a quick-fix for the worst offenders. I have written  #271 so we can do a more extensive refactor in the future.

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -3288,6 +3288,7 @@ int pb_init(void)
     p=pb_push1(p,NV097_SET_FLAT_SHADE_OP,1); //FIRST_VTX
     p=pb_push4f(p,NV097_SET_EYE_POSITION,0.0f,0.0f,0.0f,1.0f);
     p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_EDGE_FLAG,1);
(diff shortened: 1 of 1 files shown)


dracc:
Will these changes break `triangle` harder than it already is?

PR #274, opened 2020-01-06
C++11 thread requirements

thrimbor:
This implements all the winapi functionality required to enable the C++11 thread support in libc++: The InitOnce* API, Slim Reader/Writer Locks, and condition variables (I intended to split them up, but it just caused merge conflicts everywhere).

The support for condition variables is basically equivalent to how I implemented it for PDCLib, but it's modified for the lock types winapi provides. One crucial difference to the way this works on Windows, is that Windows uses keyed events to implement condition variables. We don't have that feature in the Xbox kernel, so instead, all functions use the InitOnce* API for initialization of normal events, and an additional function (which doesn't exist on Windows) is provided, `DestroyConditionVariable`, which frees the events. My thread-support patch for libc++ hooks that function up when building on nxdk.

I've written some test code for SRW locks [here](https://gist.github.com/thrimbor/7c4fe2207f5354e0966d3dcfb5df1a35), and some test code for the InitOnce* API [here](https://gist.github.com/thrimbor/7d8515ad443edafc400fc046e414329e).


dracc:
Looks good to me! :+1:

<details>
<summary>I ran the supplied tests successfully.</summary>

srwTest
initOnce
</details>

sam-itt:
Built and ran the two test cases: Gives the expected output on both xqemu and real hardware.
Note: The lock/unlock sequence of SRW locks sample is exactly the same on xqemu and my real xbox.

thrimbor:
Just noticed that one of the code changes was accidentally spread over two commits. I fixed it with a rebase, the actual code didn't change.

lib/winapi/sync.c
```diff
@@ -38,6 +40,135 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+{
+    assert(fPending != NULL);
+
+    if (dwFlags & INIT_ONCE_CHECK_ONLY) {
+        if (dwFlags & INIT_ONCE_ASYNC) {
+            *fPending = FALSE;
+            SetLastError(ERROR_INVALID_PARAMETER);
+            return FALSE;
+        }
+        if (((ULONG_PTR)lpInitOnce->Ptr & 3) != 3) {
```
JayFoxRox: Where do these pointer checks come from? They aren't documented in the MSDN for example.

lib/winapi/sync.c
```diff
@@ -40,6 +40,106 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
     RtlLeaveCriticalSection(lpCriticalSection);
 }
 
+void AcquireSRWLockExclusive (PSRWLOCK SRWLock)
+{
+    // Grab the global lock (second-highest bit)
+    uintptr_t unlocked_val = SRWLock->Ptr & 0xBFFFFFFF;
```
JayFoxRox: This does the opposite to what the comment says? It "grabs" (poor word choice) all bits, except the second highest bit.

nit: I'd prefer `~0x40000000` or `~(1 << 30)`

lib/winapi/sync.c
```diff
@@ -40,6 +40,106 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+
+    while (!__sync_bool_compare_and_swap(&SRWLock->Ptr, unlocked_val, unlocked_val | 0x40000000)) {
```
JayFoxRox: nit: You actually do use `0x40000000` here instead of `~0xBFFFFFFF` - so it's harder to see the relation.

lib/winapi/winbase.h
```diff
@@ -45,6 +45,16 @@ typedef struct _INIT_ONCE
 #define INIT_ONCE_INIT_FAILED 0x00000004UL
 typedef BOOL CALLBACK PINIT_ONCE_FN(PINIT_ONCE, PVOID, PVOID *);
 
+typedef struct  _SRWLOCK
+{
+    DWORD_PTR Ptr;
+    // highest bit: reader lock (0=free, 1=locked)
+    // second-highest bit: global lock (0=free, 1=locked)
+    // remainder: blocking readers
```
JayFoxRox: nit: Consider defining masks for this in the file where this is implemented, something like `_SRWLOCK_READER_LOCKED`, `_SRWLOCK_GLOBAL_LOCK`, `_SRWLOCK_READERS`

I had a lot of trouble reading the code with all the numeric `AND` masks - I wouldn't be surprised if there are bugs as a result.

lib/winapi/synchapi.h
```diff
@@ -19,6 +20,15 @@ VOID EnterCriticalSection (LPCRITICAL_SECTION lpCriticalSection);
 BOOL TryEnterCriticalSection (LPCRITICAL_SECTION lpCriticalSection);
 VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection);
 
+VOID InitializeConditionVariable (PCONDITION_VARIABLE ConditionVariable);
+BOOL SleepConditionVariableCS (PCONDITION_VARIABLE ConditionVariable, PCRITICAL_SECTION CriticalSection, DWORD dwMilliseconds);
+BOOL SleepConditionVariableSRW (PCONDITION_VARIABLE ConditionVariable, PSRWLOCK SRWLock, DWORD dwMilliseconds, ULONG Flags);
+VOID WakeConditionVariable (PCONDITION_VARIABLE ConditionVariable);
+VOID WakeAllConditionVariable (PCONDITION_VARIABLE ConditionVariable);
+// DestroyConditionVariable is an Xbox extension to free associated system resources
```
JayFoxRox: nit: This should make clear that it's an nxdk extension. "Xbox extension" could mean xboxkrnl.exe extension or XDK extension (or something along those lines).

lib/winapi/sync.c
thrimbor: They are, but implicitly:
> [...] If INIT_ONCE_CHECK_ONLY is specified and initialization has completed, the return value is TRUE.
> Otherwise, the return value is FALSE. To get extended error information, call GetLastError.

The `INIT_ONCE_CTX_RESERVED_BITS` lower bits of the pointer are used to keep track of state. Initialization hasn't completed yet if not both bits are set.

lib/winapi/sync.c
thrimbor: The comment refers to all of the following lines, which grab the lock by trying to set the bit.

lib/winapi/winbase.h
thrimbor: Did that. I even found a bug where I copied some piece of code while forgetting to adjust the mask.

lib/winapi/synchapi.h
thrimbor: Changed it.

lib/winapi/sync.c
```diff
@@ -38,6 +174,240 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
     RtlLeaveCriticalSection(lpCriticalSection);
 }
 
+#define SRW_GLOBAL_LOCK_BIT 0x40000000
+#define SRW_READER_LOCK_BIT 0x80000000
+#define SRW_GLOBAL_LOCK_MASK (~SRW_GLOBAL_LOCK_BIT) // 0xBFFFFFFF
+#define SRW_READER_LOCK_MASK (~SRW_READER_LOCK_BIT) // 0x7FFFFFFF
+#define SRW_READER_COUNT_MASK (SRW_GLOBAL_LOCK_MASK & SRW_READER_LOCK_MASK) // 0x3FFFFFFF
```
JayFoxRox: The `_MASK` are very weird / confusing:
- `LOCK_MASK` describe the bits that are *not* the `LOCK_BIT`.
- `COUNT_MASK` is describing the bits that *are* the `COUNT_BITS` (if it existed).

I'd suggest to only have the `_BIT` and use the well recognized `& ~FOO_BIT` and `| FOO_BIT`.
So `SRW_READER_COUNT_MASK` would also just be `SRW_READER_COUNT_BITS`. 

*(Wether you use `_MASK` or `_BIT` doesn't matter to me. My point is to have 1 of them and negate in the code where you use it)*

I feel the current way of having both is confusing because of inconsistency mentioned above; and also because you use 2 names for the same thing.

lib/winapi/sync.c
JayFoxRox: Consider defining names (in the source file) for these 4 states and using a mask created from `INIT_ONCE_CTX_RESERVED_BITS` (or manually document that your masks and that value are related).

This will help making the code more self documenting, too (instead of having to describe the current state in a comment as currently done).

thrimbor: Done.

lib/winapi/sync.c
thrimbor: Alright, changed.

dracc:
Tested 3aae926 both on XQEMU and on real hardware. Still behaves as expected. :+1:

lib/winapi/sync.c
```diff
@@ -38,6 +174,244 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+    __sync_fetch_and_and(&SRWLock->Ptr, ~SRW_GLOBAL_LOCK_MASK);
+}
+
+void ReleaseSRWLockShared (PSRWLOCK SRWLock)
+{
+    // Grab the reader lock
+    uintptr_t unlocked_val = SRWLock->Ptr & ~SRW_READER_LOCK_MASK;
+    while (!__sync_bool_compare_and_swap(&SRWLock->Ptr, unlocked_val, unlocked_val | SRW_READER_LOCK_MASK)) {
+        SwitchToThread();
+        unlocked_val = SRWLock->Ptr & ~SRW_GLOBAL_LOCK_MASK;
```
JayFoxRox, line 228: This looks wrong? Shouldn't this be the `READER_LOCK_MASK`?

This code pattern also repeats fairly often. I feel like you should have a helper function like `TryAquireLockMask(uintptr_t lock_mask)` or maybe even `AquireLock(uintptr_t lock_mask)`.

You could just do:

```c
while(!TryAquireLockMask(SRW_READER_LOCK_MASK)) {
  SwitchToThread();
}
```

lib/winapi/sync.c
```diff
@@ -38,6 +174,244 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+    }
+
+    // Decrease reader count by one
+    uintptr_t prev_val = __atomic_fetch_sub(&SRWLock->Ptr, 1, __ATOMIC_ACQ_REL);
+    if ((prev_val & SRW_READER_COUNT_MASK) == 1) {
```
JayFoxRox, line 233: nit: I might be a bit paranoid, but we could consider having `assert((prev_val & SRW_READER_COUNT_MASK) != 0)` to avoid people from releasing non-owned locks.

lib/winapi/winbase.h
```diff
@@ -34,6 +34,35 @@ typedef VOID (WINAPI *PFLS_CALLBACK_FUNCTION)(PVOID);
...
+} CONDITION_VARIABLE, *PCONDITION_VARIABLE;
+#define CONDITION_VARIABLE_INIT {INIT_ONCE_STATIC_INIT, {INVALID_HANDLE_VALUE, INVALID_HANDLE_VALUE}, 0}
+#define CONDITION_VARIABLE_LOCKMODE_SHARED 0x01
+
+typedef struct  _SRWLOCK
+{
+    DWORD_PTR Ptr;
+    // highest bit: reader lock (0=free, 1=locked)
+    // second-highest bit: global lock (0=free, 1=locked)
+    // remainder: blocking readers
```
JayFoxRox, line 61: nit: I think we normally put comments about variables above the field / variable?

lib/winapi/sync.c
thrimbor: Oops, can't believe I missed that, especially since I fixed that exact same error last time.
I've now introduced `TryAcquireLockMask` and `ReleaseLockMask` to hopefully make the code less error-prone.

lib/winapi/sync.c
thrimbor: Added.

lib/winapi/winbase.h
thrimbor: Changed.

lib/winapi/sync.c
JayFoxRox: That helped readability a lot! Thanks!

lib/winapi/sync.c
```diff
@@ -38,6 +40,141 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+        if (lpContext || (dwFlags & INIT_ONCE_ASYNC)) {
+            SetLastError(ERROR_INVALID_PARAMETER);
+            return FALSE;
+        }
+    } else {
+        lpContext = (PVOID)((ULONG_PTR)lpContext | INITONCE_DONE);
+    }
+
+    while (true) {
+        void *cur_val = lpInitOnce->Ptr;
```
JayFoxRox, line 389: nit: For clarity, `PVOID` would be good as type.

The name `lpInitOnce->Ptr` suggested to me that taking the pointer to it was wrong in `__sync_val_compare_and_swap`, but it appears to be correct?
If this was `PVOID`, then it would be more obvious that the first argument has to be `PVOID*` (so the `&` below is correct, as the intention is to change where the pointer points / state bits in pointer; not change data where the pointer points).

lib/winapi/sync.c
```diff
@@ -38,6 +40,141 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+                    return TRUE;
+                }
+                break;
+            case INITONCE_IN_PROGRESS:
+                if (dwFlags & INIT_ONCE_ASYNC) {
+                    SetLastError(ERROR_INVALID_PARAMETER);
+                    *fPending = TRUE;
+                    return FALSE;
+                }
+
```
JayFoxRox, line 327: nit: This empty line feels useless?

lib/winapi/sync.c
```diff
@@ -38,6 +40,141 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+                SwitchToThread();
+                __sync_synchronize();
```
JayFoxRox, line 329: I assume this is so that `lpInitOnce->Ptr & INITONCE_MASK` is read from memory again in the next loop iteration?

lib/winapi/sync.c
```diff
@@ -38,6 +40,141 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+                }
+                *fPending = TRUE;
+                return TRUE;
+            case INITONCE_DONE:
+                if (lpContext) {
+                    *lpContext = (PVOID)((ULONG_PTR)lpInitOnce->Ptr & ~INITONCE_MASK);
+                }
+                *fPending = FALSE;
+                return TRUE;
+                break;
```
JayFoxRox, line 345: This break is redundant

lib/winapi/sync.c
```diff
@@ -38,6 +40,141 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+
+BOOL InitOnceComplete (LPINIT_ONCE lpInitOnce, DWORD dwFlags, LPVOID lpContext)
+{
+    if (((ULONG_PTR)lpContext & INITONCE_MASK) != INITONCE_UNINITIALIZED) {
+        SetLastError(ERROR_INVALID_PARAMETER);
+        return FALSE;
+    }
+
+    if (dwFlags & INIT_ONCE_INIT_FAILED) {
+        if (lpContext || (dwFlags & INIT_ONCE_ASYNC)) {
```
JayFoxRox, line 380: I don't see anything about this disallowing `lpContext` on MSDN; I assume this was tested against Windows?

lib/winapi/sync.c
```diff
@@ -38,6 +40,141 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+            return FALSE;
+        }
+        if (((ULONG_PTR)lpInitOnce->Ptr & INITONCE_MASK) != INITONCE_DONE) {
+            *fPending = TRUE;
+            SetLastError(ERROR_GEN_FAILURE);
+            return FALSE;
+        }
+
+        if (lpContext) {
+            *lpContext = (LPVOID)((ULONG_PTR)lpInitOnce & ~INITONCE_MASK);
```
JayFoxRox, line 306: Shouldn't this be `lpInitOnce->Ptr`? This feels like a bug that would even break the implementation?

lib/winapi/winbase.h
```diff
@@ -34,6 +34,17 @@ typedef VOID (WINAPI *PFLS_CALLBACK_FUNCTION)(PVOID);
...
+typedef struct _INIT_ONCE
+{
+    PVOID Ptr;
+} INIT_ONCE, *PINIT_ONCE, *LPINIT_ONCE;
+#define INIT_ONCE_STATIC_INIT {0}
+#define INIT_ONCE_CTX_RESERVED_BITS 2
+#define INIT_ONCE_CHECK_ONLY 0x00000001UL
+#define INIT_ONCE_ASYNC 0x00000002UL
+#define INIT_ONCE_INIT_FAILED 0x00000004UL
+typedef BOOL CALLBACK PINIT_ONCE_FN(PINIT_ONCE, PVOID, PVOID *);
```
JayFoxRox, line 46: nit: Wouldn't one normally write this as `typdef BOOL(CALLBACK *PINIT_ONCE_FN)(PINIT_ONCE, PVOID, PVOID *);`? I think that's how we do it for the kernel functions.

lib/winapi/sync.c
```diff
@@ -1,9 +1,145 @@
...
+    assert(NT_SUCCESS(status));
+    return true;
+}
+
+VOID InitializeConditionVariable (PCONDITION_VARIABLE ConditionVariable)
+{
+    ConditionVariable->initOnce.Ptr = 0;
+    ConditionVariable->eventHandles[0] = INVALID_HANDLE_VALUE;
+    ConditionVariable->eventHandles[1] = INVALID_HANDLE_VALUE;
+    atomic_store(&ConditionVariable->waitCount, 0);
```
JayFoxRox, line 26: Shouldn't this be something like `__atomic_store_n` (GNU intrinsic)? Where does `atomic_store` come from? I only seem to find it in C++?

lib/winapi/sync.c
```diff
@@ -1,11 +1,145 @@
...
```
JayFoxRox, line 26: Wouldn't the `_Atomic` keyword take care of this? Why do we still need `atomic_store`?

JayFoxRox, line 26: Why isn't this being done as part of `condvar_init`?

I assume the intention of the InitOnce was so that people could use ConditionVariables even when using `CONDITION_VARIABLE_INIT`? But then you currently don't initialize `ConditionVariable->waitCount`?

lib/winapi/sync.c
```diff
@@ -1,11 +1,145 @@
...
+    status = NtCreateEvent(&ConditionVariable->eventHandles[1], NULL, NotificationEvent, FALSE);
+    assert(NT_SUCCESS(status));
+    return true;
+}
+
+VOID InitializeConditionVariable (PCONDITION_VARIABLE ConditionVariable)
+{
+    ConditionVariable->initOnce.Ptr = 0;
+    ConditionVariable->eventHandles[0] = INVALID_HANDLE_VALUE;
+    ConditionVariable->eventHandles[1] = INVALID_HANDLE_VALUE;
```
JayFoxRox, line 25: Why is this necessary if `condvar_init` does it anyway? It should only have to be set in the case where `InitOnceExecuteOnce` fails.

lib/winapi/sync.c
```diff
@@ -1,11 +1,145 @@
...
+    atomic_store(&ConditionVariable->waitCount, 0);
+
+    InitOnceExecuteOnce(&ConditionVariable->initOnce, condvar_init, ConditionVariable, NULL);
```
JayFoxRox, line 28: What if this function fails? Other instances, too.

lib/winapi/winbase.h
```diff
@@ -45,6 +45,14 @@ typedef struct _INIT_ONCE
 #define INIT_ONCE_INIT_FAILED 0x00000004UL
 typedef BOOL CALLBACK PINIT_ONCE_FN(PINIT_ONCE, PVOID, PVOID *);
 
+typedef struct _CONDITION_VARIABLE {
+    INIT_ONCE initOnce;
+    PVOID eventHandles[2];
+    _Atomic int waitCount;
```
JayFoxRox, line 51: Is this struct based on the MS internals? If yes, then is this keyword compatible with how MS does this? I worry about weird warnings / errors, if we ever pass this to functions which weren't written with C11 in mind.

lib/winapi/sync.c
```diff
@@ -1,11 +1,145 @@
...
+            EnterCriticalSection(CriticalSection);
+            return FALSE;
+        }
+    }
+
+    EnterCriticalSection(CriticalSection);
+    return TRUE;
+}
+
+BOOL SleepConditionVariableSRW (PCONDITION_VARIABLE ConditionVariable, PSRWLOCK SRWLock, DWORD dwMilliseconds, ULONG Flags)
```
JayFoxRox, line 67: nit: There's a lot of redundancy with `SleepConditionVariableCS` that could have been factored out; fine as-is though.

lib/winapi/winbase.h
```diff
@@ -45,6 +45,14 @@ typedef struct _INIT_ONCE
 #define INIT_ONCE_INIT_FAILED 0x00000004UL
 typedef BOOL CALLBACK PINIT_ONCE_FN(PINIT_ONCE, PVOID, PVOID *);
 
+typedef struct _CONDITION_VARIABLE {
+    INIT_ONCE initOnce;
+    PVOID eventHandles[2];
```
JayFoxRox, line 50: nit: This could have deserved a comment about what each of these is used for.

- In my initial review-skim I was wondering why 2 of these were necessary.
- Reading the code, I still wondered a bit until I realized that one is for `WakeConditionVariable` and the other one for `WakeAllConditionVariable`.
- However, even then I was still wondering where this difference was implemented, until I remembered that Microsoft sets different auto-resets on events during creation, which eventually lead me to look for differences in `NtCreateEvent`.

Therefore a very helpful comment would have been along the lines of:

```
// eventHandles[0]: Created as SynchronizationEvent to wake a single thread
// eventHandles[1]: Created as NotificationEvent to wake all threads at once
```

(That comment answers everything I have wondered about)

lib/winapi/sync.c
```diff
@@ -1,11 +1,145 @@
...
+
+    NTSTATUS status;
+    status = NtSetEvent(ConditionVariable->eventHandles[1], NULL);
+    assert(NT_SUCCESS(status));
+}
+
+VOID DestroyConditionVariable (PCONDITION_VARIABLE ConditionVariable)
+{
+    NtClose(ConditionVariable->eventHandles[0]);
+    NtClose(ConditionVariable->eventHandles[1]);
```
JayFoxRox, line 140: For potential error spotting, I'd suggest to also mark `initOnce.Ptr` as `DONE`, and `eventHandles` to the invalid values?

lib/winapi/sync.c
```diff
@@ -1,11 +1,145 @@
...
+VOID WakeAllConditionVariable (PCONDITION_VARIABLE ConditionVariable)
+{
+    InitOnceExecuteOnce(&ConditionVariable->initOnce, condvar_init, ConditionVariable, NULL);
+
+    NTSTATUS status;
+    status = NtSetEvent(ConditionVariable->eventHandles[1], NULL);
+    assert(NT_SUCCESS(status));
+}
+
+VOID DestroyConditionVariable (PCONDITION_VARIABLE ConditionVariable)
```
JayFoxRox, line 137: `UninitializeConditionVariable` would sound more logical to me. Initialization creates the events; so `Uninitialize` would be a counterpart.
`Destroy` would, to me, imply that there'd be a matching `Create` somewhere, which allocated that memory (which is not the case).

lib/winapi/winbase.h
thrimbor: I've added the comments.

lib/winapi/sync.c
thrimbor: Ouch. Yes, that should be `lpInitOnce->Ptr` instead of just `lpInitOnce`. The mask is correct though.

lib/winapi/sync.c
thrimbor: Maybe? Tbh, I don't remember and currently can't find any test code on my HDD. I did switch the machine I do my experiments on a while ago though, so maybe I just wiped it with the old Windows installation.

I did just look up that part in ReactOS though, and it does the same here.

lib/winapi/sync.c
```diff
@@ -38,6 +40,377 @@ VOID LeaveCriticalSection (LPCRITICAL_SECTION lpCriticalSection)
...
+                SetLastError(ERROR_GEN_FAILURE);
+                return FALSE;
+        }
+    }
+}
+
+static BOOL WINAPI condvar_init (PINIT_ONCE InitOnce, PVOID Parameter, PVOID *Context)
+{
+    PCONDITION_VARIABLE ConditionVariable = Parameter;
+    NTSTATUS status;
```
JayFoxRox, line 283: nit: Before creating new events, you could probably check if this thing is really clean at this point:

- `ConditionVariable->initOnce.Ptr == 0`
- `ConditionVariable->waitCount == 0`
- `ConditionVariable->eventHandles[0] == INVALID_HANDLE_VALUE`
- `ConditionVariable->eventHandles[1] == INVALID_HANDLE_VALUE`

You'd just have to reintroduce the initialization of the events to `INVALID_HANDLE_VALUE` in `InitializeConditionVariable`.
These assumptions should always be true then, because they come from `CONDITION_VARIABLE_INIT` or `InitializeConditionVariable`.

However, I'll leave up to you wether you consider this a useful addition that's worth adding.

lib/winapi/sync.c
JayFoxRox: > I can't follow here

I have no idea what I meant here either. I'll just mark this conversation as resolved.

I agree with your response. `waitCount` will be initialized by `CONDITION_VARIABLE_INIT` (or by `InitializeConditionVariable` on the linked line).

lib/winapi/sync.c
JayFoxRox: See https://github.com/XboxDev/nxdk/pull/274#discussion_r424076346 ; I'll mark this conversation as resolved.

lib/winapi/sync.c
JayFoxRox: In that case this either deserves an `assert` or at very least a comment.
I have tried to follow the logic, and there appear to be various non-trivial ways in which `InitOnceExecuteOnce` could fail (through `InitOnceBeginInitialize` / `InitOnceComplete`).

A similar instance is the call to `InitOnceComplete` in `InitOnceExecuteOnce` where the return value is never checked. I assume that the assumption is that it always succeeds?
An `assert` could help to clarify if this is assumed, or a comment would help to understand why that wouldn't always be the case.

Such assertions would mostly help with self-checking and finding race-conditions or similar things. Given the complexity of the code I think it would be a good idea to add them.

lib/winapi/sync.c
JayFoxRox: This was resolved by using builtins over `_Atomic` keyword.

lib/winapi/sync.c
thrimbor: I've added the requested asserts now.

lib/winapi/sync.c
thrimbor: Ah, I forgot about this.
We can't really check for `ConditionVariable->initOnce.Ptr == 0` here, because the InitOnce API functions will change the value of this variable before calling this function, so it's never zero here.
I did add checks for the other conditions though.

JayFoxRox:
I'm merging this untested, because it has been here for way too long.
We can iterate if it's broken.

> [...] condition variables is basically equivalent to how I implemented it for PDCLib

There have been a couple of improvements in this new implementation during this review, which probably aren't in PDCLib yet.
So please consider improving other variants of this code in PDCLib if they still exist @thrimbor

PR #275, opened 2020-01-06
libc: Add alloca() / _alloca()

thrimbor:
I'm not a huge fan of `alloca`, but some code out there (like libraries used by SDLPoP) use it, and it's an easy thing to add. Some test code can be found [here](https://gist.github.com/thrimbor/2815f00a97966b97028ac0a23b94bde3), it mainly checks whether we corrupt the stack.

### DIFF
--- a/lib/xboxrt/libc_extensions/malloc.h
+++ b/lib/xboxrt/libc_extensions/malloc.h
@@ -1,0 +1,10 @@
+#ifndef _XBOXRT_MALLOC_H_
+#define _XBOXRT_MALLOC_H_
+
+// Some code expects to find a declaration of malloc here, so we simply include the whole C header
+#include <stdlib.h>
+
+// Forward alloca to the builtin with the Windows-specific name
+static void *alloca (size_t size) { return _alloca(size); }
+
+#endif


lib/xboxrt/libc_extensions/malloc.h
```diff
@@ -0,0 +1,10 @@
+#ifndef _XBOXRT_MALLOC_H_
+#define _XBOXRT_MALLOC_H_
+
+// Some code expects to find a declaration of malloc here, so we simply include the whole C header
+#include <stdlib.h>
+
+void *_alloca (size_t size);
+static void *alloca (size_t size) { return _alloca(size); }
```
JayFoxRox: My own implementation in some apps used `void *__builtin_alloca (size_t size)` from https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html. I'm not sure about benefits or drawbacks, but would assume better error-checking at reduced compiler-compatibility.
Did you consider that?

JayFoxRox: Please add a comment to note which standard this extension comes from.

This implementation probably won't work for me, because most code I have seen depends on alloca.h instead of malloc.h; also see Linux man-page:

```
SYNOPSIS
       #include <alloca.h>

       void *alloca(size_t size);

[...]

CONFORMING TO
       This function is not in POSIX.1.

       There is evidence that the alloca() function appeared in 32V, PWB, PWB.2, 3BSD, and 4BSD.  There is a man page
       for it in 4.3BSD.  Linux uses the GNU version.
```

thrimbor: tbh I didn't expect that to exist. I've done some experimenting now and discovered that `_alloca` is a builtin, too, so it didn't even use the label I added in `chkstk.s`, but just emitted code that called `_chkstk`.

I modified the header to just forwards `alloca` to `_alloca` now.

thrimbor: It's implemented as described by Microsoft in their C runtime docs. The code is different now, but I added a comment mentioning that `_alloca` is Windows-specific.

Microsoft describes that `_alloca` lives in `malloc.h` (they often don't provide the non-underscored variant by default). We could move the definition of `alloca` to a new `alloca.h` and include that here, if it's a portability issue.

JayFoxRox: Did anyone test this patch? This should have been a `#define` - otherwise the `_alloca` runs in the local stack frame of this function and will be invalid by the time it returns.

mborgerson: Good catch! Agree, we should make this a #define instead.

PR #278, opened 2020-01-09
build: Add support for 32 bits build environment

sam-itt:
The current code only works on x86_64. The proposed patch adds support for i686.

cgc: version 3.1.0013, build date Apr 18 2012 13:43:10.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,13 +15,19 @@ NXDK_STACKSIZE = 65536
 endif
 
 UNAME_S := $(shell uname -s)
+UNAME_M := $(shell uname -m)
+
 ifeq ($(UNAME_S),Linux)
 LD           = lld -flavor link
 LIB          = llvm-lib
 AS           = clang
 CC           = clang
 CXX          = clang++
+ifneq ($(UNAME_M),x86_64)
+CGC          = $(NXDK_DIR)/tools/cg/linux/cgc.i386
+else
 CGC          = $(NXDK_DIR)/tools/cg/linux/cgc
+endif #UNAME_M != x86_64
 endif
 ifeq ($(UNAME_S),Darwin)
 LD           = /usr/local/opt/llvm/bin/lld -flavor link


Makefile
```diff
@@ -15,13 +15,18 @@ NXDK_STACKSIZE = 65536
...
 
 UNAME_S := $(shell uname -s)
+UNAME_M := $(shell uname -m)
 ifeq ($(UNAME_S),Linux)
 LD           = lld -flavor link
 LIB          = llvm-lib
 AS           = clang
 CC           = clang
 CXX          = clang++
+ifneq ($(UNAME_S),x86_64)
```
mborgerson: Did you mean to compare against `UNAME_M`?

Makefile
```diff
@@ -15,13 +15,18 @@ NXDK_STACKSIZE = 65536
...
+CGC          = $(NXDK_DIR)/tools/cg/linux/cgc.i386
+else
 CGC          = $(NXDK_DIR)/tools/cg/linux/cgc
+endif #UNAME_S != x86_64
```
mborgerson: Same as above

mborgerson:
While we don't go out of our way to support 32-bit compilation, having a copy of cgc for 32-bit platforms isn't a bad idea. Though ideally we could have our own tool that a user can build from source. Can you please provide the location where you've downloaded this binary (was it from [here](https://developer.nvidia.com/cg-toolkit-download)?)

Makefile
```diff
@@ -15,13 +15,18 @@ NXDK_STACKSIZE = 65536
 endif
 
 UNAME_S := $(shell uname -s)
+UNAME_M := $(shell uname -m)
```
mborgerson: nit: Prefer blank line after this

JayFoxRox:
Thanks for the patch.

I agree with @mborgerson - adding 32 bit is fine, but also don't expect great support.

> Though ideally we could have our own tool that a user can build from source.

I have created a new issue to discuss (and keep track of it) in #280 

---

@sam-itt  For the PR, please adapt our standard commit style of `prefix: Action something`:
- Usually in present tense
- Use prefix that was also used in similar context before (check history of files you edit)
- Respect case-sensitiveness
    - Lowercase "prefix:"
    - Capital "Action"
    - Respect spelling of third party projects ("Cg" or "cgc" but not "CG") and company names
- Keep it short enough so it doesn't line-break

Also please mention in your PR description which version of cgc this is.

As you haven't contributed before (therefore untrusted) I think we should also confirm the hashes of the binaries.

Makefile
sam-itt: Yes, exactly. Typo...

Makefile
sam-itt: same mistake

sam-itt:
> While we don't go out of our way to support 32-bit compilation, having a copy of cgc for 32-bit platforms isn't a bad idea. Though ideally we could have our own tool that a user can build from source. Can you please provide the location where you've downloaded this binary (was it from [here](https://developer.nvidia.com/cg-toolkit-download)?)

I got it from gentoo ebuild media-gfx/nvidia-cg-toolkit which fetches http://developer.download.nvidia.com/cg/Cg_3.1/Cg-3.1_April2012_x86.tgz

JayFoxRox:
I probably should have been more clear, my comment wasn't about the PR title, but the commit title. Please squash up into a single commit and use a name similar to the PR title (although it should probably mention Cg; I'd just recommend `cg: Add x86 support`).

dracc:
Maybe `cg: Add Linux i386 support` even?

Makefile
```diff
@@ -15,13 +15,19 @@ NXDK_STACKSIZE = 65536
...
+
 ifeq ($(UNAME_S),Linux)
 LD           = lld -flavor link
 LIB          = llvm-lib
 AS           = clang
 CC           = clang
 CXX          = clang++
+ifneq ($(UNAME_M),x86_64)
+CGC          = $(NXDK_DIR)/tools/cg/linux/cgc.i386
```
JayFoxRox: I'd suggest to instead move the files around, so we have `linux/x86_64/cgc` and `linux/???/cgc` for the i386 variant. We can then probably just use `$(NXDK_DIR)/tools/cg/linux/$(UNAME_M)/cgc`.
This will also lead to meaningful error messages.

Extending the original filename sounds non-standard and confusing.

I don't feel too strongly but if this is not done in this PR, then I'd probably open an issue about this.

mborgerson:
I have confirmed md5sum of this new binary matches the [official distribution from nvidia](http://developer.download.nvidia.com/cg/Cg_3.1/Cg-3.1_April2012_x86_64.tgz).

Makefile
mborgerson: Let's do it later, as the other platforms don't have such directory structure yet

JayFoxRox: I think this is primarily a Linux issue, so I'd be fine if this is on Linux. I believe macOS is exclusively x64 now, and I *assume* the existing Windows binary is x86 which also works on x64?

mborgerson:
Thank you @sam-itt! I've taken the liberty of squashing your commits and re-wording slightly to match our conventions.

PR #282, opened 2020-01-10
build: Fix LLD hang on Windows by disabling multi-threaded linking

### LINKED ISSUE #279: lld hanging at linking stage
There have previously been multiple unexplained reports of this happening on the XboxDev Discord in the past.
In my case, I've just set up nxdk on a fresh Windows 10 + msys2 installation, following the instructions found [here](https://github.com/XboxDev/nxdk/wiki/Install-the-Prerequisites#option-2-msys2), and have encountered an issue with the latest package to date, lld 9.0.0-3, where compilation will hang at the linking stage (`[LD] main.exe`), even for any of the samples.

While I have not tried any other minor versions of 9.0.0, downgrading the lld package to 8.0.1-3 no longer caused any hangs at `[LD] main.exe` and proceeds to link just fine.

Likely an lld issue rather than nxdk, but just making an issue here to keep a record of things.

thrimbor:
Since we seem to hit the LLD hang with every single CI run now, I did some investigating on a Windows 10 VM. I was able to reproduce the issue in 100% of all attempts and found out that it is related to LLD using multi-threaded linking by default, and that we can work around the issue by disabling that.

This is an issue that has been encountered by others [before](https://github.com/android/ndk/issues/855) and was [seemingly fixed](https://reviews.llvm.org/D53968) but seems to have returned in some other form.

Since, according to reports in Discord, this problem doesn't always happen, I tested this patch by building the recently merged gamma sample ten times with and ten times without the patch. Without it, it always hung, with it, it always worked.

Fixes #279.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,10 @@ NXDK_CXXFLAGS = -I$(NXDK_DIR)/lib/libcxx/include $(NXDK_CFLAGS) -fno-threadsafe-
 NXDK_LDFLAGS = -subsystem:windows -dll -entry:XboxCRTEntry \
                -stack:$(NXDK_STACKSIZE) -safeseh:no
 
+# Multithreaded LLD on Windows hang workaround
+ifneq (,$(findstring MINGW,$(UNAME_S)))
+NXDK_LDFLAGS += -threads:no
+endif
 
 ifeq ($(DEBUG),y)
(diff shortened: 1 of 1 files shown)


mborgerson:
Thanks for looking into this!

PR #285, opened 2020-01-14
Fix some xboxkrnl prototypes

JayFoxRox:
* `KeRestoreFloatingPointState` was not marked as `NTAPI` and wasn't linkable (as name decoration in import lib is different). I ran into this, because a future submodule PR of mine will use a DPC which might require the FPU.
* `HalBootSMCVideoMode` is a variable, not a function. I found this using: `grep XBAPI -R ./ | grep -v FASTCALL | grep -v NTAPI | grep -v CDECL | grep \(`.


thrimbor:
The changes look good. Merging.

PR #286, opened 2020-01-14
Fix OpenXDK XAudio in XQEMU

JayFoxRox:
OpenXDK XAudio actually contains some code that probably doesn't conform with AC97.
It still works on hardware, but it doesn't work in XQEMU.

This PR fixes that, so that audio works in XQEMU.

I also have an explanation and a patch for XQEMU in https://github.com/JayFoxRox/xqemu/pull/41 (including reasons as to why I don't intend to upstream it).

The primary motivation for this change is to make it more standard, and also to make it easier to test in XQEMU.

**I have not tested this patch on hardware yet.**

### DIFF
--- a/lib/hal/audio.c
+++ b/lib/hal/audio.c
@@ -155,16 +155,18 @@ void XAudioInit(int sampleSizeInBits, int numChannels, XAudioCallback callback, 
 void XAudioPlay()
 {
 	volatile AC97_DEVICE *pac97device = &ac97Device;
-	pac97device->mmio[0x118>>2] = (unsigned int)0x1d000000; // PCM out - run, allow interrupts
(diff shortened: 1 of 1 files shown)


thrimbor:
I tested this in XQEMU and on my Xbox. I came across another issue in the code (PRed in #294), and while it still works on my Xbox, the changes seem to have no effect on my XQEMU, I still get no audio (I'm still using Audio PR #3, and I have audio during the BIOS boot).

dracc:
Works for me with XQEMU 0ced417e73.
Screenshot

thrimbor:
Alright, re-tested on xqemu master, and it works just fine. Merging.

PR #287, opened 2020-01-15
winapi: Add INT / PINT / LPINT / UINT / PUINT / LPUINT types

JayFoxRox:
This exposes these types through windef.h (which primarily includes xboxdef.h).
This is necessary for https://github.com/XboxDev/nxdk-sdl/pull/24#discussion_r366668605.

I wasn't sure wether I should add these types to windef.h directly as they aren't used in the kernel (yet?).
I decided for xboxdef.h because our windef.h is really slim right now. Also our xboxdef.h already has `DWORD_PTR` which isn't used in the kernel either (also, that one should actually be in BaseTsd.h instead, so we have to refactor in the future anyway). 

You can find the documentation for these types here: https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types

*(Random note: Our windef.h should probably also be renamed to WinDef.h in the future, to deal with case-sensitivity)*


thrimbor:
Looks good. Merging.

PR #295, opened 2020-02-09
SDL2: Update submodule to 63992b5

dracc:
This update will bring a bunch of fixes and a few changes;
- SDL_GameController thumbstick Y axes have been inverted to match PC behavior
- C++ header compatibility have been fixed
- Windows timer support enabled
- Windows thread support enabled 
- Dummy audio driver inherited
- Video mode is now set to match GPU framebuffer

I'd really like for NevolutionX to have the header compatibility, SDL timers and the GameController fix.
Could probably wait for XboxDev/nxdk-sdl#20 but I feel these changes alone are urgent enough to warrant a submodule update.


JayFoxRox:
Please follow conventions of commit and PR description seen here <https://github.com/XboxDev/nxdk/pulls?utf8=%E2%9C%93&q=is%3Apr+%22update+submodule%22+is%3Amerged+>

That is:

* Lowercase commit prefix.
* Mention version you are updating to in commit title.
* Use PR description to explain summary of changes (which is critical for people to update their code, as not many people follow submodule development).
* Possibly provide reasons as to why this needs to be updated now (and why we shouldn't wait for possible other PRs on that repo first).

thrimbor:
Merging. Thanks!

PR #297, opened 2020-02-10
lwip C++ compat fixes

### LINKED ISSUE #296: lwip doesn't work with C++ projects
I'm guessing it's due to the fact that the `-I` flags only get added to `CFLAGS` and thus the files aren't found by the C++ compiler.

thrimbor:
This fixes some issues that occur when using lwip from C++, such as missing include paths, name mangling, and issues with including `errno.h.` (which was encountered by @dracc).

Mostly tested by building one of the http server samples as C++ code, and @dracc's ftpserver-branch of NevolutionX.

Closes #296.

### DIFF
--- a/lib/net/Makefile
+++ b/lib/net/Makefile
@@ -18,3 +18,6 @@ SRCS += $(LWIPSRCS) $(DRIVERSRCS)
 CFLAGS += -I $(LWIPDIR)/include
 CFLAGS += -I $(NXDK_DIR)/lib/net/nforceif/include
 CFLAGS += -I $(NXDK_DIR)/lib/net/pktdrv
+CXXFLAGS += -I $(LWIPDIR)/include
+CXXFLAGS += -I $(NXDK_DIR)/lib/net/nforceif/include
+CXXFLAGS += -I $(NXDK_DIR)/lib/net/pktdrv
--- a/lib/net/nforceif/include/lwipopts.h
+++ b/lib/net/nforceif/include/lwipopts.h
@@ -43,7 +43,7 @@
 #define LWIP_IPV4 1
 #define LWIP_IPV6 1
 #define LWIP_DEBUG 1
-#define LWIP_ERRNO_INCLUDE <errno.h>
+#define LWIP_ERRNO_STDINCLUDE 1
 #define LWIP_COMPAT_MUTEX_ALLOWED
 #define SYS_LIGHTWEIGHT_PROT 0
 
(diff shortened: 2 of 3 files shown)


lib/net/pktdrv/pktdrv.h
```diff
@@ -1,6 +1,10 @@
 #ifndef _Pktdrv_
 #define _Pktdrv_
 
+#ifdef __cplusplus
+extern "C" {
+#endif
```
JayFoxRox: I wonder if we should even C++-ify all of our headers.
If it's well defined that certain APIs are C APIs, then I don't see the issue with doing it like this instead:

```c
// (in app.cpp)

#include <windows.h>

extern "C" {
#include "foo.h"
}
```

lib/net/nforceif/include/lwipopts.h
```diff
@@ -437,7 +437,13 @@
 #define TCP_QLEN_DEBUG   LWIP_DBG_ON
 #define TCP_RST_DEBUG    LWIP_DBG_ON
 
+#ifdef __cplusplus
+extern "C" {
+#endif
 extern unsigned char debug_flags;
+#ifdef __cplusplus
+}
+#endif
```
JayFoxRox: Wouldn't this file only be included by C code in lwip? When would this ever be compiled by a C++ compiler?

thrimbor: It gets pulled in indirectly by lwip headers.

lib/net/pktdrv/pktdrv.h
thrimbor: I think we should - people usually expect C headers to just work in C++, even though the languages are quite different internally, and I consider it to be much cleaner to adjust the header once.

lib/net/nforceif/include/lwipopts.h
JayFoxRox: I see. lwip seems to have this C++ handling in each header *after* `#include` lines.
Resolved.

lib/net/Makefile
```diff
@@ -18,3 +18,6 @@ SRCS += $(LWIPSRCS) $(DRIVERSRCS)
 CFLAGS += -I $(LWIPDIR)/include
 CFLAGS += -I $(NXDK_DIR)/lib/net/nforceif/include
 CFLAGS += -I $(NXDK_DIR)/lib/net/pktdrv
+CXXFLAGS += -I $(LWIPDIR)/include
```
mborgerson: Ideally we'd just have one `INCLUDES` variable that is included for both C and C++ definitions. Don't have to fix this now, but thought for the future

PR #298, opened 2020-02-11
xboxrt: Fix lld dropping the _fltused symbol when using LTO

thrimbor:
For some reason `_fltused` seems to get dropped somewhere when using LTO, even though it's implicitly referenced. I'm not sure what exactly causes this, and whether this is expected with this special symbol, or a bug in clang, or lld...

Adding a linker pragma fixes this, though. It should only be added when the obj file is pulled in, so it should be fine to have it there.

### DIFF
--- a/lib/xboxrt/c_runtime/_fltused.c
+++ b/lib/xboxrt/c_runtime/_fltused.c
@@ -1,1 +1,2 @@
+#pragma comment(linker, "/include:__fltused")
 int _fltused = 1;


JayFoxRox:
Also see https://github.com/XboxDev/nxdk-pdclib/pull/26#issuecomment-585841729

JayFoxRox:
I plan to merge this later today. See https://github.com/XboxDev/nxdk-pdclib/pull/26#issuecomment-608487095

JayFoxRox:
Merging. See https://github.com/XboxDev/nxdk-pdclib/pull/26#issuecomment-610114742

PR #299, opened 2020-02-22
build: Use /fixed:no instead of /dll

thrimbor:
Just a little thing I noticed when I was investigating optimization options - we're currently using `/dll` in order to get a PE file with a `.reloc` section, but we can have that by just using `/fixed:no`, which has the benefit of not generating an unnecessary import lib (which may confuse people).

I tested the change, and everything still seems to work fine.


Teufelchen1:
[Not being deep into the topic]
Is testing on real hardware required?

JayFoxRox:
Merging untested. Looks good to me and no vetos or tests in a long time.

PR #300, opened 2020-02-24
lwip: Use malloc instead of a memory pool

dracc:
While developing an FTP server, I noticed that the default setting in FileZilla requires more than 2 concurrent connections.
Using the memory pool implementation required a hard limit for concurrent connections.
By instead going with malloc usage, those limits are ignored.

I've also removed some defines rendered unused from this change.


Teufelchen1:
Hi 👋 
the documentation for your changes in lwIP is [here](https://www.nongnu.org/lwip/2_1_x/group__lwip__opts__mem.html#ga4ef345cc270912bd2230b1c5ec51dfc8).

Also, note that the documentation states that this change can impact the performance. I do not believe that this is really an issue on the xbox(as she is much more powerful than typical lwIP use cases) but you might want to check it out regarding your performance issues with you FTP.

thrimbor:
> The lwip docs warn about heap fragmentation and I'm not sure what they meant
> by that :
> a) slower due to contiguous virtual memory mapped to non-contiguous hardware blocs
> b) the hardware is fed with stop/start non-virtual adresses and you can have something
> in the middle due to the virtual region is mapped to a non-contiguous hardware memory region.
Neither. It refers to the problem where you theoretically have enough free heap memory to satisfy a request, but there isn't a large enough contiguous free area ([Wikipedia has an example for this](https://en.wikipedia.org/wiki/Fragmentation_(computing)#External_fragmentation)).
There is no ultimate solution for this, but the problem can be reduced by using an allocator that uses binning (which dlmalloc does). Virtual memory also helps here.
Overall, I don't expect fragmentation to cause issues anytime soon.

> Would it be possible to use MmAllocateContiguousMemoryEx/MmFreeContiguousMemory as
> mem_clib_malloc/mem_clib_free ? I have no idea about allocation rates/sizes when enabling
> MEM_LIBC_MALLOC. It would be interesting to see what happens when dealing with some traffic
> (uploading/downloading with big files).
Possible, but unnecessary / wasteful. The memory lwIP requires here doesn't need to be physically contiguous, and you'd waste at least an entire page on a potentially much smaller request (changing the virtual memory mapping also impacts the TLB).

thrimbor:
Testing showed no issues. I'm intending to do some more changes to this file during my network code work, so I'm merging this now.

PR #302, opened 2020-03-01
lwip: Enable support for DHCPv6

### LINKED ISSUE #301: lwip: DHCP6 not supported
The define `LWIP_IPV6_DHCP6` is not set in [lib/net/nforceif/include/lwipopts.h](https://github.com/XboxDev/nxdk/blob/master/lib/net/nforceif/include/lwipopts.h). This means we currently don't support DHCP on IPv6, which we probably should.

dracc:
Title says it all.
Closes #301


GXTX:
Is this tested working on just enabling it via a define?

dracc:
@GXTX: I don't have an IPv6 network at hand but I couldn't build NevolutionX using the dhcp6 functions without this change. Fixing any lwip issues is probably out of scope for this repo as lwip is currently a submodule.

Teufelchen1:
Hi,

the documentation reference is [here](https://www.nongnu.org/lwip/2_1_x/group__lwip__opts__dhcpv6.html#ga1ba67b6665026ec0c688dc4b0df047a6).

sam-itt:
I think that this should not be enabled by default and be left to the user to enable/disable 
 in the Makefile. IPv6 has no real use as of today and as been known to impact performance.

Teufelchen1:
Yes, IPv6 normally performs better than IPv4 - but mainly in routing. IPv6 is definitely used as of today since many ISP have deployed it. In addition, IPv6 is widely used in embedded systems. Take a look at 802.15.4 & 6LoWPAN.
I only see benefits in enabling it, could you specify what exactly would be the benefit of disabling it as a default?

thrimbor:
I'd suggest naming the commit something like `Enable support for DHCPv6` instead since the feature only gets enabled and not actually added here.

dracc:
@thrimbor Agreed; Commit title has now been updated. Thanks!

thrimbor:
I expect more changes to be necessary for fully working and usable IPv6 support, but as a quick fix, I consider this to be fine. Merging.

PR #304, opened 2020-03-04
Add libpng & zlib

### LINKED ISSUE #162: zlib port
I have prototyped a zlib port for https://github.com/madler/zlib in https://github.com/JayFoxRox/neverball/blob/xbox/xbox/Makefile-zlib.xbox

Someone should clean it up, and upstream it; we can probably get around having a fork by doing configuration out-of-tree.

### LINKED ISSUE #163: libpng port
I have prototyped a libpng port for https://github.com/glennrp/libpng in https://github.com/JayFoxRox/neverball/blob/xbox/xbox/Makefile-libpng.xbox

Someone should clean it up, and upstream it; we can probably get around having a fork by doing configuration out-of-tree.

Depends on #162

thrimbor:
This adds the latest releases of libpng and zlib. I've hacked together some ugly test code [here](https://gist.github.com/thrimbor/47d47b4cd4f42be8d53eb1ab93966b4c), but it displays incorrectly - it's probably an error in using the library, since it is able to decode the image, and the image is decipherable on the screen (the code is ugly and I'm tired so I probably won't fix it). I'm assuming most future users will use SDL_image anyway.

I've decided against introducing more Makefile variables to control linking against these libraries, instead, they're always added to the linker command line. This increases build time slightly when doing a full rebuild, but binaries not requiring them won't get any larger because the linker only includes objects from libraries if they're referenced somewhere.
Our current system of enabling additional libraries to link against is quite messy, and while this isn't ideal either, I'm hoping to avoid making it more complex until we transition to a better system, where users can just specify which libraries they want to link against (which isn't possible yet because we require dependencies in Makefiles to actually build the libraries).

Fixes #162.
Fixes #163.


sam-itt:
Here is a fix for the test case: https://gist.github.com/sam-itt/715bf2c255c06ebeae86491f4f802624.

I've tried to build teo-ng with this PR, it produces an executable that fails with an "out of memory" error. The code builds and runs successfully with my local nxdk. 

I've started to investigate what are the differences but I need more time.

JayFoxRox:
I'm personally not too happy about Makefile-zlib and Makefile-libpng in lib/; we usually would have lib/zlib/Makefile and the submodule in lib/zlib/zlib. This would also make them more modular in case we ever made a package manager (as we could move the lib folders into separate submodules)

Willing to merge regardless but would like to get a comment on how we should structure our lib folder in the future.

dracc:
I, too, would prefer the layout described by @jayfoxrox.

thrimbor:
> I'm personally not too happy about Makefile-zlib and Makefile-libpng in lib/; we usually would have lib/zlib/Makefile and the submodule in lib/zlib/zlib. This would also make them more modular in case we ever made a package manager (as we could move the lib folders into separate submodules)

There's no Makefile-libpng, libpng already uses the layout you describe (I chose that because libpng requires an extra header for configuration, and I didn't want to have that in lib/).
I've changed the zlib submodule layout now, too.

JayFoxRox:
Thanks! Changes look fine; except for commit title. You are using libpng 1.6.37 (April 2019); not 1.6.36 as stated in commit title.

I hope someone can do a final test to avoid any errors from moving the Makefiles around. I'll merge next time I visit this PR if no other mainer reviews / tests / merges it before me.

thrimbor:
> Thanks! Changes look fine; except for commit title. You are using libpng 1.6.37 (April 2019); not 1.6.36 as stated in commit title.

Oops. Fixed.

.gitmodules
```diff
@@ -19,3 +19,9 @@
 	path = lib/libcxx
 	url = https://github.com/XboxDev/nxdk-libcxx.git
 	branch = nxdk
+[submodule "lib/zlib"]
+	path = lib/zlib/zlib
+	url = https://github.com/madler/zlib
```
JayFoxRox, line 24: nit: This is the only submodule to use the short URL without `.git`

thrimbor: Changed.

JayFoxRox:
Only tested compilation, but that seems to work fine. Merging.

PR #305, opened 2020-03-09
Implement GetLogicalDrives() / GetLogicalDriveStringsA()

thrimbor:
#230 introduced the ability to mount drives, but we didn't have an option to list mounted drives yet.
This PR introduces `GetLogicalDrives()`, which returns a bitmask representing mounted drive letters, and `GetLogicalDriveStringsA()`, which puts null-terminated drive letter strings into a buffer. The buffer length parameter for `GetLogicalDriveStringsA()` is a bit weird (it doesn't count the space required for the null-terminator despite the function writing it to the buffer), so I added a sample demonstrating it. I also checked the behavior against Windows.


lib/winapi/filemanip.c
```diff
@@ -235,3 +239,80 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+DWORD GetLogicalDrives (VOID)
+{
+    DWORD result = 0;
+    NTSTATUS status;
+    ANSI_STRING path;
+    HANDLE handle;
+    OBJECT_ATTRIBUTES attributes;
+    struct {
+        OBJECT_DIRECTORY_INFORMATION objDirInfo;
+        CHAR filenameBuf[100];
```
JayFoxRox: 100 is a bit arbitrary? Why not keep this shorter? We check for length of 2 shortly afterwards anyway.

If there's a custom device with a longer name (is that possible?) this would break this function, too.

Instead, why not just ignore entries which are too long? I'd assume that `NtQueryDirectoryObject` would report some error if the buffer is too short?

lib/winapi/filemanip.c
```diff
@@ -235,3 +239,80 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+
+    InitializeObjectAttributes(&attributes, NULL, OBJ_CASE_INSENSITIVE, ObDosDevicesDirectory(), NULL);
+    status = NtOpenDirectoryObject(&handle, &attributes);
+
+    if (!NT_SUCCESS(status)) {
+        SetLastError(RtlNtStatusToDosError(status));
+        return 0;
+    }
+
+    for (ULONG index = 0;;) {
```
JayFoxRox: nit: This feels like a `while` loop to me.

lib/winapi/filemanip.c
```diff
@@ -235,3 +239,80 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+
+    if (nBufferLength == 0 || nBufferLength < requiredBufLength) {
+        return requiredBufLength+1;
+    }
+
+    for (int bit = 0; bit < 26; bit++) {
+        if (driveBitmask & (1 << bit)) {
+            *lpBuffer++ = 'A' + bit;
+            *lpBuffer++ = ':';
+            *lpBuffer++ = '\\';
```
JayFoxRox: The MSDN isn't clear about the format of the returned strings, did you confirm this format on Windows?

samples/winapi_drivelist/main.c
```diff
@@ -0,0 +1,58 @@
...
+    DWORD charsWritten = GetLogicalDriveStringsA(sizeof(buffer)-1, buffer);
+    if (charsWritten == 0) {
+        // Additional error info can be retrieved with GetLastError()
+        debugPrint("Failed to retrieve byte strings!\n");
+        Sleep(5000);
+        return 1;
+    }
+
+    debugPrint("Drives found:\n");
+    for (char *drive = buffer; drive < buffer + charsWritten;) {
```
JayFoxRox: I'd prefer a while loop. A `for` loop doesn't help with readability in my opinion.

samples/winapi_drivelist/main.c
```diff
@@ -0,0 +1,58 @@
...
+        debugPrint("%s\n", drive);
+        while(*drive++);
```
JayFoxRox: Wouldn't this stop on the `\0` after the first iteration (checks `'\\'` but post-increment stops at `\0`)? So on the second iteration, you'd print an empty string.

I also think that checking for the final `\0` sounds like a better idea.

samples/winapi_drivelist/main.c
```diff
@@ -0,0 +1,58 @@
...
+        return 1;
+    }
+    debugPrint("Drive bitmask: 0x%x\n\n", driveBits);
+
+
+    // Reserve buffer long enough for all possible drive strings plus null-terminator
+    char buffer[26 * 4 + 1];
+    // IMPORTANT: The size passed to GetLogicalDriveStringsA is WITHOUT the null-terminator, even though it gets written
+    DWORD charsWritten = GetLogicalDriveStringsA(sizeof(buffer)-1, buffer);
+    if (charsWritten == 0) {
```
JayFoxRox: nit: I think you should check for: `charsWritten == sizeof(buffer)-1` to also cover:
> If the buffer is not large enough, the return value is greater than `nBufferLength`.

I'd at least expect a warning in that case.

I know it won't happen here because our buffer is large enough, but this looks like a typical issue a user would make to me.

lib/winapi/filemanip.c
```diff
@@ -235,3 +239,80 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+
+        if (status == STATUS_NO_MORE_ENTRIES) {
+            break;
+        } else if (!NT_SUCCESS(status)) {
+            NtClose(handle);
+            SetLastError(RtlNtStatusToDosError(status));
+            return 0;
+        }
+
+        // Ignore entries like "CdRom0:"
```
JayFoxRox: nit: I'd change this comment to `// Only allow entries like "C:" or "E:"` (as this is what the code does)

lib/winapi/filemanip.c
thrimbor: I had to check whether we can ignore entries when it returns `STATUS_BUFFER_TOO_SMALL`, turns out we can, so the buffer is now only 2 bytes.

lib/winapi/filemanip.c
thrimbor: I tried to keep the scope of `index` small, but it's changed now.

lib/winapi/filemanip.c
thrimbor: Yes.

samples/winapi_drivelist/main.c
thrimbor: Changed.

samples/winapi_drivelist/main.c
thrimbor: This code is tested, it does work as expected. The while loop stops when `drive` points to the byte after the `\0`.

samples/winapi_drivelist/main.c
thrimbor: I added a check for `charsWritten > sizeof(buffer) - 1`, which I assume is what you meant? `charsWritten == sizeof(buffer)-1` would only be the case if the buffer gets completely filled, in the sample we get a much smaller value for `charsWritten`.

lib/winapi/filemanip.c
thrimbor: I changed the comment, but kept "CdRom0:" because I think it helps to understand why we're doing this.

samples/winapi_drivelist/main.c
JayFoxRox: Oh, right. I'm an idiot. The '\\' test will still succeed, so it will fail on `\0` and go to the next letter.

lib/winapi/filemanip.c
```diff
@@ -235,3 +240,84 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+        } else if (status == STATUS_BUFFER_TOO_SMALL) {
+            // If the object name is too long, we're not interested in it anyway
+            continue;
+        } else if (!NT_SUCCESS(status)) {
+            NtClose(handle);
+            SetLastError(RtlNtStatusToDosError(status));
+            return 0;
+        }
+
+        // Only allow entries like "C:\" or "E:\", ignore stuff like "CdRom0:"
```
JayFoxRox: nit: Comment is wrong still, because it now has the backspace, which wouldn't fit into 2 characters which are being checked for. This might be my mistake (not sure what my review comment said).

thrimbor: Oh, right. Fixed.

samples/winapi_drivelist/main.c
```diff
@@ -0,0 +1,66 @@
...
+        // Additional error info can be retrieved with GetLastError()
+        debugPrint("Failed to retrieve byte strings!\n");
```
JayFoxRox: nit: Was this meant to say "drive strings"?

thrimbor: It was. Fixed.

JayFoxRox:
I believe code looks good now. Waiting for test results (and / or reviews) by someone independent.

thrimbor:
I just rebased this onto master to fix a merge conflict caused by the #307 merge.

JayFoxRox:
Code still looks fine. Waiting for independent test but will merge untested soon if no test or further review happens.

JayFoxRox:
Merging untested. I skimmed over the code and it looks fine.

PR #306, opened 2020-03-13
Deprecate and remove uses of XCreateThread

thrimbor:
This marks `XCreateThread` as deprecated and replaces its use in the network code with `CreateThread`. Tested on my Xbox with the httpd servers.

### DIFF
--- a/lib/net/nforceif/src/sys_arch.c
+++ b/lib/net/nforceif/src/sys_arch.c
@@ -70,7 +70,7 @@ struct sys_sem {
 
 struct sys_thread {
   struct sys_thread *next;
-  int handle;
+  HANDLE handle;
 };
 #endif /* !NO_SYS */
 
@@ -109,7 +109,6 @@ sys_thread_new(const char *name, lwip_thread_fn function, void *arg, int stacksi
 {
   struct sys_thread *st = NULL;
   LWIP_UNUSED_ARG(name);
-  LWIP_UNUSED_ARG(stacksize);
   LWIP_UNUSED_ARG(prio);
 
   st = (struct sys_thread *)malloc(sizeof(struct sys_thread));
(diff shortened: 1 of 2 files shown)


lib/net/nforceif/src/sys_arch.c
```diff
@@ -118,7 +117,7 @@ sys_thread_new(const char *name, lwip_thread_fn function, void *arg, int stacksi
     abort();
   }
 
-  st->handle = XCreateThread((void *)function, arg, NULL);
+  st->handle = CreateThread(NULL, stacksize, (void *)function, arg, 0, NULL);
```
JayFoxRox: Considering this is a driver, I wonder if we should use kernel functions directly instead.
I'm not sure if this might affect things like callbacks (done on behalf of these thread) in which people might query thread information.

thrimbor: I did consider that, but I'm not sure whether that thread may touch TLS-stuff (like `errno`, which became thread local in upstream PDCLib), so I decided to go the safer route.

JayFoxRox: Fair enough; ideally none of this should touch `errno`. We might want to look for a better solution in the future.

JayFoxRox: `sys_thread.handle` is still an `int` but it should be `HANDLE` now.

dracc:
Tested with NevolutionX ftpserver branch. Works as expected. :+1:

lib/net/nforceif/src/sys_arch.c
thrimbor: Oops, missed that. Fixed.

JayFoxRox:
Code looks good, so I merged this without further testing (trusting @dracc with his report).

PR #307, opened 2020-03-17
Implement GetDiskFreeSpaceExA() / GetDiskFreeSpaceA()

thrimbor:
This implements `GetDiskFreeSpaceExA()` and `GetDiskFreeSpaceA()`.
I checked the results of both functions against the values reported by the dashboard on my Xbox, and they do match. The error reporting (both `SetLastError(ERROR_PATH_NOT_FOUND);` and returning `0xFFFFFFFF` in `GetDiskFreeSpaceA()`) were checked against the behavior of Windows 7.

### DIFF
--- a/lib/winapi/filemanip.c
+++ b/lib/winapi/filemanip.c
@@ -288,3 +288,54 @@ BOOL GetDiskFreeSpaceExA (LPCSTR lpDirectoryName, PULARGE_INTEGER lpFreeBytesAva
 
     return TRUE;
 }
+
+BOOL GetDiskFreeSpaceA (LPCSTR lpRootPathName, LPDWORD lpSectorsPerCluster, LPDWORD lpBytesPerSector, LPDWORD lpNumberOfFreeClusters, LPDWORD lpTotalNumberOfClusters)
+{
+    NTSTATUS status;
+    HANDLE handle;
+    ANSI_STRING path;
+    OBJECT_ATTRIBUTES objectAttributes;
+    IO_STATUS_BLOCK ioStatusBlock;
+    FILE_FS_SIZE_INFORMATION fsSizeInfo;
+
+    assert(lpRootPathName);
+
+    RtlInitAnsiString(&path, lpRootPathName);
+    InitializeObjectAttributes(&objectAttributes, &path, OBJ_CASE_INSENSITIVE, ObDosDevicesDirectory(), NULL);
+
+    status = NtOpenFile(&handle, FILE_LIST_DIRECTORY | SYNCHRONIZE, &objectAttributes, &ioStatusBlock, FILE_SHARE_READ | FILE_SHARE_WRITE, FILE_SYNCHRONOUS_IO_NONALERT | FILE_DIRECTORY_FILE | FILE_OPEN_FOR_FREE_SPACE_QUERY);
+    if (!NT_SUCCESS(status)) {
+        SetLastError(RtlNtStatusToDosError(status));
+        if (GetLastError() == ERROR_FILE_NOT_FOUND) {
+            SetLastError(ERROR_PATH_NOT_FOUND);
+        }
+        return FALSE;
+    }
+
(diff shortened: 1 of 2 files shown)


lib/winapi/filemanip.c
```diff
@@ -235,3 +236,106 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+        *lpSectorsPerCluster = fsSizeInfo.SectorsPerAllocationUnit;
+    }
+    if (lpBytesPerSector) {
+        *lpBytesPerSector = fsSizeInfo.BytesPerSector;
+    }
+    if (lpNumberOfFreeClusters) {
+        *lpNumberOfFreeClusters = fsSizeInfo.AvailableAllocationUnits.HighPart == 0 ? fsSizeInfo.AvailableAllocationUnits.LowPart : 0xFFFFFFFF;
+    }
+    if (lpTotalNumberOfClusters) {
+        *lpTotalNumberOfClusters = fsSizeInfo.TotalAllocationUnits.HighPart == 0 ? fsSizeInfo.TotalAllocationUnits.LowPart : 0xFFFFFFFF;
```
JayFoxRox: nit: `fsSizeInfo.TotalAllocationUnits.HighPart ? 0xFFFFFFFF : fsSizeInfo.TotalAllocationUnits.LowPart;`

This helps with readability in my opinion (as the connection between highpart and maxing out the value is more apparent). It's also shorter and doesn't add more operations into the ternary operator (which users often struggle with anyway).
Also matches more closely with the implicit `!= NULL` in the assert for the pathname above.

*(Other instance, too)*

JayFoxRox:
The code looks good (as usual). Waiting for people to test this (or add their own reviews).

lib/winapi/filemanip.c
thrimbor: Good point, I changed it.

dracc:
Tested, both `GetDiskFreeSpace` and `GetDiskFreeSpaceEx` confirmed working on my 1.6.
Returns the expected amount of free storage for the controlled folder. (`F:\Games\`)

Code also looks good. :+1:

lib/winapi/filemanip.c
```diff
@@ -235,3 +236,106 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+BOOL GetDiskFreeSpaceA (LPCSTR lpRootPathName, LPDWORD lpSectorsPerCluster, LPDWORD lpBytesPerSector, LPDWORD lpNumberOfFreeClusters, LPDWORD lpTotalNumberOfClusters)
+{
+    NTSTATUS status;
+    HANDLE handle;
+    ANSI_STRING path;
+    OBJECT_ATTRIBUTES objectAttributes;
+    IO_STATUS_BLOCK ioStatusBlock;
+    FILE_FS_SIZE_INFORMATION fsSizeInfo;
+
+    assert(lpRootPathName);
```
dracc: "If this parameter is NULL, the function uses the root of the current disk."
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getdiskfreespacea

Same applies to `GetDiskFreeSpaceEx`.

thrimbor: True, but we don't support the concept of current disks / current directories yet.

dracc: Makes sense.
I personally wouldn't want the MS docs behavior anyways tbh.

JayFoxRox: We probably do want to support it in the future, so we can set the current directory to `D:\` and then we can probably use relative paths again?

If we don't follow MS docs we'll usually document it. However, in this case we can make an exception (as there's no way to set the current folder so it's obviously problematic). Also the `assert` explicitly handles it.
I guess ideally we'd add the sentence from the MSDN to that `assert` so if people trigger it in a naive port they'll immediately know what's going on.

*(I had actually noticed this in my initial review, too but was fine with this)*

dracc:
Looks good to me, test results turned out as expected. :muscle:

PR #309, opened 2020-03-28
xboxkrnl: Fix calling convention of all callbacks

thrimbor:
I encountered this today when experimenting with using an APC to handle packet reception, where I got weird return value issues because the calling convention of my APC routine (which followed the kernel headers) was wrong.
I tested the changes to the APC-routines, and checked the other prototypes against ReactOS where possible.

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -963,7 +963,7 @@ typedef struct _OBJECT_TYPE
     ULONG PoolTag;
 } OBJECT_TYPE, *POBJECT_TYPE;
 
-typedef NTAPI VOID (*PKDEFERRED_ROUTINE) (
+typedef VOID (NTAPI *PKDEFERRED_ROUTINE) (
     IN PKDPC Dpc,
     IN PVOID DeferredContext,
     IN PVOID SystemArgument1,
@@ -978,7 +978,7 @@ typedef struct _KDEVICE_QUEUE
     LIST_ENTRY DeviceListHead;
 } KDEVICE_QUEUE, *PKDEVICE_QUEUE;
 
-typedef VOID (*PKSTART_ROUTINE) (
+typedef VOID (NTAPI *PKSTART_ROUTINE) (
     IN PVOID StartContext
 );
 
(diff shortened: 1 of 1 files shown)


RadWolfie:
NTAPI isn't in the right place. Any calling type must be after the return type. For example:
```cpp
typedef VOID (NTAPI *PKSTART_ROUTINE) (
    IN PVOID StartContext
);
VOID NTAPI PKSTART_ROUTINE(
    IN PVOID StartContext
);
```

Here's some references:
https://docs.microsoft.com/en-us/cpp/cpp/stdcall?view=vs-2019
https://github.com/reactos/reactos/blob/cb5aa7bee44f57306d6c093bd5c93395ab7db7b7/sdk/include/psdk/winbase.h#L1405-L1418
https://github.com/reactos/reactos/blob/cb5aa7bee44f57306d6c093bd5c93395ab7db7b7/sdk/include/psdk/winbase.h#L1460-L1463

I can't find in GCC's doc for this specific location from [Function-Attributes](https://gcc.gnu.org/onlinedocs/gcc/Function-Attributes.html#Function-Attributes) page.

JayFoxRox:
@RadWolfie Agreed.

Note that I also added it in this place some time ago in https://github.com/XboxDev/nxdk/pull/48/commits/5722e0fd0314ecf40aab20237e52f22e61ca4f1b - it worked fine there. However, we should adapt this standard as documented in MSDN. Our existing xboxkrnl.h file is inconsistent anyway, so for new code we should probably add it in the "right" place to reduce amount of necessary cleanup.

thrimbor:
We've used the `typedef NTAPI [...]` syntax quite a bit in the kernel headers, and it did have the intended effect on clang (i.e. you get a warning when you pass a pointer to a cdecl function) - the syntax described in the msdn seems to only be required in MSVC.
However, I just amended the commit to use the MS-syntax everywhere.

RadWolfie:
@thrimbor What do you mean about intended effect on clang? Perhaps it wasn't properly declared `cdecl` attribute function?

Another reference from clang:
https://github.com/llvm-mirror/clang/blob/master/test/Sema/callingconv.c

P.S. I don't know what Sema is or what type of test it supposedly do. Heh

thrimbor:
> @thrimbor What do you mean about intended effect on clang? Perhaps it wasn't properly declared `cdecl` attribute function?

No, what I mean is that it doesn't matter to clang where you specify the calling convention. When using the syntax we had before, it would correctly warn you when passing a cdecl function, like it should.

PR #310, opened 2020-03-29
ci: Use shallow cloning

thrimbor:
This changes the GH workflow to do shallow clones of git repos, which reduces the amount of traffic we cause on the build VM and the cloning time a bit.

### DIFF
--- a/.github/workflows/build_samples.yml
+++ b/.github/workflows/build_samples.yml
@@ -11,6 +11,7 @@ jobs:
       - name: Clone Tree
         uses: actions/checkout@v1
         with:
+          fetch-depth: 1
           submodules: recursive
       - name: Install MSYS2 & Dependencies
         run: |
@@ -33,6 +34,7 @@ jobs:
       - name: Clone Tree
         uses: actions/checkout@v1
         with:
+          fetch-depth: 1
           submodules: recursive
       - name: Install Dependencies
         run: |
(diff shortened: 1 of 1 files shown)


dracc:
Looks good to me. :+1:

JayFoxRox:
Doesn't seem to offer any performance benefit , but change still looks good to me.

For those wondering, this GitHub Actions feature is documented here: https://github.com/actions/checkout/blob/v1/action.yml (the v1 README doesn't mention `fetch-depth` but the actions.yml does).

---

On the topic of shallow-merging:

I had also experimented with https://www.git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtshallow a while ago but it was prevented by the lwIP repository not supporting shallow clones (which has been resolved in #308). Did you consider that @thrimbor ?

Alternatively (or additionally), we should improve the documentation on the wiki to shallow clone explicitly (to avoid wasting peoples bandwidth and storage space).

thrimbor:
The performance difference is rather small, but usually I saw it requiring a few seconds less for cloning. There's a lot of variance in the build times though (network/IO/CPU load factors etc.), so it's not very visible.

> On the topic of shallow-merging:

@JayFoxRox I didn't think about that yet. I'm not sure how submodule behavior changes when shallow cloning them - on the other hand, how likely is it that "normal users" ever fiddle with the submodules?
Mentioning this possibility in the wiki seems like a good idea anyway, though.

JayFoxRox:
Merging.

I don't think it will cause any issues. I'll update the wiki with instructions to shallow-clone later (Remind me if I don't).

PR #312, opened 2020-04-08
sdl2: Update submodule to ed7840a

thrimbor:
This updates our SDL submodule to include the recently merged audio work (https://github.com/XboxDev/nxdk-sdl/pull/20).


JayFoxRox:
Note: SDL Audio is still problematic as described in https://github.com/XboxDev/nxdk-sdl/issues/26 via pinned issue https://github.com/XboxDev/nxdk/issues/314.

However, we'll still merge it; just don't expect too much (yet). We still recommend to plan with SDL Audio as stable audio API for the next years.

PR #313, opened 2020-04-08
pdclib: Update submodule to 9101a06

thrimbor:
Updates the PDCLib submodule to include all changes done in this year so far:
* errno values from Win32
* `isfinite()` C++ syntax error fix
* LTO fix for `assert()`
* LTO fix for `_tls_index`
* epoch fix for `time()`
* crash fix for `fwrite()`


JayFoxRox:
Tested compilation and that still works. Merging.

PR #319, opened 2020-04-11
Add SDL_image and libjpeg-turbo

### LINKED ISSUE #164: libjpeg port
I have prototyped a libjpeg (specifically libjpeg-turbo) port for https://github.com/libjpeg-turbo/libjpeg-turbo in https://github.com/JayFoxRox/neverball/blob/xbox/xbox/Makefile-libjpeg-turbo.xbox

Someone should clean it up, and upstream it; we can probably get around having a fork by doing configuration out-of-tree.

### LINKED ISSUE #167: SDL_image port
@SoundAxis has prototyped a SDL_image port in https://github.com/SoundAxis/nxdk/tree/master/lib/sdl/SDL2_image

This would possibly benefit from #163 and #164 by using those existing ports of the latest versions. Someone should check wether this is possible, so we can use an existing SDL_image repository without changes.

If not possible to use existing libjpeg and libpng ports, someone should reconstruct the history of SDL_image to extract nxdk specific patches.

thrimbor:
This adds SDL_image and libjpeg-turbo, along with a small sample that demonstrates SDL_image use by displaying a CC0 jpg image.

Usage of MMX/SSE is possible in libjpeg-turbo, but requires additional work because it uses nasm, so it's not enabled yet.

Closes #164.
Closes #167.


lib/sdl/Makefile
```diff
@@ -40,3 +40,36 @@ clean-sdl_ttf:
 	           $(FREETYPE_OBJS) \
 	           $(NXDK_DIR)/lib/libSDL_ttf.lib \
 	           $(NXDK_DIR)/lib/libfreetype.lib
+
+LIBSDLIMG_DIR = $(NXDK_DIR)/lib/sdl/SDL2_image
```
JayFoxRox, line 44: nit: I'm not a big fan of calling this SDLIMG / the sample sdl_img.
Is this a common alias for SDL_image, other than for their API prefix? This is the first time I see someone calling it like that, and I'm worried it will lead to potential confusion or errors later.

lib/libjpeg/jconfigint.h
```diff
@@ -0,0 +1,31 @@
...
+#define PACKAGE_NAME  "libjpeg-turbo"
+
+/* Version number of package */
+#define VERSION  "2.0.4"
+
+/* The size of `size_t', as computed by sizeof. */
+#define SIZEOF_SIZE_T  4
+
+/* Define if your compiler has __builtin_ctzl() and sizeof(unsigned long) == sizeof(size_t). */
+#undef HAVE_BUILTIN_CTZL
```
JayFoxRox, line 20: nit: Why is this done? I'd assume this is a potential optimization that we could support?

samples/sdl_img/Makefile
```diff
@@ -0,0 +1,10 @@
+XBE_TITLE = sdl_img
+GEN_XISO = $(XBE_TITLE).iso
+SRCS += $(wildcard $(CURDIR)/*.c)
+NXDK_DIR = $(CURDIR)/../..
+NXDK_SDL = y
+include $(NXDK_DIR)/Makefile
+
+$(XBE_TITLE).iso: $(OUTPUT_DIR)/testimg.jpg
```
JayFoxRox, line 8: This is an interesting way of getting the files into the ISO. However, shouldn't this use `$(GEN_XISO)`?

We also do this differently in the SDL_ttf sample (and maybe elsewhere).

lib/sdl/Makefile
thrimbor: I've renamed the sample and the variables.

lib/libjpeg/jconfigint.h
thrimbor: Yes, that should've been enabled.

samples/sdl_img/Makefile
thrimbor: > This is an interesting way of getting the files into the ISO. However, shouldn't this use `$(GEN_XISO)`?

That's probably better, yes. I changed it.
> We also do this differently in the SDL_ttf sample (and maybe elsewhere).

Afaik SDL_ttf is the only other sample where we need additional files in the ISO. I wasn't entirely happy with the solution user there, though - it doesn't properly use dependencies, so it won't rebuild the ISO when the file changes, and it looks like it might cause race conditions, too (ISO creation might run parallel to the cp_font rule).

lib/libjpeg/jconfigint.h
```diff
@@ -0,0 +1,31 @@
...
+#define INLINE  __inline__ __attribute__((always_inline))
+
+/* Define to the full name of this package. */
+#define PACKAGE_NAME  "libjpeg-turbo"
+
+/* Version number of package */
+#define VERSION  "2.0.4"
+
+/* The size of `size_t', as computed by sizeof. */
+#define SIZEOF_SIZE_T  4
```
JayFoxRox, line 17: nit: I'm not sure why this isn't just `sizeof(size_t)`; but this is fine, too.

lib/libjpeg/jconfigint.h
```diff
@@ -0,0 +1,31 @@
+/* libjpeg-turbo build number */
+#define BUILD  "20191111"
+
+/* Compiler's inline keyword */
+#undef inline
```
JayFoxRox, line 5: What is this for? This looks.. weird?
It implies that someone previously defined `inline` (which should never happen).

lib/libjpeg/Makefile
```diff
@@ -0,0 +1,70 @@
...
+                     $(LIBJPEG_TURBO_DIR)/jquant1.c \
+                     $(LIBJPEG_TURBO_DIR)/jquant2.c \
+                     $(LIBJPEG_TURBO_DIR)/jutils.c \
+                     $(LIBJPEG_TURBO_DIR)/jmemmgr.c \
+                     $(LIBJPEG_TURBO_DIR)/jmemnobs.c \
+                     $(LIBJPEG_TURBO_DIR)/jaricom.c \
+                     $(LIBJPEG_TURBO_DIR)/jcarith.c \
+                     $(LIBJPEG_TURBO_DIR)/jdarith.c
+
+# FIXME: MMX/SSE is currently disabled due to requiring nasm
```
JayFoxRox, line 55: Please create an issue about this after merge

lib/libjpeg/jconfigint.h
thrimbor: No idea. It's part of the template from which `jconfigint.h` gets auto generated (same above).

samples/sdl_img/Makefile
JayFoxRox: Please create an issue about this. It should be more consistent in the future. I feel like we have to improve this.

*(I also hope we'll move the data files into a folder like "samples/common/data" eventually, to avoid duplication of things like audio files for OpenAL, SDL Audio, xaudio, ... samples; or images for SDL_image, libpng, ... samples; or fonts in general)*

lib/libjpeg/jconfigint.h
JayFoxRox: Fair enough

lib/libjpeg/jconfigint.h
JayFoxRox: Response in https://github.com/XboxDev/nxdk/pull/319#discussion_r408102734

lib/libjpeg/Makefile
```diff
@@ -0,0 +1,70 @@
...
+LIBJPEG_TURBO_SRCS += $(LIBJPEG_TURBO_DIR)/jsimd_none.c
+
+LIBJPEG_TURBO_OBJS = $(addsuffix .obj, $(basename $(LIBJPEG_TURBO_SRCS)))
+
+NXDK_CFLAGS += -isystem $(LIBJPEG_TURBO_DIR) -I$(NXDK_DIR)/lib/libjpeg
```
JayFoxRox, line 60: This will probably cause the usual issues of only making it work with C, but not C++.

lib/sdl/Makefile
```diff
@@ -40,3 +40,36 @@ clean-sdl_ttf:
 	           $(FREETYPE_OBJS) \
 	           $(NXDK_DIR)/lib/libSDL_ttf.lib \
 	           $(NXDK_DIR)/lib/libfreetype.lib
+
+LIBSDLIMAGE_DIR = $(NXDK_DIR)/lib/sdl/SDL2_image
+
+LIBSDLIMAGE_SRCS = $(LIBSDLIMAGE_DIR)/IMG.c \
+                 $(LIBSDLIMAGE_DIR)/IMG_png.c \
```
JayFoxRox, line 47: nit: Indentation is a bit weird?

samples/sdl_img/Makefile
JayFoxRox: (Handled in issue 322)

JayFoxRox:
I've tested this and it works fine. Please consider my comments and reply if you intend to change these things.

I could nitpick some more about the sample quality to be honest. I feel like we are gradually having less consistent code style and weird licenses (mostly by including samples from elsewhere).
We also currently don't document where our artwork comes from or what the license is (I know this image is CC0 according to you, but this isn't documented elsewhere). We should probably create an issue about re-organizing this and create a style-guide for samples.

Anyhow, I'll merge this within the next days if there is no veto.

.gitmodules
```diff
@@ -25,3 +25,6 @@
 [submodule "lib/libpng/libpng"]
 	path = lib/libpng/libpng
 	url = https://github.com/glennrp/libpng.git
+[submodule "lib/libjpeg/libjpeg-turbo"]
+	path = lib/libjpeg/libjpeg-turbo
+	url = https://github.com/libjpeg-turbo/libjpeg-turbo.git
```
JayFoxRox, line 30: I just realized that the commits which add the packages have no prefix in their names. I'd expect them to have a prefix like "libjpeg:" and "sdl:" / "sdl_image:" / ...

Also see earlier commits:
- https://github.com/XboxDev/nxdk/commit/632fe50b4017c523d9598a7d2782748e7a4ee227
- https://github.com/XboxDev/nxdk/commit/c40c34bc61b75ed2f5da01a9296be1f1326d6c5e

lib/libjpeg/Makefile
thrimbor: I added the `NXDK_CXXFLAGS` line, should be fixed now.

lib/sdl/Makefile
thrimbor: Fixed.

.gitmodules
thrimbor: I added a prefix to the two commits now.

JayFoxRox:
Thanks @thrimbor ; please create the issue mentioned in https://github.com/XboxDev/nxdk/pull/319#discussion_r408100517

PR #320, opened 2020-04-13
libc: Implement case-insensitive string comparison functions

thrimbor:
This adds `_stricmp`, `stricmp`, `_strnicmp` and `strnicmp` as libc extensions. This hopefully eliminates the need for some workarounds, such as in @Ryzee119's SDLPoP port.

The code is copied from PDCLib's normal comparison functions, but adjusted to behave like the MSDN describes. The only difference is that it doesn't invoke an invalid parameter handler routine when encountering a null pointer, instead it behaves as if the invalid parameter handler routine allowed to continue execution. I added asserts to still catch those errors in debug builds.

Interestingly, ReactOS's version of these functions does not do that at all, and also uses `toupper` instead of `tolower` (the latter being what the MSDN describes).

### DIFF
--- a/lib/xboxrt/libc_extensions/string_ext_.h
+++ b/lib/xboxrt/libc_extensions/string_ext_.h
@@ -9,6 +9,27 @@ static char *_strdup (const char *s)
     return strdup(s);
 }
 
+// Compared to their Win32 counterparts, these functions currently don't invoke
+// an invalid parameter handler routine as described in the MSDN, and instead
+// behave as if the invalid parameter handler allowed to continue execution.
+int _strnicmp (const char *s1, const char *s2, size_t n);
+int _stricmp (const char *s1, const char *s2);
+
+__attribute__((deprecated)) static int strnicmp (const char *s1, const char *s2, size_t n)
+{
+    return _strnicmp(s1, s2, n);
+}
+
+__attribute__((deprecated)) static int stricmp (const char *s1, const char *s2)
+{
+    return _stricmp(s1, s2);
+}
+
+#ifndef NLSCMP_DEFINED
+#define _NLSCMPERROR 0x7FFFFFFF
+#define _NLSCMP_DEFINED
+#endif
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/xboxrt/libc_extensions/string_ext_.c
+++ b/lib/xboxrt/libc_extensions/string_ext_.c
@@ -1,3 +1,6 @@
+#include <assert.h>
+#include <errno.h>
+#include <ctype.h>
 #include <string.h>
 #include <stdlib.h>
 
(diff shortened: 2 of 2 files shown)


lib/xboxrt/libc_extensions/string_ext_.h
```diff
@@ -9,6 +9,27 @@ static char *_strdup (const char *s)
...
+    return _strnicmp(s1, s2, n);
+}
+
+static int stricmp (const char *s1, const char *s2)
+{
+    return _stricmp(s1, s2);
+}
+
+#ifndef NLSCMP_DEFINED
+#define _NLSCMPERROR 2147483647
```
JayFoxRox, line 29: nit: `0x7FFFFFFF` for readability?

lib/xboxrt/libc_extensions/string_ext_.h
```diff
@@ -9,6 +9,27 @@ static char *_strdup (const char *s)
     return strdup(s);
 }
 
+// Compared to their Win32 counterparts, these functions currently don't invoke
+// an invalid parameter handler routine as described in the MSDN, and instead
+// behave is if the invalid parameter handler allowed to continue execution.
```
JayFoxRox, line 14: > [...] behave is if [...]

lib/xboxrt/libc_extensions/string_ext_.h
thrimbor: Oh yeah, that's probably better. Changed.

lib/xboxrt/libc_extensions/string_ext_.h
thrimbor: Fixed.

lib/xboxrt/libc_extensions/string_ext_.h
```diff
@@ -9,6 +9,27 @@ static char *_strdup (const char *s)
     return strdup(s);
 }
 
+// Compared to their Win32 counterparts, these functions currently don't invoke
+// an invalid parameter handler routine as described in the MSDN, and instead
+// behave as if the invalid parameter handler allowed to continue execution.
+int _strnicmp (const char *s1, const char *s2, size_t n);
+int _stricmp (const char *s1, const char *s2);
+
+static int strnicmp (const char *s1, const char *s2, size_t n)
```
JayFoxRox, line 18: These should be deprecated; [see MSDN](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strnicmp-wcsnicmp?view=vs-2019)

> The Microsoft-specific function names strnicmp and wcsnicmp are deprecated aliases for the _strnicmp and _wcsnicmp functions. By default, they generate Compiler warning (level 3) C4996.

Then [see the description for C4996](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=vs-2019):

> Your code uses a function, class member, variable, or typedef that's marked deprecated. Symbols are deprecated by using a __declspec(deprecated) modifier, or the C++14 [[deprecated]] attribute.

thrimbor: Done.

PR #325, opened 2020-04-26
Build winapi and xboxrt as libraries

JayFoxRox:
For using CMake or other build systems, it is more suitable to have these "modules" as actual libraries.

### DIFF
--- a/lib/winapi/Makefile
+++ b/lib/winapi/Makefile
@@ -1,0 +1,27 @@
+WINAPI_SRCS := \
+	$(NXDK_DIR)/lib/winapi/debug.c \
+	$(NXDK_DIR)/lib/winapi/errhandlingapi.c \
+	$(NXDK_DIR)/lib/winapi/error.c \
+	$(NXDK_DIR)/lib/winapi/fiber.c \
+	$(NXDK_DIR)/lib/winapi/fileio.c \
+	$(NXDK_DIR)/lib/winapi/filemanip.c \
+	$(NXDK_DIR)/lib/winapi/findfile.c \
+	$(NXDK_DIR)/lib/winapi/handleapi.c \
+	$(NXDK_DIR)/lib/winapi/memory.c \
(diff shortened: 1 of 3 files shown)


lib/winapi/Makefile
```diff
@@ -0,0 +1,14 @@
+WINAPI_SRCS := \
+	$(wildcard $(NXDK_DIR)/lib/winapi/*.c)
+
+WINAPI_OBJS = $(addsuffix .obj, $(basename $(WINAPI_SRCS)))
+
+$(NXDK_DIR)/lib/libwinapi.lib: $(WINAPI_OBJS)
```
JayFoxRox, line 6: This should probably be split into user32.lib, kernel32.lib etc., so that `#pragma` and build-systems can refer to the actual libs.
However, it will need some more reorganization, so this is fine for now. I'll create an issue after merge.

lib/winapi/Makefile
```diff
@@ -0,0 +1,14 @@
+WINAPI_SRCS := \
+	$(wildcard $(NXDK_DIR)/lib/winapi/*.c)
```
thrimbor, line 2: I think we should take this opportunity to get rid of the wildcard and list the files explicitly.

lib/winapi/Makefile
thrimbor: I agree, this can be done in a later PR.
`#pragma` also is not compatible with the way we currently track dependencies, so some deeper changes to our build system are probably required.

lib/winapi/Makefile
JayFoxRox: I did the suggested change.

*I think we could have sorted files when creating user32.lib / kernel32.lib / .. to reduce git diff size. But doing it now is probably good to avoid errors (with unwanted files in winapi.lib), too. So personally it's fine with me eitherway.*

thrimbor:
Looks good now, merging :+1:

PR #327, opened 2020-04-30
Spell "nxdk" in lowercase

JayFoxRox:
This makes the spelling of "nxdk" a bit more consistent.

We have been going back and forth with "NXDK" and "nxdk", but most instances are "nxdk" (in lowercase) now.

Some instances of uppercase spellings might still exist in code comments / submodules. We can fix those later / when we touch such code.

`NXDK` remains a preprocessor macro for finding out if compilation is happening with nxdk.


thrimbor:
Looks good, merging

PR #329, opened 2020-05-01
libc: Provide tchar.h

thrimbor:
This adds `tchar.h`, as a preparation to reduce the diff to upstream for the next PDCLib rebase (it's included by dlmalloc).

PDCLib actually builds fine with an empty header, but I decided to provide at least some functionality. I'm not too worried about it being complete or correct - it's not very likely that we'll encounter code that requires it, so I consider it to be more of a bonus that we can improve when required.

### DIFF
--- a/lib/xboxrt/libc_extensions/tchar.h
+++ b/lib/xboxrt/libc_extensions/tchar.h
@@ -1,0 +1,116 @@
+#ifndef _XBOXRT_TCHAR_H
+#define _XBOXRT_TCHAR_H
+
+/**
+ * @file tchar.h
+ * @author Stefan Schmidt
+ * This specific file is licensed under the CC0 1.0.
+ * See https://creativecommons.org/publicdomain/zero/1.0/ for details.
+ *
+ * tchar.h is an MS extension to provide mappings for building applications that use
+ * multi-byte character strings, Unicode or single-byte character strings, from
+ * the same sourcecode.
+ * This provides a (not necessarily complete) implementation for nxdk to
+ * hopefully help with porting Windows-specific software.
+ * For more information, see the following link:
+ * https://docs.microsoft.com/en-us/cpp/c-runtime-library/generic-text-mappings?view=vs-2019
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#if defined(UNICODE) || defined(_MBCS)
+#error nxdk does not support the Unicode API
+#endif
+
+#define _TEOF EOF
+
+#ifndef __TCHAR_DEFINED
+#define __TCHAR_DEFINED
+typedef char _TCHAR;
+typedef int _TINT;
+typedef signed char _TSCHAR;
+typedef unsigned char _TUCHAR;
+typedef char _TXCHAR;
+#endif
+
+#define __T(x) x
+#define _T __T
+#define _TEXT __T
+
+#define _tmain main
+
+#define _tprintf printf
+#define _ftprintf fprintf
+#define _stprintf sprintf
+#define _sntprintf _snprintf
+#define _vtprintf vprintf
+#define _vftprintf vfprintf
+#define _vstprintf vsprintf
+#define _vsntprintf vsnprintf
+#define _tscanf scanf
+#define _ftscanf fscanf
+#define _stscanf sscanf
+
+#define _fgettc fgetc
+#define _fgetts fgets
+#define _fputtc fputc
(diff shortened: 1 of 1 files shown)


lib/xboxrt/libc_extensions/tchar.h
```diff
@@ -0,0 +1,101 @@
...
+typedef char _TCHAR;
+typedef int _TINT;
+typedef signed char _TSCHAR;
+typedef unsigned char _TUCHAR;
+typedef char _TXCHAR;
+#endif
+
+#define __T(x) x
+#define _T __T
+#define _TEXT __T
```
JayFoxRox, line 25: nit: Does MS also use `__T` for this step? I'm a bit worried about side-effects (overwriting `__T` / defining a very generic name like `__T` that might conflict with some lib).

If MS doesn't use this name, then we could `#undef __T` afterwards; but it still might have side-effects (still ruining `__T`, so include order is relevant now).

lib/xboxrt/libc_extensions/tchar.h
```diff
@@ -0,0 +1,101 @@
...
+#define _istupper isupper
+#define _totupper toupper
+#define _totlower tolower
+#define _istlegal(c) (1)
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
```
JayFoxRox, line 101: nit: Newline missing at end of file. not sure if we usually have this; but I'm sure that my editor will complain / add it.

JayFoxRox:
I have read all defines and made sure their names make sense.

Looks good to me aside from my nits (which are mostly questions).

Another nit: I feel there could be more comments in the header about what tchar is even for / who made the extension / where to find information about it. I belive I have seen `_T("Foo")` and `_TEXT("Foo")` before, but I was initially confused by what tchar.h even is.

And final nit: I think your PR description should explain where this code came from / what your source is for implementing this. During review I was looking at https://docs.microsoft.com/en-us/cpp/text/generic-text-mappings-in-tchar-h?view=vs-2019 but that didn't have `_istlegal` and `_tmain`. It took me a while to also find: https://docs.microsoft.com/en-us/cpp/cpp/main-function-command-line-args?view=vs-2019 and https://docs.microsoft.com/en-us/cpp/c-runtime-library/routine-mappings?view=vs-2019; had I not had this existing code, I'd have had trouble finding more complete information about tchar.h. So I'm slightly worried this is missing some critical tchar feature or caveat that I'm not aware of (example: do MS headers auto-define `SBCS` if "_UNICODE & MBCS not defined"? It's unclear from docs that I have read; if your source is the MS code then this would remove this uncertainty).

Someone should include this once to test compile, but other than that this feels "done".

lib/xboxrt/libc_extensions/tchar.h
thrimbor: The MS header indeed uses `__T`. It's used to do the actual work (i.e. attaching the `L` prefix when `UNICODE` is defined: `#define __T(x) L ## x`), while `_T` and `_TEXT` then just reference it.

lib/xboxrt/libc_extensions/tchar.h
thrimbor: Oops, fixed.

thrimbor:
> Another nit: I feel there could be more comments in the header about what tchar is even for / who made the extension / where to find information about it. I belive I have seen `_T("Foo")` and `_TEXT("Foo")` before, but I was initially confused by what tchar.h even is.

I added a Doxygen-style comment at the beginning.

> And final nit: I think your PR description should explain where this code came from / what your source is for implementing this. During review I was looking at https://docs.microsoft.com/en-us/cpp/text/generic-text-mappings-in-tchar-h?view=vs-2019 but that didn't have `_istlegal` and `_tmain`.

My source mostly is the MSDN - `_tmain` is something I remembered seeing before, so I looked it up. I briefly skimmed the MS tchar.h to check if I missed something major, but I didn't directly copy from it (they use underscore prefixes for some standard library functions etc.).

> So I'm slightly worried this is missing some critical tchar feature or caveat that I'm not aware of (example: do MS headers auto-define `SBCS` if "_UNICODE & MBCS not defined"? It's unclear from docs that I have read; if your source is the MS code then this would remove this uncertainty).

I can't guarantee that not being the case, but I'm not too worried about it either. As for `SBCS`: I just checked, and it doesn't even occur in tchar.h outside of comments.

PR #330, opened 2020-05-02
build: Replacing wildcards with paths

### LINKED ISSUE #67: Finding files using wildcards can lead to unexpected behaviour
It already came up during review of #60: we should stop using wildcards in our Makefiles. Both, for nxdk toolchain files, and in our sample code.

Teufelchen1:
Hi,
closes #67

I replaced all wildcards in all Makefiles I could find. There are still wildcards in these sub-repositories:

- `lib/zlib`
- `lib/pdcib`
- `lib/sdl`
- `lib/net/lwip`

### DIFF
--- a/samples/mesh/Makefile
+++ b/samples/mesh/Makefile
@@ -1,6 +1,6 @@
 XBE_TITLE = mesh
 GEN_XISO = $(XBE_TITLE).iso
-SRCS = $(wildcard $(CURDIR)/*.c)
+SRCS = $(CURDIR)/math3d.c $(CURDIR)/main.c
 SHADER_OBJS = ps.inl vs.inl
 NXDK_DIR = $(CURDIR)/../..
 
--- a/samples/sdl/Makefile
+++ b/samples/sdl/Makefile
@@ -1,6 +1,6 @@
 XBE_TITLE = sdl
 GEN_XISO = $(XBE_TITLE).iso
-SRCS += $(wildcard $(CURDIR)/*.c)
+SRCS = $(CURDIR)/main.c
 NXDK_DIR = $(CURDIR)/../..
 NXDK_SDL = y
 include $(NXDK_DIR)/Makefile
--- a/tools/cxbe/Makefile
+++ b/tools/cxbe/Makefile
@@ -1,5 +1,18 @@
-INCLUDES := $(wildcard *.h)
-SRCS := $(wildcard *.cpp)
+INCLUDES := \
+	AlignPosfix1.h \
+	AlignPrefix1.h \
+	Common.h \
+	Cxbx.h \
+	Error.h \
+	Exe.h \
+	Xbe.h
+SRCS := \
+	Main.cpp \
+	Common.cpp \
+	Error.cpp \
+	Exe.cpp \
+	OpenXDK.cpp \
+	Xbe.cpp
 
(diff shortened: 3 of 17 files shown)


tools/cxbe/Makefile
```diff
@@ -1,5 +1,16 @@
-INCLUDES := $(wildcard *.h)
-SRCS := $(wildcard *.cpp)
+INCLUDES := AlignPosfix1.h \
+	AlignPrefix1.h \
```
JayFoxRox, line 2: nit: Weird indentation. Either start it with an empty `\` in the first line, then align all with a tab; or align with spaces to align to the first line.

tools/fp20compiler/Makefile
```diff
@@ -1,7 +1,15 @@
 MAIN = fp20compiler
 CXXFLAGS= -DYY_NO_UNPUT -Wall -g
 
-INCLUDES = $(wildcard *.h) \
+INCLUDES = nvparse_errors.h \
```
JayFoxRox, line 4: nit: Weird indentation. Start this with an empty first line which just has `\`. Otherwise it's easy to miss "nvparse_errors.h"

(Other instances, too)

JayFoxRox:
nit: You should consider splitting this into separate commits in case we have to revert a change. Personally I wouldn't do it, but it would have been nice maybe (not too detailed, maybe just `pbkit:` `tools` and `samples:`).

nit: I think the usual commit style would be a command-style / future tense (what this will do; not what it is doing): "build: Replace wildcards by explicit path".


Other than that it looked fine during a first review pass. I'll probably build this myself from scratch to make sure nothing was forgotten.

samples/mesh/Makefile
```diff
@@ -1,6 +1,6 @@
 XBE_TITLE = mesh
 GEN_XISO = $(XBE_TITLE).iso
-SRCS = $(wildcard $(CURDIR)/*.c)
+SRCS += $(CURDIR)/math3d.c $(CURDIR)/main.c
```
thrimbor, line 3: Why the change to `+=`?

tools/fp20compiler/Makefile
```diff
@@ -1,12 +1,22 @@
...
+INCLUDES = \
+	nvparse_errors.h \
+	nvparse_externs.h \
+	ps1.0_program.h \
+	rc1.0_combiners.h \
+	rc1.0_final.h \
+	rc1.0_general.h \
+	rc1.0_register.h \
+	ts1.0_inst.h \
+	ts1.0_inst_list.h\
```
thrimbor, line 13: nit: No space before `\`

samples/sdl/Makefile
```diff
@@ -1,6 +1,6 @@
 XBE_TITLE = sdl
 GEN_XISO = $(XBE_TITLE).iso
-SRCS += $(wildcard $(CURDIR)/*.c)
+SRCS += $(CURDIR)/main.c
```
JayFoxRox, line 3: This one used to use `+=`, but I think this should just be `=` to be more consistent with other samples.

The httpd sample is also affected by this.

samples/mesh/Makefile
JayFoxRox: You pushed since this review, but this one isn't fixed yet? @Teufelchen1

thrimbor:
Looks good now, merging.

PR #332, opened 2020-05-04
Implement more WaitFor* functions for winapi

JayFoxRox:

This adds the `WaitForSingleObject` function (simply forwarding to the existing `WaitForSingleObjectEx`).

I also decided to add `WaitForMultipleObjects` and `WaitForMultipleObjectsEx` (based on the existing `WaitForSingleObjectEx`).

None of this code was tested. I assume it works fine. If there are bugs, they likely already existed in the existing `WaitForMultipleObjects` functions or our xboxkrnl header.

### DIFF
--- a/lib/winapi/sync.c
+++ b/lib/winapi/sync.c
@@ -79,6 +79,30 @@ DWORD WaitForSingleObject (HANDLE hHandle, DWORD dwMilliseconds)
     return WaitForSingleObjectEx(hHandle, dwMilliseconds, FALSE);
 }
 
+DWORD WaitForMultipleObjectsEx (DWORD nCount, const HANDLE *lpHandles, BOOL bWaitAll, DWORD dwMilliseconds, BOOL bAlertable)
+{
+    LARGE_INTEGER duration;
+    duration.QuadPart = ((LONGLONG)dwMilliseconds) * -10000;
+
+    while (true) {
+        NTSTATUS status = NtWaitForMultipleObjectsEx(nCount, lpHandles, bWaitAll ? WaitAll : WaitAny, UserMode, bAlertable, &duration);
+
+        if (status == STATUS_ALERTED) continue;
+
+        if (!NT_SUCCESS(status)) {
+            SetLastError(RtlNtStatusToDosError(status));
+            return WAIT_FAILED;
+        }
+
+        return status;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


lib/winapi/sync.c
```diff
@@ -74,6 +74,35 @@ DWORD WaitForSingleObjectEx (HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertab
...
 
+DWORD WaitForSingleObject (HANDLE hHandle, DWORD dwMilliseconds)
+{
+    return WaitForSingleObjectEx(hHandle, dwMilliseconds, FALSE);
+}
+
+DWORD WaitForMultipleObjectsEx (DWORD nCount, const HANDLE *lpHandles, BOOL bWaitAll, DWORD dwMilliseconds, BOOL bAlertable)
+{
+    LARGE_INTEGER duration;
+    duration.QuadPart = ((LONGLONG)dwMilliseconds) * -10000;
```
Teufelchen1, line 85: Just for clarification: Its `-10000` because:
- making it negativ makes it a relative 
- the underlying function take n * 100ns so we got `1ms = 1000µs = 10.000 * 100ns` =>`10000`

I ask because I couldn't find proper documentation on `NtWaitForMultipleObjectsEx()`, at least nothing wich would explain the `-10000`. However, in `SleepEx()`, which looks similar, the used `KeDelayExecutionThread()` is documented [here](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-kedelayexecutionthread) and fits the `-10000`.

JayFoxRox: Agreed that this is not very clear. But already an issue in the existing `WaitForSingleObjectEx`.

It's documented for https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-kewaitformultipleobjects
This should be documented on XboxDevWiki (not recommended anymore; I think the xboxkrnl documentation is slowly being moved into nxdk instead) or in nxdk xboxkrnl.h (recommended).

There's another pitfall here, in that the code is written under the assumption that `duration` will be updated. Disassembly of `KeWaitForSingleObject` suggests this, but MSDN does not document this. Someone should fix this in the MSDN or document this in our ecosystem.

I do not plan to change / document this as part of this PR.

thrimbor:
Code looks good and tested fine. Merging.

PR #333, opened 2020-05-04
winapi: Stub SHGetSpecialFolderPathA

JayFoxRox:

This adds a stub for `SHGetSpecialFolderPathA` which will always fail on Xbox (at least for now).
I also added CSIDL_APPDATA as I found it in mingw source-code (they keep it in shfolder.h), because that's required for OpenAL-soft.

Normally this function is supposed to return a common path (like the users document folder), so a potential implementation would return UDATA / TDATA in the future (it's still a bad fit).

I tried to keep the filenames as named in MSDN - I don't have a Windows SDK to check if this is right.

### DIFF
--- a/lib/winapi/Makefile
+++ b/lib/winapi/Makefile
@@ -9,6 +9,7 @@ WINAPI_SRCS := \
 	$(NXDK_DIR)/lib/winapi/handleapi.c \
 	$(NXDK_DIR)/lib/winapi/memory.c \
 	$(NXDK_DIR)/lib/winapi/profiling.c \
+	$(NXDK_DIR)/lib/winapi/shlobj_core.c \
 	$(NXDK_DIR)/lib/winapi/sync.c \
 	$(NXDK_DIR)/lib/winapi/sysinfo.c \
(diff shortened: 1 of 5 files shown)


thrimbor:
Code looks fine, but I'm a bit unsure whether we really should add this. We'll probably never support most of the features of this function (`CSIDL_PRINTERS`, `CSIDL_SENDTO`, `CSIDL_BITBUCKET` ...), which makes me wonder if changing the code in the port might make more sense.

JayFoxRox:
I'm also torn about this, but I have come across this a couple of times.
It's important when we want to build code out of the box. I'm also considering closed-source libraries where we won't have control over this.

In my opinion, it's okay as a stub that always fails.

However, I also agree that users should primarily port code to avoid these issues in the first place.
If we were to decide to implement this (due to demand), then we could just return folders in the games `UDATA` / `TDATA` folders if it exists, and fail otherwise. We could also break this feature of into a separate lib for different options as part of the build-system.

thrimbor:
I'm still a bit skeptical, but I'll merge it anyway.

PR #334, opened 2020-05-04
winapi: Stub IsBadWritePtr

JayFoxRox:

This is a very weird function, which seems to probe page access rights (similar to `VirtualQuery` or `VirtualProtect`?). I'm not sure how this is implemented on Windows or how we should implement this on Xbox (also think of physical vs. virtual memory APIs on Xbox).

OpenAL seems to use it to check if enough elements in an array exist before trying to resize it.
I decided to stub it for now (and start upstreaming my changes as soon as possible, so they don't pile up). I might have to revisit / implement it in the near future, as I begin testing the port.

I wasn't sure about calling the file BaseTsd.h or basetsd.h - MSDN usually calls it BaseTsd.h, and due to case sensitivity, I decided to go with that.

The implementation is in badptr.c, because more similarly named functions like this exist, to check different permissions. [See "See also" in MSDN](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-isbadwriteptr).

### DIFF
--- a/lib/winapi/badptr.c
+++ b/lib/winapi/badptr.c
@@ -1,0 +1,13 @@
+#include <windows.h>
+
+BOOL IsBadWritePtr (LPVOID lp, UINT_PTR ucb)
+{
+    // Windows appears to implement this by registering an SEH handler.
+    // The Microsoft implementation then probes each page.
+    // For Xbox, this is not suitable because `lp` potentially points at MMIO.
+
+    //FIXME: Walk allocations in region with NtQueryVirtualMemory and check write permission?
+
+    // We disallow all access for now, because memory is potentially unsafe
+    return TRUE;
+}
--- a/lib/winapi/winbase.h
+++ b/lib/winapi/winbase.h
@@ -4,6 +4,7 @@
 #include <windef.h>
 #include <minwinbase.h>
 #include <winnt.h>
+#include <basetsd.h>
 
 #ifdef __cplusplus
 extern "C"
@@ -92,6 +93,8 @@ DWORD GetLastError (void);
 void SetLastError (DWORD error);
 
 void WINAPI OutputDebugStringA (LPCTSTR lpOutputString);
+
+BOOL IsBadWritePtr (LPVOID lp, UINT_PTR ucb);
 
 #ifndef UNICODE
 #define OutputDebugString OutputDebugStringA
--- a/lib/winapi/Makefile
+++ b/lib/winapi/Makefile
@@ -1,4 +1,5 @@
 WINAPI_SRCS := \
+	$(NXDK_DIR)/lib/winapi/badptr.c \
 	$(NXDK_DIR)/lib/winapi/debug.c \
 	$(NXDK_DIR)/lib/winapi/errhandlingapi.c \
 	$(NXDK_DIR)/lib/winapi/error.c \
--- a/lib/winapi/basetsd.h
+++ b/lib/winapi/basetsd.h
@@ -1,0 +1,6 @@
+#ifndef __BASETSD_H__
+#define __BASETSD_H__
+
+typedef unsigned int UINT_PTR;
+
+#endif


lib/winapi/BaseTsd.h
```diff
@@ -0,0 +1,6 @@
+#ifndef __BASETSD_H__
```
thrimbor, line 1: The rest of winapi uses all-lowercase filenames (like MinGW).

lib/winapi/badptr.c
```diff
@@ -0,0 +1,9 @@
+#include <windows.h>
+
+BOOL IsBadWritePtr(LPVOID lp, UINT_PTR ucb)
+{
+  //FIXME: Run NtQueryVirtualMemory for the pages?
```
thrimbor, line 5: The implementation on Windows registers an SEH handler (checked in Ghidra, but the MSDN description already hints it) and then just write probes each page. I didn't check the actual handler, but it's safe to assume it just catches the exception and returns `TRUE`.

As for safety on the Xbox: Write probing can't really be safe here (the pointer could point to MMIO or something). I don't think `NtQueryVirtualMemory` can solve the problem though, since the pages that make up the memory to be checked might not belong to the same allocation. I think the best we can do is replicate the write probing once we have SEH support, and add a warning about this function being unsafe.

The stub is fine for now, but I recommend creating an issue about this after merge.

lib/winapi/badptr.c
```diff
@@ -0,0 +1,9 @@
+#include <windows.h>
+
+BOOL IsBadWritePtr(LPVOID lp, UINT_PTR ucb)
```
thrimbor, line 3: nit: Other winapi functions have a space between function name and parameter parentheses.

lib/winapi/badptr.c
```diff
@@ -0,0 +1,9 @@
...
+{
+  //FIXME: Run NtQueryVirtualMemory for the pages?
+
+  /* We disallow all access for now, because memory is potentially unsafe */
+  return TRUE;
```
thrimbor, line 8: nit: The rest of winapi uses 4 space indentation.

lib/winapi/winbase.h
```diff
@@ -93,6 +94,8 @@ void SetLastError (DWORD error);
 
 void WINAPI OutputDebugStringA (LPCTSTR lpOutputString);
 
+BOOL IsBadWritePtr(LPVOID lp, UINT_PTR ucb);
```
thrimbor, line 97: nit: Other winapi functions have a space between function name and parameter parentheses (same as above).

lib/winapi/badptr.c
JayFoxRox: I'll add a note about the SEH handler implementation on Windows.

---

My idea would have been to just keep looping until all all `(ucb + 0xFFF) / 0x1000` pages starting at `lp & ~0xFFF` have been checked. If any of them doesn't have write permission, report that.
When implementing this algorithm with `NtQueryVirtualMemory` I'd skip `RegionSize` bytes at a time (instead of walking each page, I'd just walk allocations).

It would at least do what MSDN says:

> If the calling process has write access to all bytes in the specified memory range, the return value is zero.

Hence, in addition to the SEH comment, I'd like to keep the existing comment. I'll improve it to say something like "//FIXME: Walk all allocations / pages in the region with NtQueryVirtualMemory".

It was already my plan to create an issue to document this being a stub.

lib/winapi/winbase.h
JayFoxRox: Good catch - I assume this affects all my PRs from today.
Should have been fixed here, I'll fix the rest later.

lib/winapi/badptr.c
JayFoxRox: Yes, stupid mistake. Should have been fixed.

lib/winapi/badptr.c
JayFoxRox: Should have been fixed.

lib/winapi/BaseTsd.h
JayFoxRox: Should have been fixed.

lib/winapi/badptr.c
thrimbor: > My idea would have been to just keep looping until all all `(ucb + 0xFFF) / 0x1000` pages starting at `lp & ~0xFFF` have been checked. If any of them doesn't have write permission, report that.
> When implementing this algorithm with `NtQueryVirtualMemory` I'd skip `RegionSize` bytes at a time (instead of walking each page, I'd just walk allocations).

What I meant is that I'm not sure `NtQueryVirtualMemory()` can do that. MSDN mentions `state, protection, and type of a region of pages`, which, to me, sounds like it's only meant to be called for addresses that mark the begin of an allocation done by `NtAllocateVirtualMemory()`. This is only a suspicion though, based on the MSDN wording and the fact that it only takes a base address and no size / page count parameter. I did not check the actual behavior.

thrimbor:
It's fine for a stub now, so I'm merging. It would probably be best to create an issue about it to track the fixme.

PR #335, opened 2020-05-04
winapi: Stub parts of dynamic-linker API

JayFoxRox:
Pulled from my newton-dynamics port.

For now, it's just to make it easier to compile libraries or applications. Most applications will handle failure of `LoadLibrary` very gracefully (typically it's used for optional plugins).

Entirely untested at runtime, but tested compilation with:

```c
HMODULE* dll = LoadLibrary("sample.dll");
if (dll != NULL) {
  void* f = GetProcAddress(dll, "sample_function");
  FreeLibrary(dll);
}
```

I might implement a linker in the future (if nobody beats me to it).

### DIFF
--- a/lib/winapi/libloaderapi.c
+++ b/lib/winapi/libloaderapi.c
@@ -1,0 +1,42 @@
+#include <libloaderapi.h>
+#include <winbase.h>
+#include <windef.h>
+#include <winerror.h>
+#include <xboxkrnl/xboxkrnl.h>
+
+#include <assert.h>
+#include <stdlib.h>
+
+HMODULE LoadLibraryExA (LPCSTR lpLibFileName, HANDLE hFile, DWORD dwFlags)
+{
+    assert(hFile == NULL);
+    assert(dwFlags == 0);
+
+    // Always fail with not having found the library
+    SetLastError(ERROR_MOD_NOT_FOUND);
+    return NULL;
+}
+
+HMODULE LoadLibraryA (LPCSTR lpLibFileName)
+{
+    return LoadLibraryExA(lpLibFileName, NULL, 0);
+}
+
+BOOL FreeLibrary (HMODULE hLibModule)
+{
+    assert(hLibModule != NULL);
+
+    // Always claim success when free'ing a library
+    return TRUE;
+}
+
+FARPROC GetProcAddress (HMODULE hModule, LPCSTR lpProcName)
+{
+    // FIXME: If hModule is NULL, the symbol is looked up in the current module
+
+    // FIXME: If the module handle is invalid, fail with ERROR_MOD_NOT_FOUND
+
+    // Always fail with not having found the export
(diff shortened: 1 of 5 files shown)


lib/winapi/libloaderapi.c
```diff
@@ -0,0 +1,37 @@
...
+    assert(hLibModule != NULL);
+
+    // Always claim success when free'ing a library
+    return TRUE;
+}
+
+FARPROC GetProcAddress (HMODULE hModule, LPCSTR lpProcName)
+{
+    // Always fail with not having found the export
+    return NULL;
```
JayFoxRox, line 36: This is incomplete.

This should be calling `SetLastError`  but I don't know what value. I have already had trouble finding the value for `LoadLibraryExA` (would be nice if someone can also confirm that).

lib/winapi/libloaderapi.c
```diff
@@ -0,0 +1,37 @@
...
+{
+    assert(hLibModule != NULL);
+
+    // Always claim success when free'ing a library
+    return TRUE;
+}
+
+FARPROC GetProcAddress (HMODULE hModule, LPCSTR lpProcName)
+{
+    // Always fail with not having found the export
```
JayFoxRox, line 35: nit: This could `assert(hModule != NULL)` according to the MSDN (because that doesn't list `NULL` as valid input).

However, I believe I have actually seen DLLs use `GetProcAddress(NULL, "symbol")` to find their own symbol?
Would be interesting if someone could check on Windows.

thrimbor: I just checked that and it indeed works like this on Windows (even works in normal exe files for functions marked as `dllexport`).

lib/winapi/libloaderapi.c
thrimbor: `ERROR_PROC_NOT_FOUND` when the function wasn't found in the module, and `ERROR_MOD_NOT_FOUND` if the HANDLE isn't a proper handle to a module.

`ERROR_MOD_NOT_FOUND` for `LoadLibraryExA` is correct.

lib/winapi/libloaderapi.c
JayFoxRox: I added a comment about this, please verify I described the behaviour correctly.

lib/winapi/libloaderapi.c
JayFoxRox: I fail with `ERROR_PROC_NOT_FOUND` now, but left a comment about `ERROR_MOD_NOT_FOUND`.

lib/winapi/libloaderapi.c
thrimbor: Yeah it's fine.

thrimbor:
I think the code looks good now. I've taken the liberty to fix a merge conflict caused by #333 touching the same header, and I'll merge this now.

PR #338, opened 2020-05-05
libc: Stub Microsoft _ASSERT macros in crtdbg.h

JayFoxRox:
This stubs the _ASSERT macros. See MSDN: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/assert-asserte-assert-expr-macros?view=vs-2019

This file is included by newton-dynamics. It uses at least one of these macros (I forgot which one, and added all 3 for completeness).

The implementation is very naive and does not follow the description in the MSDN. I consider this a stub for now.

### DIFF
--- a/lib/xboxrt/libc_extensions/crtdbg.h
+++ b/lib/xboxrt/libc_extensions/crtdbg.h
@@ -1,0 +1,11 @@
+#ifndef XBOXRT_CRTDBG
+#define XBOXRT_CRTDBG
+
+#include <assert.h>
+
+// Part of Microsoft CRT
+#define _ASSERT_EXPR(booleanExpression, message) assert(booleanExpression)
+#define _ASSERT(booleanExpression) assert(booleanExpression)
+#define _ASSERTE(booleanExpression) assert(booleanExpression)
+
+#endif


lib/xboxrt/libc_extensions/crtdbg.h
```diff
@@ -0,0 +1,11 @@
+#ifndef XBOXRT_CRTDBG
+#define XBOXRT_CRTDBG
+
+#include <assert.h>
+
+// Part of Microsoft CRT
+#define _ASSERT_EXPR(booleanExpression, message) assert(booleanExpression)
```
thrimbor, line 7: What do you think about defining it as  `assert((booleanExpression) && (message))` instead to incorporate the message text when the assert fails?

JayFoxRox: I'd be a bit paranoid over `NULL` - so I'd rather avoid such hacks which might suddenly do something unexpected.

The benefit of including `message` would be that logic would still run (if message was `message_list[i++]` for example).

However, I don't think anyone will care about these message for now anyway. They are merely to not fail compilation and catch the biggest issues - code will have to be looked at if they trigger anyway.

If we see them being used we can implement them properly (and avoid all the hacks).

PR #339, opened 2020-05-05
Change default debug font to something more readable

JayFoxRox:
@InternalLoss complained (on Discord) that the default 8x8 font isn't readable on a display connected via composite. I'm also unhappy with the readability of the text.

@InternalLoss recommended to use [the font from DevKitPro libogc (for use in Wii homebrew)](https://github.com/devkitPro/libogc/blob/master/libogc/console_font_8x16.c).
However, I was unhappy with the charset of that, because I wanted U+0080 to U+00FF to be the [Latin-1 Supplement Unicode block](https://en.wikipedia.org/wiki/Latin-1_Supplement_%28Unicode_block%29). Instead it's [Code page 437](https://en.wikipedia.org/wiki/Code_page_437).

I then came across fonts in http://pelulamu.net/unscii/. They look a bit worse than the clean "typical VGA" fashion libogc fonts, but have the Latin-1 Supplement glyphs.

The hex fonts were converted with this python3 script:

```py
import sys

f = open(sys.argv[1])
lines = f.readlines()

print_index = 0
for line in lines:

  index, _, data = line.partition(":")
  index = int(index,16)
  data = data.strip()

  assert(index == print_index)

  s = []
  while len(data):
    s += ["0x%s" % data[0:2]] 
    data = data[2:]
  print(",".join(s) + ", // Symbol 0x%02X" % index)

  print_index += 1
  if index >= 0xFF:
    break
```

I eventually chose unscii-16 for this PR (but am willing to change it during review, as I already have the other fonts ready):

unscii-16

Here are screenshots of alternatives which could be discussed during review:
- [unscii-16 (shown above)](https://user-images.githubusercontent.com/360330/81093240-16982e00-8f02-11ea-8482-a66216959d99.png)
- [unscii-8-tall](https://user-images.githubusercontent.com/360330/81093343-40515500-8f02-11ea-94d1-ced5e3749cc6.png) (storage size could be reduced by adding a line-doubler)
- [8x16 font from libogc](https://user-images.githubusercontent.com/360330/81093374-4f380780-8f02-11ea-8cb4-6b75ebb348d3.png)
- [Current nxdk font_terminal.h](https://user-images.githubusercontent.com/360330/81093428-64149b00-8f02-11ea-82bb-5454af416d76.png)
- [Unused nxdk font_sateroids2.h](https://user-images.githubusercontent.com/360330/81093437-6aa31280-8f02-11ea-8a10-31ea0b0ca9b2.png)

*(If anyone wants another font, they should provide it in usable format)*

**This would be a breaking change, because less text fits on the screen.**

Some apps will probably break if they render using the debug font and assume a certain amount of lines to fit.

I strongly recommend these applications to migrate to rendering through SDL / SDL_ttf or a more sophisticated rendering technique like pbkit. The debug font should not be used for "terminal style" applications.

I assume the debug code will eventually disappear (potentially replaced with a more sophisticated terminal library, maybe without alternative). The proposed font change is only for life-support.


dracc:
Out of the presented alternatives I very much prefer unscii-16.

InternalLoss:
I'll test later on my Xbox with both composite and video ~~once I actually compile~~, but of these either libogc's font or the uniscii-16 look fine from your NXDK screenshots.

dracc:
A slightly modified hello sample (hard coded 640*480 showing the first 256 glyphs of each font) gave these results:

<details>
<summary>
V1.6 with component to HDMI adapter
</summary>
[!old font](https://i.imgur.com/5SvHrXh.jpg)
[!new font](https://i.imgur.com/0S1uCKU.jpg)
</details>

<details>
<summary>
V1.0 with composite cable (yes, tilted)
</summary>
[!old font](https://i.imgur.com/0B6PVGh.jpg)
[!new font](https://i.imgur.com/JlQb5Hz.jpg)
</details>

<details>
<summary>
XQEMU for reference
</summary>
[!old font](https://i.imgur.com/Ma1aY1w.png)
[!new font](https://i.imgur.com/wKIerRD.png)
</details>

"new font" == unscii-16

InternalLoss:
It seems my 1.4 Xbox can't run NXDK samples right now (it gets stuck on loading...), but from the photos shown I also agree that the font looks a lot better than the old one [my 1.6 is currently not working, if I can get it working I'll post appropriate photos]

InternalLoss:
Tested on my 1.4 Xbox: c:

PAL, Composite, 576i


NTSC-M, Composite, 480i


NTSC-M, component, 480p

lib/hal/debug.c
```diff
@@ -42,6 +42,11 @@ static void drawChar(unsigned char c, int x, int y, int fgColour, int bgColour)
 		mask = 0x01;
 #else
 		mask = 0x80;
+#endif
+
+		// This loop draws up to 1 byte (8 bit) per line; so 8 pixels at most
+#if FONT_WIDTH > 8
+#error Font can't be wider than 8 pixels
 #endif
```
Teufelchen1, line 50: You are checking the value of `FONT_WIDTH` _after_ you already used it for calculations. I can see that in this case it doesn't matter much but looks strange to me?

JayFoxRox: I'm only checking it here, because the problems only arise here. The code above is fine.

In the code above we consider 1 byte (= 8 bit) per 8 width-pixels. The old code looked like it handled this, but it didn't. It assumed `FONT_WIDTH` bytes per symbol, which was 8; that luckily happened to match `FONT_HEIGHT` which was also 8. So it correctly skipped 8 bytes per pixel (8 rows x 1 byte per row = 8x8 symbol), but the math was wrong. The new code fixes the math by rounding up to the necessary number of bytes per line using `((FONT_WIDTH+7)/8)`, then using `FONT_HEIGHT` to account for the number of lines per symbol.
So the code above will now handle *any* `FONT_WIDTH` and `FONT_HEIGHT` (until numbers overflow - but let's ignore that for now).
Hence, there's no error there.

However, as indicated by the new comment for this loop: If we have anything other than 8 for this loop, then we'd need to split drawing into groups of 8 bits each (to process each byte per line individually). We'll also have to deal with endianess and more complex mirroring logic then.
Hence, this has the error next to it.

thrimbor: I think it's fine to have the check there. I'd just like to mention that the C11 `static_assert` / `_Static_assert` is also an option for checks like this.

thrimbor:
Considering the much better readability of this font on my TV I'd like to merge this rather soon - or should we give others more time to adjust their applications @JayFoxRox?

JayFoxRox:
Looks like no [project using nxdk](https://github.com/XboxDev/nxdk/wiki/Projects-using-nxdk), other than Xenium-Tools, is using debug.h as their primary drawing method. I assume @Ryzee119 (author of Xenium-Tools) will update his software soon (or at least create an issue on his own repository). [On Discord, he said this](https://discord.com/channels/428359196719972353/428359408204906497/707355988910407720):

> It probably will cause issues in xeniumtools, but that's fine.
> I wanted to switch to sdl_ttf at some point anyway

So I think we can merge this already. It has been here a week already and got a lot of attention.
People can check-out an older nxdk if really necessary.

lib/hal/debug.c
JayFoxRox: I'd like to avoid `static_assert` for this. There's no reason to bump compiler requirements for a minor change like this.

thrimbor:
Alright, merging then.

PR #345, opened 2020-05-13
vp20compiler: Various improvements and better support for NV2A

JayFoxRox:
This PR is best reviewed one commit at a time:

See https://www.khronos.org/registry/OpenGL/extensions/NV/NV_vertex_program.txt for more background information.

### vp20compiler: Remap input registers for NV2A

The register indices for registers of the same semantics of NV2A and NV_vertex_program differ.
When we support the fixed-function-pipeline or certain immediate-mode setters in the GPU, where we can't just remap the vertex attributes, this will be a problem.

To solve this, this commit creates an array with the hardware registers. These must match the names of the NV_vertex_program registers, but be in hardware order.

The parser will check if the used NV_vertex_program registers exist in hardware. If the index is different, it warns (to avoid people assuming a 1:1 mapping and using indices from the NV_vertex_program spec on the CPU side).

The emitter will try to find the correct hardware register and `assert(false)` if nothing was found.

This also requires a change in the mesh sample, which actually uses `v[8]` (TEX0 in NV_vertex_program) when it should be using `v[9]` (TEX0 in NV2A).

The hardware mappings for this commit have been taken from XQEMU: <https://github.com/xqemu/xqemu/blob/42579ba366b69b04a79f41fe9b5744ea6c2e7193/hw/xbox/nv2a/nv2a_shaders.c#L296>.
The NV_vertex_program mappings existed in the code already and should match the spec linked above (Table X.3).
I did not check the correctness of these arrays.

### vp20compiler: Remap output registers for NV2A

Like the previous commit, but for output registers.

In this case, it's not allowed to access registers by index, so this could only cause errors, but no warnings.

This replaces some ugly table in main.c. However, this change does mix hardware and software related things in the parser. I think this is fine because I intend to create extension `OPTION`s for directly accessing NV2A registers soon.

The hardware mappings for this commit have been taken from XQEMU: <https://github.com/xqemu/xqemu/blob/master/hw/xbox/nv2a/nv2a_vsh.c#L243> (A0 is special, so we don't consider it as an output register).
The NV_vertex_program mappings existed in the code already and there are no explicit indices, so the order should be irrelevant; see spec linked above (Table X.1).

### vp20compiler: Improve vertex attribute register checks for VSPs

The old check allowed registers like `v[05]` which should not have been allowed. This fixes that.

Also it is a bit confusing that VSPs can only be accessed by index, so I made an explicit error for it.

### vp20compiler: Do not list compiler-code location in error-messages

The old error printer also printed the location in the vp20compiler-code which triggered the message. I feel this doesn't add anything. It only confuses users of the tool, so I removed it.

### vp20compiler: Improve error-messages for syntax-errors

This replaces a frequently appearing pattern using the `RETURN_ERROR;` macro by `EXPECT("symbol")`, to provide more explicit error messages.

### vp20compiler: Improve error-messages for version checks

The version checks incorrectly claimed that version 1.0 did not support something. However, this message can does not tell the user which version is actually required.

The new message informs the user that version 1.1 is required.

### DIFF
--- a/tools/vp20compiler/config.h
+++ b/tools/vp20compiler/config.h
@@ -11,17 +11,7 @@
 #define MAX_NV_VERTEX_PROGRAM_PARAMS       192
 #define MAX_NV_VERTEX_PROGRAM_INPUTS        16
 #define MAX_HARDWARE_INPUTS                 16
-/*
- Actually Xbox has 11 outputs, however, the parser is using the GL spec and
- expects 8 units. From the GL spec (NV_vertex_program, section 2.14.1.5):
-
-   "If the number of texture units supported is less than eight, the values of
-    vertex result registers that do not correspond to existent texture units
-    are ignored."
-
- So despite only having 4 texture units, we provide 8 texture units here.
- = 11 actual outputs (including 4 texture units) + 4 fake texture-units.
-*/
-#define MAX_NV_VERTEX_PROGRAM_OUTPUTS       (11+4)
+#define MAX_NV_VERTEX_PROGRAM_OUTPUTS       15
+#define MAX_HARDWARE_OUTPUTS                15
 
 #endif
--- a/tools/vp20compiler/config.h
+++ b/tools/vp20compiler/config.h
@@ -10,6 +10,7 @@
 #define MAX_NV_VERTEX_PROGRAM_TEMPS         13
 #define MAX_NV_VERTEX_PROGRAM_PARAMS       192
 #define MAX_NV_VERTEX_PROGRAM_INPUTS        16
+#define MAX_HARDWARE_INPUTS                 16
 /*
  Actually Xbox has 11 outputs, however, the parser is using the GL spec and
  expects 8 units. From the GL spec (NV_vertex_program, section 2.14.1.5):
--- a/tools/vp20compiler/main.c
+++ b/tools/vp20compiler/main.c
@@ -540,9 +540,18 @@ void translate(const char* str)
                     // TODO: the index needs ajustment?
                     vsh_set_field(vsh_ins, FLD_CONST, reg.Index+96);
                 } else if (reg.File == PROGRAM_INPUT) {
+                    int in_reg;
+                    const char* name = _mesa_nv_vertex_input_register_name(reg.Index);
+                    for (in_reg = 0; _mesa_nv_vertex_hw_input_register_name(in_reg); in_reg++) {
+                        if (strcmp(name, _mesa_nv_vertex_hw_input_register_name(in_reg)) == 0) {
+                            break;
+                        }
+                    }
+                    if (!_mesa_nv_vertex_hw_input_register_name(in_reg)) {
+                        assert(false);
+                    }
                     vsh_set_field(vsh_ins, mux_field[j], PARAM_V);
-                    //TODO: also needs ajustment?
-                    vsh_set_field(vsh_ins, FLD_V, reg.Index);
+                    vsh_set_field(vsh_ins, FLD_V, in_reg);
                 } else {
                     assert(false);
                 }
--- a/tools/vp20compiler/main.c
+++ b/tools/vp20compiler/main.c
@@ -164,21 +164,6 @@ static const VshFieldMapping field_mapping[] = {
     {  FLD_FINAL,            3,    0,     1 }
 };
 
-
-typedef enum {
-    OUTREG_oPos = 0,
-    OUTREG_oD0 = 3,
-    OUTREG_oD1,
-    OUTREG_oFog,
-    OUTREG_oPts,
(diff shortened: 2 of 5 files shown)


tools/vp20compiler/config.h
```diff
@@ -10,17 +10,8 @@
...
-    vertex result registers that do not correspond to existent texture units
-    are ignored."
-
- So despite only having 4 texture units, we provide 8 texture units here.
- = 11 actual outputs (including 4 texture units) + 4 fake texture-units.
-*/
-#define MAX_NV_VERTEX_PROGRAM_OUTPUTS       (11+4)
+#define MAX_HARDWARE_INPUTS                 16
+#define MAX_NV_VERTEX_PROGRAM_OUTPUTS       15
+#define MAX_HARDWARE_OUTPUTS                15
```
JayFoxRox, line 15: nit: This is a bit arbitrarily sized.

The field is actually 8 bits, so theoretically we should have 256 registers. However, register index 0xF is index register `A0` (according to XQEMU code), so it's safe to assume that the registers above are garbage or aliases.

In a future PR I plan to remove these preprocessor defines to replace them with a more dynamic system which could be used to experiment with higher output indices. For now, 15 is fine.

tools/vp20compiler/main.c
```diff
@@ -540,9 +540,18 @@ void translate(const char* str)
                     // TODO: the index needs ajustment?
                     vsh_set_field(vsh_ins, FLD_CONST, reg.Index+96);
                 } else if (reg.File == PROGRAM_INPUT) {
+                    int in_reg;
+                    const char* name = _mesa_nv_vertex_input_register_name(reg.Index);
+                    for (in_reg = 0; _mesa_nv_vertex_hw_input_register_name(in_reg); in_reg++) {
+                        if (strcmp(name, _mesa_nv_vertex_hw_input_register_name(in_reg)) == 0) {
+                            break;
```
Teufelchen1, line 515: This code checks, if a given input_register actually exists in hardware, right?

tools/vp20compiler/nvvertparse.c
```diff
@@ -119,15 +119,16 @@ _mesa_find_line_column(const unsigned char *string, const unsigned char *pos,
...
 static void
 record_error(struct parse_state *parseState, const char *msg, const char* file,
-             int lineNo)
+             int lineNo, bool onlyWarn)
 {
    int line, column;
    const unsigned char *lineStr;
    lineStr = _mesa_find_line_column(parseState->start,
                                     parseState->pos, &line, &column);
    // _mesa_debug(parseState->ctx,
```
Teufelchen1, line 128: Why not also remove unused, out-comment'ed, undocumented code? Is there a reason to keep it?

tools/vp20compiler/main.c
JayFoxRox: The program in memory is in OpenGL format, but this code translates it to the hardware.

So this code retrieves the GL name by index using:
- `_mesa_nv_vertex_input_register_name[]`
and then loops over the hardware names using (note the `hw_`):
- `_mesa_nv_vertex_hw_input_register_name[]`
to find the respective hardware index (by matching the GL name and hw name in a sorted array).

tools/vp20compiler/nvvertparse.c
JayFoxRox: It's not touched by this PR, so I avoided touching it. Can be done in a cleanup PR later (I probably won't do it, because it's not really important and the code quality in this file isn't too great to begin with).

tools/vp20compiler/main.c
Teufelchen1: I see. I got confused because it looks similar to [this](https://github.com/XboxDev/nxdk/blob/629e55b196e72a23a5b0894b5021799a78e5a810/tools/vp20compiler/nvvertparse.c#L571) in `nvvertparse.c#L571`.  Case closed :)

JayFoxRox: It *is* the same.

nvvertparse is the parser which creates structures from the program code, and at that time we have to check for parser errors / syntax errors. So that's when we have to report any problem with non-existing hardware registers (so we have to map it).

We could store the hardware index at that point (during parsing), but I decided against it because the mesa code expects the GL indices / names.
We might also want to provide extensions which adds virtual registers (which don't exist in hardware), so using hardware indices is probably worse than using the loosely defined GL indices which can be extended more easily.

Therefore, the `translate` function has to do it again (and using another interface, as it's in another file).

---


It's very possible that this will be refactored in the future. For now, I just wanted something that works. And this is suitable.

thrimbor:
I've finally reviewed the changes, and they look fine to me. Also seemed to cause no regressions, so I'm merging it now.

PR #350, opened 2020-05-19
build: Work around LLVM LTO symbol issues

thrimbor:
This fixes lld crashes with LLVM 10 (and maybe more versions) when using LTO. The crash is caused by issues in lld with regards to handling of symbols that get emitted during code generation, in this case, `__fltused` and `__chkstk`/`__xlibc_check_stack`. I've opened a bug report upstream for this: https://bugs.llvm.org/show_bug.cgi?id=45985

The solution in this PR is to add linker parameters that make lld add the symbols properly. I would prefer a cleaner solution, but this will probably require waiting for a future LLVM release. In the meantime, this makes it possible to build SDLPoPX with LTO turned on.


Ryzee119:
I was successfully able to build SDLPoPX with LTO on after this change.
Played through the first few levels on my xbox. No issues seen.

JayFoxRox:
Almost 2 weeks with no response upstream (I had hoped for an alternative solution).

Change in code looks fine.
I'm merging this, trusting @Ryzee119 review and @thrimbor experience with LLVM.

PR #351, opened 2020-05-24
Improvements for nforceif and lwip config

thrimbor:
First of all, this will not dramatically change the network performance. It does however reduce busy-waiting and makes some lwip config changes, so slight performance improvements may be observed.

This PR also includes some formatting changes, just to make it easier to work with the code in an IDE (which doesn't like the different formatting and unnecessary whitespaces). Some whitespace changes slipped into commits that actually change functionality, and while I normally would cleanly separate them, I really don't file like working around whitespace issues rn.
I recommend going through the changes by commit because the formatting changes blow up the total diff quite a bit and make it harder to see the actual changes.

As for the functional changes:
This PR mostly consists of modifications to how semaphores and mboxes are implemented in our lwip compatibility layer (nforceif). The old implementation used semaphores in a weird way that actually made them use busy-waiting, this changes it so that they make better use of the semaphore functionality provided by the kernel (the maximum count of `10000` passed to `NtCreateSemaphore` was taken from lwip's Windows backend).
The improved semaphores are then used to re-implement parts of the mbox code, so that it too doesn't use any busy-waiting.

Windows CI currently fails due to msys2 issues unrelated to this PR.


dracc:
Tested with NevolutionX FTP server, transfer speeds rose from ~1.3MiB/s to ~3.0MiB/s. :+1:

lib/net/nforceif/src/sys_arch.c
```diff
@@ -347,31 +350,26 @@ sys_sem_new(struct sys_sem **sem, u8_t count)
...
-            if (time_needed >= timeout) return SYS_ARCH_TIMEOUT;
-        }
+    NTSTATUS status = NtWaitForSingleObject(sem->handle, FALSE, timeout_ptr);
+    if (status == STATUS_TIMEOUT) {
+        return SYS_ARCH_TIMEOUT;
     }
+    assert(status == STATUS_SUCCESS);
 
-    return (u32_t)time_needed;
+    return 0;
```
JayFoxRox, line 313: *(This is mostly a note to myself, that can be answered by anyone)*

I still have to check the lwip docs, but why is this constant 0 now, when it used to return how much time was spent? How is this return value used?

Also what is `SYS_ARCH_TIMEOUT`? A duration? in what unit? or a magic constant?

lib/net/nforceif/src/sys_arch.c
```diff
@@ -372,6 +313,25 @@ sys_arch_sem_wait(struct sys_sem **s, u32_t timeout)
...
+    timeout.QuadPart = 0;
+
+    NTSTATUS status = NtWaitForSingleObject(sem->handle, FALSE, &timeout);
+    if (NT_SUCCESS(status)) {
+        if (status == STATUS_TIMEOUT) {
+            return 0;
+        }
+    }
+
+    return NT_SUCCESS(status);
```
JayFoxRox, line 332: nit: I feel this is a bit ugly. I prefer:

```c
    if (NT_SUCCESS(status)) {
        if (status == STATUS_TIMEOUT) {
            return 0;
        }
        return 1;
    } else {
        return 0;
    }
```

or

```c
    if (!NT_SUCCESS(status)) {
        return 0;
    }
   
    if (status == STATUS_TIMEOUT) {
        return 0;
    }

    return 1;
```

lib/net/nforceif/src/sys_arch.c
```diff
@@ -372,6 +313,25 @@ sys_arch_sem_wait(struct sys_sem **s, u32_t timeout)
     return 0;
 }
 
+int sys_arch_sem_try(struct sys_sem **s)
```
JayFoxRox, line 316: Is this for a more recent lwip? I thought the `sys_arch_*` functions were the backend for lwip.
So it's a bit weird to see a new one. I wonder if this should have another prefix / none, and be made `static`.

This is also the only function to be platform dependent in this commit:
- Was the (old or new) mbox implementation taken from elsewhere in lwip?
- Is there a reference implementation we could just use?

lib/net/nforceif/include/arch/sys_arch.h
```diff
@@ -12,21 +12,21 @@
...
- * IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY 
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ * WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
+ * OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+ * IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY
```
JayFoxRox, line 25: nit: I'd prefer if those whitespace changes were at least in the "nforceif: Clean up and reformat sys_arch.c" commit.

If you rename that commit to "nforceif: Clean up and reformat" then you can also include the whitespace changes to lwipopts.h below.

lib/net/nforceif/include/lwipopts.h
```diff
@@ -46,20 +46,53 @@
...
+ * ATTENTION: this does not work when tcpip_input() is called from
+ * interrupt context!
+ */
+#define LWIP_TCPIP_CORE_LOCKING_INPUT   1
+
+/**
+ * SYS_LIGHTWEIGHT_PROT==1: enable inter-task protection (and task-vs-interrupt
+ * protection) for certain critical regions during buffer allocation, deallocation
+ * and memory allocation and deallocation.
+ * ATTENTION: This is required when using lwIP from more than one context! If
```
JayFoxRox, line 92: nit: This is some weird and inconsistent formatting, was this copied 1:1 from lwip?

More importantly: why didn't our options file include these parts in the first place?

lib/net/nforceif/include/lwipopts.h
```diff
@@ -310,6 +310,24 @@
...
+#define TCP_MSS                         1432
+
+/**
+ * TCP_WND: The size of a TCP window.  This must be at least
+ * (2 * TCP_MSS) for things to work well.
+ * ATTENTION: when using TCP_RCV_SCALE, TCP_WND is the total size
+ * with scaling applied. Maximum window value in the TCP header
+ * will be TCP_WND >> TCP_RCV_SCALE
+ */
+#define TCP_WND                         (45 * TCP_MSS)
```
JayFoxRox, line 329: How did you arrive at these values? `TCP_MSS` and `TCP_WND`?

lib/net/nforceif/include/lwipopts.h
```diff
@@ -332,6 +362,25 @@
...
+ * sys_thread_new() when the thread is created.
+ */
+#define TCPIP_THREAD_STACKSIZE          4096
+
+/**
+ * DEFAULT_THREAD_STACKSIZE: The stack size used by any other lwIP thread.
+ * The stack size value itself is platform-dependent, but is passed to
+ * sys_thread_new() when the thread is created.
+ */
+#define DEFAULT_THREAD_STACKSIZE        4096
```
JayFoxRox, line 383: nit: How are these thread stacksizes used?

I'm always a bit paranoid about some code doing something like `DEFAULT_THREAD_STACKSIZE+4 /* make room for return address */` (fantasy example - not real code in lwip) and then it consumes more than one page.

Also is one page enough? Could there be some bad recursion or something like `alloca` for large packet streams?

lib/net/nforceif/include/lwipopts.h
```diff
@@ -46,20 +46,53 @@
...
  */
 #define NO_SYS                          0
 
+/*
+   ------------------------------------
+   ----------- Core locking -----------
+   ------------------------------------
+*/
+
+/**
```
Teufelchen1, line 68: Whats the difference between `/*` for a block comment and `/**`?

lib/net/nforceif/include/lwipopts.h
```diff
@@ -277,6 +310,24 @@
...
 #define LWIP_TCP                        1
 
+/**
+ * TCP_MSS: TCP Maximum segment size. (chosen to be the maximum
+ * that can be transmitted over IPv6+PPPoE unfragmented)
+ * For the receive side, this MSS is advertised to the remote side
+ * when opening a connection. For the transmit size, this MSS sets
+ * an upper limit on the MSS advertised by the remote host.
+ */
+#define TCP_MSS                         1432
```
Teufelchen1, line 320: An ethernet frame is 1500 bytes. Subtracting 8 bytes overhead for PPPoE and another 40 bytes for the IPv6 header  gives us 1452 bytes. 🤔 What am I missing to match your proposed 1432 bytes?

Teufelchen1: Oh, I forgot TCP itself, right? Can't resolve the comment myself, sorry!

lib/net/nforceif/src/sys_arch.c
thrimbor: Oops, screwed that up during rewrite. It's supposed to return the milliseconds spent waiting, which it does now.

`SYS_ARCH_TIMEOUT` is just some constant (`0xffffffff`) to signal timeouts.

lib/net/nforceif/src/sys_arch.c
thrimbor: There's no need for exact return values here, so I'd rather avoid additional comparison logic.

lib/net/nforceif/src/sys_arch.c
thrimbor: It's an extension to be able to implement the `sys_mbox_try*` functions. It was already static, but I added the `static` keyword to the definition now, too, to make it clear.

I don't know where the old mbox implementation came from, but the new one isn't taken from lwip, although the implementations there are probably similar (they're all system-specific ofc).

lib/net/nforceif/include/arch/sys_arch.h
thrimbor: Done.

lib/net/nforceif/include/lwipopts.h
thrimbor: Yes, copied from lwip (like the other config options). As for why we didn't have it before, idk.

lib/net/nforceif/include/lwipopts.h
thrimbor: `TCP_MSS` is set to the Ethernet frame size minus PPPoE, IPv6 and TCP headers. I wanted to support this configuration without risking fragmenting packets, while still having the MSS large enough to get good performance.

`TCP_WND` is set to the largest multiple of `TCP_MSS` (required by lwip) that fits into the 16 bit field (2^16 / 1432 = 45.765).

lib/net/nforceif/include/lwipopts.h
thrimbor: In our case they're only used for the tcp thread, which takes packets in pbufs from the driver via an mbox, does some parsing and endianess conversion, and then hands them to application threads via another mbox.

All buffers are managed by pbufs, so I don't expect any large stack usage, and it never caused issues during testing. With `LWIP_TCPIP_CORE_LOCKING_INPUT`, the tcp thread should barely be doing anything, anyway.

lib/net/nforceif/src/sys_arch.c
JayFoxRox: But your code does more checks? (Your checking logic starts 5 lines up from where I left my comment)

Your logic checks `NT_SUCCESS` twice, and then it's not very clear if `NT_SUCCESS` is just 0 or 1, or if it might actually just forward positive values (which might not be intended).

lib/net/nforceif/src/sys_arch.c
JayFoxRox: If this is your own extension and lwip code normally doesn't have that function (including other platform specific code), then I'd recommend to rename that function; or at least a comment in the code would help, to clarify that this was added for our backend but that it isn't part of the lwip API.

lib/net/nforceif/include/lwipopts.h
JayFoxRox: I feel like such comments should be part of the config file / code.
It's not obvious but it does make sense.

lib/net/nforceif/src/sys_arch.c
thrimbor: Hm, ok, I changed it to the second variant (using `-1` instead of `1` though).

lib/net/nforceif/src/sys_arch.c
thrimbor: I've changed the name to `ext_sys_arch_sem_try` and added a comment.

lib/net/nforceif/include/lwipopts.h
thrimbor: I added comments to those two entries now.

lib/net/nforceif/include/arch/sys_arch.h
```diff
@@ -41,7 +41,7 @@
 extern unsigned char debug_flags;
 #endif
 
-typedef u32_t sys_prot_t;
+typedef unsigned char sys_prot_t;
```
JayFoxRox, line 44: nit: This could have a comment that explains that it's `KIRQL`

JayFoxRox:
Consider adding yourself to the copyright header to get legal clarity.

I have tested this with nxdk-rdt and skimmed over the code a couple of times. I'll merge this later if there's no veto.

PR #354, opened 2020-05-25
ci: Fix path conflict with preinstalled msys2

thrimbor:
Our Windows CI recently started failing. The issue came down to a path conflict - the newer images contain an msys2 installation in `C:\msys2`, which is also where our own msys2 installation got extracted to. Partially overwriting the pre-existing installation then caused weird issues when trying to update or install packages.

This fixes it by changing the path where we extract msys2 to `C:\tools\msys2`, which is the path where chocolatey normally puts it (this path is also used in xqemu). Since we don't use chocolatey here to install msys2, there should be no path conflicts.


JayFoxRox:
Looks okay and seems to solve our problem. Merging.

PR #355, opened 2020-05-28
Allocate framebuffer when XVideoSetMode is called

### LINKED ISSUE #149: vmCurrent uninitialized before XVideoSetMode call
[From @JayFoxRox's comment](https://github.com/XboxDev/nxdk-pdclib/pull/3#issuecomment-506748349)

With the updated pdclib, `XVideoSetMode` is no longer being automatically called in ` platform/xbox/crt0.c`. The current implementation will leave `vmCurrent in lib/hal/video.c` uninitialized until `XVideoSetMode` is called within the application.

We don't currently handle this gracefully, and we should adjust the implementation. Some implementation suggestions from JFR:

> A temporary solution will be to initialize it to an unknown state, and to disallow certain APIs (including `debugPrint`) until `XVideoSetMode` was explicitly called.
> A possible permanent solution would be to refactor all APIs, or to detect if the mode is unknown, and read the hardware state.
> [We should be able to recover the pVidMode->dwMode from XVideoInit](https://github.com/XboxDev/nxdk/blob/master/lib/hal/video.c#L332), using:
```c
ULONG mode;
AvSendTVEncoderOption(0xFD000000, 13, 0, &mode);
```
> That should give us the width and height.
> For the framebuffer format (bits per pixel) we'll probably have to query the `CR[0x28]` and `RAMDAC` registers.

### LINKED ISSUE #169: `XVideoInit` never allocated the framebuffer memory that it advertises via `XVideoGetFB`

### LINKED ISSUE #170: `XVideoGetFB` can return outdated address
*(This isn't much of a bug, but more of a warning about the video.h API; it needs guidelines, not necessarily a code-change in the API itself)*

Aside from #169, there's an issue where setting `PCRTC_START` manually, without `XVideoSetDisplayStart`, will cause `XVideoGetFB` to return old / outdated values.

pbkit abuses this to keep `XVideoGetFB` at the original framebuffer which is used by `debugPrint`, and controls the displayed framebuffer using `PCRTC_START` manually (to either its own, or the debug framebuffer).

Other libs might also do this. This is bad.

It makes it easy to cause issues with drawing functions which use `XVideoGetFB` (such as `debugPrint`): the drawing won't write to framebuffer (= nothing visible) and potentially writes to random memory (= memory corruption).

This API should be made safer using policies:

- video.h could take ownership of `PCRTC_START`, meaning it would be disallowed to be used with some other APIs (like pbkit, until it also uses video.h correctly).
- Implicit drawing (as done by debug.h) should be discouraged.

Code changes could also be done:

- Fix: Querying the framebuffer from `PCRTC_START` (would probably be too slow).
- Design change: video.h could become a pure setter (users would be responsible to inform drawing functions about the current display mode); this makes implicit drawing impossible.

### LINKED ISSUE #269: hal/video.c variables should be static
There are a lot of variables in hal/video.c which should be `static`.

https://github.com/XboxDev/nxdk/blob/ce8da0b9cf93a6ce0051a7a2399cfdb93771e6ea/lib/hal/video.c#L36-L42
https://github.com/XboxDev/nxdk/blob/ce8da0b9cf93a6ce0051a7a2399cfdb93771e6ea/lib/hal/video.c#L61-L62
https://github.com/XboxDev/nxdk/blob/ce8da0b9cf93a6ce0051a7a2399cfdb93771e6ea/lib/hal/video.c#L121

Some of this interacts with #169 because `_fb` is necessary to manually do this hack from the outside.

Ideally, we'd solve #169 and make other variables static, and provide getters where necessary (possibly none at this point).

JayFoxRox:
This PR adds a framebuffer that is owned by XVideo.

This change decouples XVideo from the actual GPU framebuffer. We only expect XVideo to reflect some virtual single-buffered display now, and all changes to that will be copied to the GPU, but not the other way around.
In other words: if the user circumvents XVideo to switch to another buffer / mode, then XVideo will stick to the configured mode / resolution for its own framebuffer.

This closes #170 because XVideo will no longer be expected to synchronized with the GPU (XVideo sets the GPU, but never reads it).

Some notes:

- Unconditionally (re-)allocate our own framebuffer in `XVideoSetMode` (write-combined); this closes #169. There is currently no way to free or avoid this buffer.
- Assert that we got a framebuffer in `XVideoInit`; ideally this would just make `XVideoSetMode` fail, but this can be done later. An issue will have to be created after merge.
- Zero-out the mode returned from `XVideoGetMode` until it was set; this closes #149 because we no longer have a random state, but a well-defined state. I decided against querying the GPU for information, because it would complicate the code and complicate APIs like debug.c which would no longer know when the framebuffer is actually owned. Another API can be added to retrieve the state in the future.
- More carefully synchronize with `XVideoGetMode` in debug.c; this prevents drawing until video mode was set (previously, bugs allowed it to `memset` some memory). So debug.c should no longer be dangerous.
- `sfence` after drawing in debug.c [to avoid data being stuck in write-combining buffers]; also see https://github.com/XboxDev/nxdk-sdl/pull/32 for the explanation / comments.
- Hide some variables in video.c; this should avoid API misuse, hopefully leading to cleaner API design in the future; this closes #269.

Another API for freeing the buffer might be added in the future.

For now, if you want to avoid the buffer, write your own API and avoid XVideo and any API that depends on initialized XVideo. Even without XVideo, you can still safely use debug.c, but it will be a no-op (except for flushing write-combine buffers and some useless math).

*This was not tested yet.*

### DIFF
--- a/lib/hal/debug.c
+++ b/lib/hal/debug.c
@@ -15,9 +15,10 @@
 #define MARGIN         25
 #define MARGINS        50 // MARGIN*2
 
-int SCREEN_WIDTH	= 640;
-int SCREEN_HEIGHT	= 480;
-int SCREEN_BPP = 32;
+unsigned char *SCREEN_FB = NULL;
+int SCREEN_WIDTH	= 0;
+int SCREEN_HEIGHT	= 0;
+int SCREEN_BPP = 0;
 
 int nextRow = MARGIN;
 int nextCol = MARGIN; 
@@ -27,9 +28,18 @@ static const unsigned char systemFont[] =
 #include "font_unscii_16.h"
 };
 
+static void synchronizeVideoMode()
+{
+	VIDEO_MODE vm = XVideoGetMode();
+	SCREEN_WIDTH = vm.width;
+	SCREEN_HEIGHT = vm.height;
+	SCREEN_BPP = vm.bpp;
+	SCREEN_FB = XVideoGetFB();
+}
+
 static void drawChar(unsigned char c, int x, int y, int fgColour, int bgColour)
 {
-	unsigned char *videoBuffer = XVideoGetFB();
+	unsigned char *videoBuffer = SCREEN_FB;
 	videoBuffer += (y * SCREEN_WIDTH + x) * ((SCREEN_BPP+7)/8);
 
 	unsigned char mask;
@@ -146,11 +156,9 @@ void debugPrint(const char *format, ...)
 	va_start(argList, format);
 	vsprintf(buffer, format, argList);
 	va_end(argList);
-	
-	VIDEO_MODE vm = XVideoGetMode();
-	SCREEN_WIDTH = vm.width;
-	SCREEN_HEIGHT = vm.height;
-	SCREEN_BPP = vm.bpp;
+
+	synchronizeVideoMode();
+
 	int fgColour;
 	int bgColour;
 	switch (SCREEN_BPP) {
@@ -198,11 +206,13 @@ void debugPrint(const char *format, ...)
 
 void advanceScreen( void )
 {
+	synchronizeVideoMode();
+
 	int pixelSize = (SCREEN_BPP+7)/8;
 	int screenSize  = SCREEN_WIDTH * (SCREEN_HEIGHT - MARGINS)  * pixelSize;
 	int lineSize    = SCREEN_WIDTH * (FONT_HEIGHT + 1) * pixelSize;
 	
-	unsigned char* thisScreen = XVideoGetFB() + (SCREEN_WIDTH * MARGIN)  * pixelSize;
+	unsigned char* thisScreen = SCREEN_FB + (SCREEN_WIDTH * MARGIN)  * pixelSize;
 	unsigned char* prevScreen = thisScreen+lineSize;
 		
 	memmove(thisScreen, prevScreen, screenSize);
@@ -213,9 +223,9 @@ void advanceScreen( void )
 
 void debugClearScreen( void )
 {
-	unsigned char* videoBuffer = XVideoGetFB();
+	synchronizeVideoMode();
 
-	memset( videoBuffer, 0, ((SCREEN_BPP+7)/8) * (SCREEN_WIDTH * SCREEN_HEIGHT) );
+	memset( SCREEN_FB, 0, ((SCREEN_BPP+7)/8) * (SCREEN_WIDTH * SCREEN_HEIGHT) );
 	nextRow = MARGIN;
 	nextCol = MARGIN; 
 }
--- a/lib/hal/debug.c
+++ b/lib/hal/debug.c
@@ -37,6 +37,11 @@ static void synchronizeVideoMode()
 	SCREEN_FB = XVideoGetFB();
 }
 
+static void writebackBuffers()
+{
+	asm __volatile__("sfence");
+}
+
 static void drawChar(unsigned char c, int x, int y, int fgColour, int bgColour)
 {
 	unsigned char *videoBuffer = SCREEN_FB;
@@ -202,6 +207,8 @@ void debugPrint(const char *format, ...)
 
 		s++;
 	}
+
+	writebackBuffers();
 }
 
 void advanceScreen( void )
@@ -219,6 +226,8 @@ void advanceScreen( void )
 
 	nextRow -= (FONT_HEIGHT+1);
 	nextCol  = MARGIN; 
+
+	writebackBuffers();
 }
 
 void debugClearScreen( void )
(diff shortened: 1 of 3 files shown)


lib/hal/debug.c
```diff
@@ -27,9 +28,23 @@ static const unsigned char systemFont[] =
...
+	VIDEO_MODE vm = XVideoGetMode();
+	SCREEN_WIDTH = vm.width;
+	SCREEN_HEIGHT = vm.height;
+	SCREEN_BPP = vm.bpp;
+	SCREEN_FB = XVideoGetFB();
+}
+
+static void writebackBuffers()
+{
+	asm __volatile__("sfence");
```
JayFoxRox, line 42: This code expects to use `XVideo`; so this should just be something like `XVideoFlush();`.

lib/hal/debug.c
```diff
@@ -27,9 +28,23 @@ static const unsigned char systemFont[] =
 #include "font_unscii_16.h"
 };
 
+static void synchronizeVideoMode()
```
JayFoxRox, line 31: Should be called `synchronizeFramebuffer` or `synchronizeScreen`

lib/hal/video.c
```diff
@@ -331,12 +331,28 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
...
 	do
 	{
 		Step = AvSetDisplayMode((PVOID)VIDEO_BASE, Step, 
-			dwMode, dwFormat, pitch, VIDEO_FRAMEBUFFER);
+			dwMode, dwFormat, pitch, (unsigned int)framebufferMemory & 0x7FFFFFFF);
 	} while(Step);
 
+	/* Store the framebuffer that we set; normally this is done in XVideoSetFB.
+	   However, the first one is set using the kernel, so we do it here, too. */
+	_fb = framebufferMemory;
```
Teufelchen1, line 354: This is a bit confusing as it could mean:
- we own two fb, one made by ourselves and one given to us by the xbox kernel
- we own one fb and the kernel does, but we do the same operation to our fb which thee kernel did to its fb
- we replace a fb given by the kernel with our own
So the comment expects abstract understanding about what's going on in the greater scope. This can be fine but I think giving it a second thought would be a good idea.
(It might be my incompetence here 🤷‍♀️)

JayFoxRox: Neither of that. I agree that the wording is very poor.

What I meant is: We always store the *displayed* framebuffer *address* in `_fb`. Normally this is done in `XVideoSetFB` which sets the address to be displayed by the GPU and then keeps a copy of the set address in `_fb`.
However, when also changing the mode, we don't use `XVideoSetFB`. Instead, we are using the kernel function `AvSetDisplayMode` which internally also sets the address that will be displayed by the GPU; so we have to modify `_fb` after that, too.

I'll have to find a more concise way of phrasing this though.

I'll fix this when I get back to this branch; so keep this unresolved for now.

Ryzee119:
I tested this successfully on a few apps:
[SDLPoPX ](https://github.com/Ryzee119/SDLPoPX) (640x480 32bpp)
[Commander Keen](https://github.com/Ryzee119/omnispeak) (640x480 16bpp)

This also allowed [NevolutionX](https://github.com/dracc/NevolutionX/) to correctly display 1080i.

Before:
20200626_210337_resized 1570

After:
20200626_210804_resized 1571

As this makes `_fb` static, homebrew that currently has hacks in place to fix the exiting framebuffer issues will fail to link and wont compile. so the hacks will need to be removed when this is merged 

As using 1080i seems very unsafe at the moment this will hopefully get merged asap :)

JayFoxRox:
I'm busy with non-XboxDev things right now, so this is superseded by #388 (thanks to @thrimbor for taking over).

PR #360, opened 2020-06-05
hal: Remove dead constants and adjust function name prefix

### LINKED ISSUE #171: debug.h pollutes namespace
- `void advanceScreen( void );` has no `debug` prefix, unlike all other functions.
- Various colors are included, even though there's no way to use them (`WHITE`, `BLACK`, `RED`, `GREEN`, `BLUE`).
- Colors are also defined for other bit-depth, even though the debug.h normally queries the video-mode itself (therefore it's weird that the user would have have to pick this themselves; even then, colors could be derived from their R8G8B8 counterparts).

Such generic definitions should be avoided.

Teufelchen1:
Hi 👋 
This PR
- removes unused color constants from `debug.h`
- renames `advanceScreen()` to `debugAdvanceScreen()` to mind the namespace

Aims to close #171 

Samples still compile and a quick search revealed that `advanceScreen()` is no where in use inside nxdk.

### DIFF
--- a/lib/hal/debug.c
+++ b/lib/hal/debug.c
@@ -172,7 +172,7 @@ void debugPrint(const char *format, ...)
 	{
 		if( nextRow >= (SCREEN_HEIGHT-MARGINS) ) {
 			debugClearScreen();
-			// advanceScreen();
+			// debugAdvanceScreen();
 		}
 		
 		if (*s == '\n')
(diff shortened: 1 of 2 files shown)


JayFoxRox:
Changes look fine.

I'd probably have used a "debug:" prefix here. However, "hal:" is also fine in this case because those constants and functions never were debug prefixed either, so people might just consider them part of "hal" but not part of "debug".

We should probably wait a couple of days, because this is an API breaking change again. But being rarely used function / constants, I don't think we should delay this for very long either.

Teufelchen1:
It close to two weeks. I think we can merge now.

thrimbor:
Looks good, merging.

PR #361, opened 2020-06-05
fp20compiler:  Replace bad emitter string for REG_ONE with assert(false)

### LINKED ISSUE #236: fp20compiler: REG_ONE has bad emitter string
`REG_ONE` is a pseudo-register that should never reach the emitter.
However, I accidentally introduced an emitter string for it in #224:

https://github.com/XboxDev/nxdk/blob/a10365a862da641b9a59e29dbcb2a90166b3ee0b/tools/fp20compiler/rc1.0_register.h#L77-L78

0x12 as register would likely result in undefined behaviour.
This isn't a real problem as this case should never be reached (`one` should always be turned into an inverse zero; it worked correctly in my tests).
However, ideally this case would be an `assert(false)`.

Teufelchen1:
Hello 👋 
This PR closes #236 
It does so, as suggested, by replacing the return value with an `assert(false);`. 
I wasn't sure if I should also add a warning `fprintf` or not. Feel free to request it 🙂

### DIFF
--- a/tools/fp20compiler/rc1.0_register.h
+++ b/tools/fp20compiler/rc1.0_register.h
@@ -75,7 +75,11 @@ static const char* GetRegisterNameString(unsigned int reg_name) {
     case REG_DISCARD:
         return "0x0";
     case REG_ONE:
-        return "0x12";
+        // REG_ONE is a pseudo-register that should have been 
+        // mapped to REG_ZERO (with modifier) in MappedRegisterStruct::Init; 
+        // therefore, this should not be reachable. 
+        assert(false);
+        break;
     default:
         fprintf(stderr, "unknown register name index %d\n", reg_name);
         assert(false);


tools/fp20compiler/rc1.0_register.h
```diff
@@ -75,7 +75,7 @@ static const char* GetRegisterNameString(unsigned int reg_name) {
     case REG_DISCARD:
         return "0x0";
     case REG_ONE:
-        return "0x12";
+        assert(false);
```
JayFoxRox, line 78: This should get a comment which explains that it's a pseudo-register which translates to `REG_ZERO`.

JayFoxRox, line 78: This needs to `break;`

Teufelchen1: I'm not happy with my comment. 😨I would love the comment to be a bit more explanation for the _why_ and not just _bla bla can't reach_. But its a complex topic(a.l. for me) and I also don't want to spam a multiline comment.

thrimbor:
Looks good and works fine, merging.

Issue #362, opened 2020-06-09
OpenXDK XAudio sometimes doesn't work when running the application a second time


JayFoxRox:
If I use xaudio, then quit my application, and then use an XDK-made homebrew dashboard to relaunch my application, then xaudio suddenly won't cause interrupts anymore and audio doesn't seem to work.
This probably breaks many applications, because they might use audio for synchronization.

This needs to be investigated.

We actually do an AC97 cold-reset, so I'm not sure what other state might affect it.

I did check the [return value of KeConnectInterrupt](https://github.com/XboxDev/nxdk/blob/e17f3c54853b3139c06a13da67673d790989a35c/lib/hal/audio.c#L156) but it's always `TRUE`.

I also looked at some AC97 registers but found nothing of interest.

I did not check if the audio is also broken in the XDK-made dashboard.

Issue #363, opened 2020-06-09
OpenXDK XAudio does not deinitialize itself


JayFoxRox:
xaudio leaks memory; it also plays garbage while the application is quitting (because it will keep buffers alive, but they'll be overwritten / free'd by the kernel or next application).

A shutdown handler should be registered and a function for de-initializing AC97 should be added.

We should probably investigate #362 first, because even just calling `XAudioInit` shortly before shutting down actually seems to hide that issue.

Teufelchen1:
Related #449 and #458 - XAudio api is most likely deprecated soon®.

JayFoxRox:
Not strictly related: XAudio *API* is deprecated, but the implementation won't be.

So people should stop using XAudio directly, but the code will continue to be used, so all bugs are still valid.
These issues also affect users who use XAudio through SDL audio (which is the recommended way of using audio on nxdk).

PR #364, opened 2020-06-09
Fix many issues in OpenXDK XAudio

JayFoxRox:
**This code was tested very little; in particular, I did not test it with any existing OpenXDK XAudio applications. I also did not test the digital audio-out.**

Please test and review before merge. For testing, I've been using https://github.com/JayFoxRox/nxdk/pull/78 which can create some metadata for audacity.

---

## Synchronize analog and digital buffers

The interrupts for analog audio come about 16 samples before digital audio interrupts. This means we might mark a buffer as processed (by calling the user callback) before the AC97 has finished processing it.

We also cleared the interrupts for the digital audio in the interrupt handler of analog audio. This could lead to race conditions.

This commit keeps counters so we always know how many buffers there are per bus master.
Only once both bus masters have processed the buffer, we call the callback.

## Only cause 1 callback per XAudioProvideSamples

When AC97 runs out of data (because no new buffer has been provided) it will actually keep repeating the last buffer. Each buffer completion causes an interrupt.

This means that submitting a single buffer usually caused infinitely many callbacks.
If the application submitted 2 buffers, it would not know if one of the buffers played twice, or if both buffers have played.

This code changes the callback behaviour:
If the last buffer is played it will have 4 (last valid buffer completion) in addition to the usual 8 (end of buffer) set as interrupt reason (MMIO register 0x116 / 0x176).
If this condition has been found, a single callback is generated. Further buffer completions are ignored until new buffers have been submitted and started processing.

Unfortunately, there might be a risk of losing buffers now: If the buffer is too small, I could imagine that AC97 fires 2 interrupts but the CPU might only see one of them (as it's not fast enough to clear the interrupt flag between the two). As such, the CPU might skip one callback. This is only a theoretical problem though. I'll create an issue about this after merge, so it can be investigated.

## Respect timing requirement of AC97 cold-reset

AMD AC97 docs say that cold-reset must be pulled to 0 for at least 1 microsecond.
Our current code never pulled it to zero and also didn't respect any timing.

This commit respects this.

*(This was supposed to prevent #362, but it did not work. It doesn't seem to have any practical effect)*

## Reset bus master registers at startup

Saw this in the documentation and figured I should add it.
It might be unnecessary (is cold-reset enough?).

*(This was supposed to prevent #362, but it did not work. It doesn't seem to have any practical effect)*

## Avoid recalculation of descriptor parameters

General cleanup.

## Rename nextDescriptorMod31 to nextDescriptor

General cleanup. Intention was correctness and readability.

## Improve use of the volatile keyword 

General cleanup. Intention was better support for optimizations and more stability.

### DIFF
--- a/lib/hal/audio.c
+++ b/lib/hal/audio.c
@@ -153,7 +153,11 @@ void XAudioInit(int sampleSizeInBits, int numChannels, XAudioCallback callback, 
 	memset((void *)&pac97device->pcmSpdifDescriptor[0], 0, sizeof(pac97device->pcmSpdifDescriptor));
 	memset((void *)&pac97device->pcmOutDescriptor[0], 0, sizeof(pac97device->pcmOutDescriptor));
 
-	// perform a cold reset
+	// perform cold reset
+	pac97device->mmio[0x12C>>2] &= ~2;
+	LARGE_INTEGER Interval;
+	Interval.QuadPart = -10;
+	KeDelayExecutionThread(KernelMode, FALSE, &Interval);
 	pac97device->mmio[0x12C>>2] |= 2;
 	
 	// wait until the chip is finished resetting...
--- a/lib/hal/audio.c
+++ b/lib/hal/audio.c
@@ -25,8 +25,10 @@
 
 #define AUDIO_IRQ 6
 
-static volatile int analogBufferCount;
-static volatile int digitalBufferCount;
+static volatile unsigned int analogBufferCount;
+static volatile unsigned int digitalBufferCount;
+static bool analogDrained;
+static bool digitalDrained;
 
 static KINTERRUPT InterruptObject;
 static KDPC DPCObject;
@@ -68,10 +70,14 @@ static BOOLEAN __stdcall ISR(PKINTERRUPT Interrupt, PVOID ServiceContext)
 		bool waitingForAnalog = (analogBufferCount > digitalBufferCount);
 
 		// Was the interrupt triggered because we were out of data?
-		if (analogInterrupt & 8) {
-			waitCompleted |= waitingForAnalog;
-			analogBufferCount--;
+		if ((analogInterrupt & 8) && !analogDrained) {
+			if (analogBufferCount > 0) {
+				waitCompleted |= waitingForAnalog;
+				analogBufferCount--;
+			}
 		}
+
+		analogDrained = analogInterrupt & 4;
 
 		*(volatile unsigned char *)0xFEC00116=0xFF; // clear all int sources
 	}
@@ -80,10 +86,14 @@ static BOOLEAN __stdcall ISR(PKINTERRUPT Interrupt, PVOID ServiceContext)
 		bool waitingForDigital = (digitalBufferCount > analogBufferCount);
 
 		// Was the interrupt triggered because we were out of data?
-		if (digitalInterrupt & 8) {
-			waitCompleted |= waitingForDigital;
-			digitalBufferCount--;
+		if ((digitalInterrupt & 8) && !digitalDrained) {
+			if (digitalBufferCount > 0) {
+				waitCompleted |= waitingForDigital;
+				digitalBufferCount--;
+			}
 		}
+
+		digitalDrained = digitalInterrupt & 4;
 
 		*(volatile unsigned char *)0xFEC00176=0xFF; // clear all int sources
 	}
@@ -167,6 +177,8 @@ void XAudioInit(int sampleSizeInBits, int numChannels, XAudioCallback callback, 
 	// reset buffer status
 	analogBufferCount = 0;
 	digitalBufferCount = 0;
+	analogDrained = false;
+	digitalDrained = false;
 	
 	// Register our ISR
 	vector = HalGetInterruptVector(AUDIO_IRQ, &irql);
--- a/lib/hal/audio.c
+++ b/lib/hal/audio.c
@@ -143,7 +143,7 @@ void XAudioInit(int sampleSizeInBits, int numChannels, XAudioCallback callback, 
 	MmLockUnlockBufferPages((PVOID)pac97device, sizeof(AC97_DEVICE), FALSE);
 
 	pac97device->mmio = (unsigned int *)0xfec00000;
-	pac97device->nextDescriptorMod31 = 0;
+	pac97device->nextDescriptor = 0;
 	pac97device->callback = callback;
 	pac97device->callbackData = data;
 	pac97device->sampleSizeInBits = sampleSizeInBits;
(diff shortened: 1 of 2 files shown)


Ryzee119:
I tested this in conjunction with https://github.com/XboxDev/nxdk-sdl/pull/27 running on a 1.4 Xbox with digital audio out (SPDIF)

I compiled these homebrew games and the 'crackling' audio that was audible was no longer present and there was no audible issues to my ears.
https://github.com/Ryzee119/SDL_audio_tests
https://github.com/Ryzee119/SDLPoPX
https://github.com/Ryzee119/opentyrian
https://github.com/Ryzee119/omnispeak

Interesting, the audio in the latest xqemu release from xqemu.com was no longer working with this PR. Omnispeak would hang, other tests just had no audio, but worked fine on hardware. Did not investigate this further.

I have not stress tested this and only played ~5 minutes. Will report back after some game time If I experience any issues or otherwise.

lib/hal/audio.c
```diff
@@ -109,26 +148,48 @@ void XAudioInit(int sampleSizeInBits, int numChannels, XAudioCallback callback,
...
 	while(!(pac97device->mmio[0x130>>2]&0x100))
 		;
 
+	// reset bus master registers for analog output
+	pb[0x11B] = (1 << 4) | (1 << 3) | (1 << 2) | (1 << 1);
+	while(pb[0x11B] & (1 << 1))
+		;
+
+	// reset bus master registers for digital output
+	pb[0x17B] = (1 << 4) | (1 << 3) | (1 << 2) | (1 << 1);
```
JayFoxRox, line 186: This can be done  at the same time as the other reset, to reduce wait times

lib/hal/audio.c
```diff
@@ -109,26 +148,48 @@ void XAudioInit(int sampleSizeInBits, int numChannels, XAudioCallback callback,
...
+	Interval.QuadPart = -10;
+	KeDelayExecutionThread(KernelMode, FALSE, &Interval);
 	pac97device->mmio[0x12C>>2] |= 2;
 	
 	// wait until the chip is finished resetting...
 	while(!(pac97device->mmio[0x130>>2]&0x100))
 		;
 
+	// reset bus master registers for analog output
+	pb[0x11B] = (1 << 4) | (1 << 3) | (1 << 2) | (1 << 1);
```
JayFoxRox, line 181: Should check if the cold-reset already resets these, too

lib/hal/audio.c
```diff
@@ -55,25 +61,58 @@ static void __stdcall DPC(PKDPC Dpc,
...
-	//KeInsertQueueDpc queues Dpc and returns TRUE if Dpc not already queued.
-	//Dpc will be queued only once. So only one Dpc is fired after ISRs cease fire.
-	//DPCs avoid crashes inside non reentrant user callbacks called by nested ISRs.
-	//CAUTION : if you use fpu in DPC you have to save & restore yourself fpu state!!!
-	//(fpu=floating point unit, i.e the coprocessor executing floating point opcodes)
+	unsigned char analogInterrupt = pb[0x116];
+	unsigned char digitalInterrupt = pb[0x176];
+
+	bool waitingForAnalog = (analogBufferCount > digitalBufferCount);
+	bool waitingForDigital = (digitalBufferCount > analogBufferCount);
```
JayFoxRox, line 71: This must be sampled later, otherwise if we aren't fast enough at handling an interrupt, then we might have to handle 2 interrupts at the same time.
In that situation, none of them would be waiting (because the counts are the same), but both of them would decrement the buffercount (without marking `waitCompleted` which would cause a buffer to be freed).

So to avoid this, this has to be done when the `analogBufferCount` has already been decremented (thereby causing `waitingForDigital` to be `true`, thereby causing `waitCompleted` to be `true`).

lib/hal/audio.c
```diff
@@ -55,25 +61,59 @@ static void __stdcall DPC(PKDPC Dpc,
...
-	//DPCs avoid crashes inside non reentrant user callbacks called by nested ISRs.
-	//CAUTION : if you use fpu in DPC you have to save & restore yourself fpu state!!!
-	//(fpu=floating point unit, i.e the coprocessor executing floating point opcodes)
+	unsigned char analogInterrupt = pb[0x116];
+	unsigned char digitalInterrupt = pb[0x176];
+
+	bool waitCompleted = false;
+
+	if (analogInterrupt) {
+
```
thrimbor, line 73: nit: This newline feels weird.

lib/hal/audio.c
```diff
@@ -55,25 +61,59 @@ static void __stdcall DPC(PKDPC Dpc,
...
+		digitalDrained = digitalInterrupt & 4;
+
+		pb[0x176]=0xFF; // clear all int sources
+	}
+
+
+	// If a buffer was consumed by analog and digital output, we ask for a DPC
+	if (waitCompleted) {
+			KeInsertQueueDpc(&DPCObject,NULL,NULL); //calls user callback soon
```
thrimbor, line 109: nit: Indentation

thrimbor:
I've tested it now, and the results look good. Even when running SDLPoPX without optimizations, where sound previously crackled like crazy, it now sounds perfect. Seems to be perfectly stable as well.
I think we can merge this (and the SDL PR) very soon.

lib/hal/audio.c
JayFoxRox: This came from the original code. I've re-arranged it, and added another commit to also improve the `DPC` function, which had similar issues.

It's still not ideal, but better. The goal of this PR is functionality, not style.

lib/hal/audio.c
JayFoxRox: Agreed; removed the empty lines.

JayFoxRox:
> There are a couple of things that I think could be changed, [...] using `unsigned char` instead of `uint8_t` when talking to hardware, formatting etc.

I agree. But formatting / cleanup wasn't really a goal here.

If I had to design a new audio driver from scratch, it would be radically different. This is really just to have the old code with the old API (which is broken by design) working.

Fixing the types would also require other changes like moving data from headers into the files.
There are also worse offenders like storing pointers in `unsigned int`.

Many problems also still remain, such as #363 and #362.

lib/hal/audio.c
JayFoxRox: At least in QEMU, the cold-reset is a no-op. So I assume it only resets the codec.
I'll create an issue about this after merge, so we can check.

For now, assume that this is required.

lib/hal/audio.c
JayFoxRox: I'm not planning to change this, because I don't want to force another round of testing.
I'll need to look at AC97 for XQEMU anyway, and I can clean this up then.
During a cleanup pass, I might also define constants for these fields etc. (my code just mirrors the existing style - which was also ugly), if I don't end up making an all new API altogether.

Issue #311, opened 2020-04-07
`NXDK_CXX` Makefile option is not documented


JayFoxRox:
At least the hello++ sample should contain `NXDK_CXX = y` so users are made aware that C++ support needs to be enabled explicitly (at least for now).

Teufelchen1:
I'm unable to find _any_ information about `NXDK_CPP`. In fact, when searching through the repo, I found that `NXDK_CPP` is nowhere mentioned. How does setting a var that never gets checked lead to C++ support? What am I missing?
Maybe its some Makefile magic, like `($projectname)_CPP`?

JayFoxRox:
I made an error when creating this issue; the option is actually called `NXDK_CXX`, not `NXDK_CPP`:
https://github.com/XboxDev/nxdk/blob/9a5e86d860eccc144000bdddb7a13f7dc1ff6df3/Makefile#L93-L95

JayFoxRox:
Closed by #366 - implicit documentation is good enough for now. There's also #367 as a catch-all issue now for explicit documentation.

PR #366, opened 2020-06-13
samples: Enable C++ support for hello++

Teufelchen1:
This addresses #311 by enabling `NXDK_CXX = y` in the hello++ `Makefile`

### DIFF
--- a/samples/hello++/main.cpp
+++ b/samples/hello++/main.cpp
@@ -1,25 +1,22 @@
 #include <hal/debug.h>
 #include <hal/video.h>
+#include <string>
+#include <vector>
 #include <windows.h>
-
-class NumberClass{
-private:
-  int hiddenValue = 0;
-public:
-  NumberClass() {hiddenValue = 5;};
-  NumberClass(int v) : hiddenValue(v) {};
-  int getValue() { return hiddenValue; };
-};
 
 int main(void) {
   XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
 
-  NumberClass defaultFive;
-  NumberClass customSix(6);
+  std::vector<std::string> words;
+  words.emplace_back("Hello");
+  words.emplace_back(" ");
+  words.emplace_back("nxdk!");
+  words.emplace_back("\n");
 
   while (true) {
-    debugPrint("Value of defaultFive: %i\n", defaultFive.getValue());
-    debugPrint("Value of customSix  : %i\n", customSix.getValue());
+    for (auto& word : words) {
+      debugPrint("%s", word.c_str());
+    }
     Sleep(2000);
   }
 
--- a/samples/hello++/Makefile
+++ b/samples/hello++/Makefile
@@ -2,4 +2,5 @@ XBE_TITLE=hello++
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.cpp
 NXDK_DIR = $(CURDIR)/../..
+NXDK_CXX = y
 include $(NXDK_DIR)/Makefile


JayFoxRox:
Ideally it would also add some dependency on the C++ STL. The previous C++ sample was kind of "forced" anyway (we added some C++ just to make it more C++-ish, but we couldn't use the STL because we didn't have it).
Now we can simply show a "Hello World" using an `std::vector` (which actually requires `NXDK_CXX = y`).

Teufelchen1:
Tested on hardware and also asserted that it does **not** compile without `NXDK_CXX = y`.

samples/hello++/main.cpp
```diff
@@ -1,27 +1,21 @@
 #include <hal/debug.h>
 #include <hal/video.h>
+#include <string>
+#include <vector>
```
Teufelchen1, line 4: I was unsure in which order I should put the includes. Settled for: nxdk-specific over cpp-specific and alphabetical.

samples/hello++/main.cpp
```diff
@@ -1,27 +1,21 @@
...
-  NumberClass customSix(6);
-
-  while (true) {
-    debugPrint("Value of defaultFive: %i\n", defaultFive.getValue());
-    debugPrint("Value of customSix  : %i\n", customSix.getValue());
-    Sleep(2000);
+  while (1) {
+    for (auto& word : words) {
+      debugPrint("%s", word.c_str());
+      Sleep(2000);
```
JayFoxRox, line 17: I'd probably move this out of the loop? Especially because " " would not be a visible addition

samples/hello++/main.cpp
```diff
@@ -1,27 +1,21 @@
...
-  NumberClass(int v) : hiddenValue(v) {};
-  int getValue() { return hiddenValue; };
-};
-
 int main(void) {
   XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+  std::vector<std::string> words;
+  words.emplace_back("Hello");
+  words.emplace_back(" ");
+  words.emplace_back("World\n");
```
JayFoxRox, line 12: nit: I'd either have `{"Hello ", "World\n"}` or `{"Hello", " ", "World", "\n"}`
I prefer the second variant.

samples/hello++/main.cpp
```diff
@@ -1,27 +1,21 @@
...
-  int hiddenValue = 0;
-public:
-  NumberClass() {hiddenValue = 5;};
-  NumberClass(int v) : hiddenValue(v) {};
-  int getValue() { return hiddenValue; };
-};
-
 int main(void) {
   XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+  std::vector<std::string> words;
```
JayFoxRox, line 9: I'd add an empty line before this, to add some structure to the code. Other instances, too.

samples/hello++/main.cpp
```diff
@@ -1,27 +1,21 @@
...
+  words.emplace_back("World\n");
 
-  NumberClass defaultFive;
-  NumberClass customSix(6);
-
-  while (true) {
-    debugPrint("Value of defaultFive: %i\n", defaultFive.getValue());
-    debugPrint("Value of customSix  : %i\n", customSix.getValue());
-    Sleep(2000);
+  while (1) {
```
JayFoxRox, line 14: The old code used C11 `bool` from `<stdbool.h>`.
Ironically, `bool` is a core feature in C++ (so no header required), but it isn't used here.

Teufelchen1: Oh wasn't aware of that. Thats a must-change imo.

samples/hello++/main.cpp
JayFoxRox, line 12: I just realized that the "hello" samples says "Hello nxdk!\n"; we should probably mirror this?

samples/hello++/main.cpp
```diff
@@ -1,25 +1,22 @@
...
-
 int main(void) {
   XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
 
-  NumberClass defaultFive;
-  NumberClass customSix(6);
+  std::vector<std::string> words;
+  words.emplace_back("Hello");
+  words.emplace_back(" ");
+  words.emplace_back("nxdk!\n");
```
JayFoxRox, line 13: This would cause 2 `\n` now.
I'd remove the one from this line.

Teufelchen1: oh, how embarrassing 😨 sorry.

PR #371, opened 2020-06-18
samples: Harmonise flag assignment style in Makefiles

Teufelchen1:
Hi! 👋 
I spotted that the assignment style for various flags(mainly `XBE_TITLE`) in the Makefiles is irregular. This PR harmonise it by making sure there is always exactly one whitespace before and exactly one whitespace after a given `=`. 
I rebuild all samples without an issue. I did not run them.

### DIFF
--- a/samples/gamma-fade/Makefile
+++ b/samples/gamma-fade/Makefile
@@ -1,4 +1,4 @@
-XBE_TITLE=gamma-fade
+XBE_TITLE = gamma-fade
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.c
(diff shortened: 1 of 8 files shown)


GXTX:
Low risk and appears to be correct.

JayFoxRox:
nit: "Make [..] consistent" or "Improve style consistency for [...]" would be better wordings for the commit title.

#372 can also be done in a another commit in this PR, or a future PR.

Teufelchen1:
I would opt against also fixing #372, I gave it a short test and it looks like solving that issue requires more then pure style/cosmetic fixes. Hence, I consider it out of scope for this PR.

thrimbor:
Looks good, merging.

Issue #318, opened 2020-04-11
"Notable features" in README is not very complete or catchy


JayFoxRox:
At the time of writing, the README considers only the LLVM base and 3D graphics as notable features.

However, what it should mention, is probably:

- Portable toolchain that works on modern versions of Windows, macOS and Linux.
- Modern C / C++ standards and compiler features.
- Supports popular APIs like Windows API and BSD sockets.
- SDL2 support for input, audio and 2D graphics.
- Custom API for 3D graphics using NVIDIA-designed shader-languages (with additional Xbox extensions).
- Open-Source drivers which can be modified to get the most out of the hardware.
- Modifiable startup code, for as much system control as necessary.
- Supported by an active community that can help with problems and responds to bug reports.

---

I'd also say that the "Status"  is a bit misleading.
While "nxdk is currently in early stages of development" is probably still true, it can be used very nicely in some domains already.
I wonder if we should remove the "Status" information altogether. I think it just scares people away.

Instead, we could link to https://github.com/XboxDev/nxdk/wiki/Projects-using-nxdk. For the current capabilities we could link to the samples/ directory; also warn users about our open issues (by linking to that tab).

PR #374, opened 2020-07-01
docs: Update README with new features list

Teufelchen1:
Hi 👋 
This PR enhances the feature- and status-section in the `README.md` as proposed in #318

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,15 @@ nxdk is a software development kit for the original Xbox. nxdk is a revitalizati
 It is maintained by the [XboxDev](https://github.com/XboxDev/XboxDev) community.
 
 Notable features:
+- Portable toolchain that works on modern versions of Windows, macOS and Linux.
 - No complicated cross-compiling or big library dependencies! Builds with `make` and just needs standard tools and llvm.
-- 3D graphics! nxdk includes and expands on pbkit, a library for interfacing with the Xbox GPU.
+- Modern C / C++ standards and compiler features.
+- Supports popular APIs like Windows API and BSD sockets.
+- SDL2 support for input, audio and 2D graphics.
+- Custom API for 3D graphics using NVIDIA-designed shader-languages (with additional Xbox extensions).
+- Open-Source drivers which can be modified to get the most out of the hardware.
+- Modifiable startup code, for as much system control as necessary.
+- Supported by an active community that can help with problems and responds to bug reports.
 
 Build Status
 ------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,16 +4,24 @@ nxdk is a software development kit for the original Xbox. nxdk is a revitalizati
 It is maintained by the [XboxDev](https://github.com/XboxDev/XboxDev) community.
 
 Notable features:
+- Portable toolchain that works on modern versions of Windows, macOS and Linux.
 - No complicated cross-compiling or big library dependencies! Builds with `make` and just needs standard tools and llvm.
```
JayFoxRox, line 8: This sounds weird and should be touched.

It can be misinterpreted as: "nxdk is not a cross-compiler". I also don't think that "make" is a good feature, and I don't think "standard tools" aren't really necessary. llvm is the only interesting bit, but could be phrased differently.

Could be something like "Interoperability with standard tooling like gcc and Microsoft Visual C++!" (which I'm not sure about, because nobody tested it yet). Or it could be "LLVM / clang based, for interoperability with other programming languages" (which I'm also not sure about, because nobody tried that yet either; my attempts with C# were unsuccessful, but largely due to shortcomings in the projects I tried).

I think I'd just remove this sentence altogether, or rephrase it to what it seems to imply: "Easy installation."

README.md
```diff
@@ -4,16 +4,24 @@ nxdk is a software development kit for the original Xbox. nxdk is a revitalizati
...
+- Supported by an active community that can help with problems and responds to bug reports.
 
 Build Status
 ------
 ![CI Status for "Build Samples"](https://github.com/XboxDev/nxdk/workflows/Build%20Samples/badge.svg?branch=master&event=push)
 
 Status
 ------
-nxdk is currently in early stages of development.
+While nxdk still is in early stages of development, it can already be used in many projects.
```
JayFoxRox, line 23: This sentence is weird. "It can be used in projects" or "It is already being used in many projects". Maybe also: "Useful for a variety of projects" or something.
But "can be used in many projects" sounds weird to me.

README.md
```diff
@@ -4,16 +4,24 @@ nxdk is a software development kit for the original Xbox. nxdk is a revitalizati
...
+Take a look at the [list of projects](https://github.com/XboxDev/nxdk/wiki/Projects-using-nxdk) that are build on top of nxdk. Additionally, the [provided samples](https://github.com/XboxDev/nxdk/tree/master/samples) show how to use common features.
```
JayFoxRox, line 24: I feel like we should also add a link to the critical issues here, and discourage binary builds until these are fixed.

If we don't have one yet, we should add a critical issue about not listing the licenses of each component / not providing a license-package (or something like SPDX).

PR #376, opened 2020-07-05
vcruntime: Implement thread-safe statics initialization

### LINKED ISSUE #238: Thread-safe initialization of static variables is disabled in C++
C++11 introduced the guarantee that static variables are initialized in a thread-safe way. nxdk currently does not support the necessary functionality enabling this feature, which is why it got disabled in #219.

I plan to fix this myself after my libc++ work has landed.

thrimbor:
The C++11 standard requires statics-initialization to be thread-safe. We had this behavior disabled till now, since we lacked the code to support the way MSVC (and clang when building for Windows) implemented it.

This code adds support for this and removes the compiler parameter that disabled it. The code is derived from and mostly identical to the one in the ReactOS CRT, but it contains two notable fixes, which I figured out by doing a bit of RE:
* `_Init_global_epoch` and `_Init_thread_epoch` get their correct starting values
* `_Init_thread_epoch` gets properly updated, so `_Init_thread_header` only gets called once per thread after initialization is done (clang automatically generates code that checks the epoch values)

I have some terrible code [here](https://gist.github.com/thrimbor/c32d373df6ebd250076a312534907015) that demonstrates that the linker errors are gone and initialization still works.

Closes #238.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
                -Wno-ignored-attributes -DNXDK -D__STDC__=1
 NXDK_ASFLAGS = -target i386-pc-win32 -march=pentium3 \
                -nostdlib -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt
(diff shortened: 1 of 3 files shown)


mborgerson:
Thanks

It's worth mentioning that this PR employs SPDX license identifiers in this newly added file. I'll merge, but it's probably worth having a discussion offline on whether we plan to update the rest of the tree to use this format.

PR #377, opened 2020-07-06
winapi: Implement GetExitCodeThread()

thrimbor:
This implements `GetExitCodeThread()`, which is required by my upcoming PDCLib PR that will simplify the C11 threading code by implementing it on top of the Win32 thread API. The implementation is roughly based on the code in `thrd_join()`.

Quick testcase:
```c
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>

DWORD WINAPI threadFunc (LPVOID lpParam) {
    return 0xDEADBEEF;
}

int main(void)
{

    HANDLE t = CreateThread(NULL, 0, threadFunc, NULL, 0, NULL);
    Sleep(2000);
    DWORD exitcode;
    GetExitCodeThread(t, &exitcode);
    debugPrint("ec: %x\n", exitcode);

    while(1) {
        Sleep(2000);
    }

    return 0;
}
```

### DIFF
--- a/lib/winapi/thread.c
+++ b/lib/winapi/thread.c
@@ -89,6 +89,26 @@ VOID ExitThread (DWORD dwExitCode)
     PsTerminateSystemThread(dwExitCode);
 }
 
+BOOL GetExitCodeThread (HANDLE hThread, LPDWORD lpExitCode)
+{
+    PETHREAD thread;
+    NTSTATUS status;
+
+    status = ObReferenceObjectByHandle(hThread, NULL, (PVOID *)&thread);
+    if (!NT_SUCCESS(status)) {
+        SetLastError(RtlNtStatusToDosError(status));
+        return FALSE;
+    }
+
+    if (thread->Tcb.HasTerminated) {
+        *lpExitCode = thread->ExitStatus;
+    } else {
+        *lpExitCode = STILL_ACTIVE;
+    }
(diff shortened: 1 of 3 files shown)


lib/winapi/thread.c
```diff
@@ -89,6 +89,22 @@ VOID ExitThread (DWORD dwExitCode)
...
+    PETHREAD thread;
+    NTSTATUS status;
+
+    status = ObReferenceObjectByHandle(hThread, NULL, (PVOID *)&thread);
+    if (!NT_SUCCESS(status)) {
+        SetLastError(RtlNtStatusToDosError(status));
+        return FALSE;
+    }
+
+    *lpExitCode = thread->ExitStatus;
```
JayFoxRox, line 103: Is this initialized to `STILL_ACTIVE`? If not, should we do this, or do we need to tag threads differently?

thrimbor: Ah, this somehow managed to slip by me. We probably shouldn't modify this kernel data, I fixed it by checking `thread->Tcb.HasTerminated` and returning `STILL_ACTIVE` when it's not true. I checked the result of `GetExitCodeThread` (both return value and exit code) before and after the thread terminated, and it works fine now. I also compared it to Windows 7 now to make sure it's supposed to return true even when the thread hasn't terminated yet.

PR #378, opened 2020-07-06
lwip: Add DHCP options to imitate XDK DHCP requests

thrimbor:
I talked about this a bit in Discord already.
Before these changes, my router would treat the DHCP requests done by nxdk apps as coming from a different device, despite coming from the same MAC address as requests done by XDK apps. This lead to my router assigning a different IP address, which made it a bit annoying to connect.

These changes fix that by using a hook provided by lwIP to add two of the three DHCP option fields the XDK uses: The "Client identifier", which just contains an address type (0x01 for Ethernet) and the MAC address (this options seems to be the most important one to achieve this behavior), and the "Vendor class identifier", which just contains the non-terminated string "XBOX 1.0".
There is a third option that the XDK uses, "Requested IP Address", which contains the last IP the Xbox received via DHCP. I did not add this one though, since it isn't required to get the behavior I wanted and would require reading and parsing the Config Sector.

I tested these changes by running the `httpd_bsd` sample on my Xbox and capturing and comparing the packets with Wireshark.

### DIFF
--- a/lib/net/nforceif/include/lwipopts.h
+++ b/lib/net/nforceif/include/lwipopts.h
@@ -422,6 +422,59 @@
 
 /*
    ---------------------------------------
+   ---------- Hook options ---------------
+   ---------------------------------------
+*/
+
+/**
+ * LWIP_HOOK_DHCP_APPEND_OPTIONS(netif, dhcp, state, msg, msg_type, options_len_ptr):
+ * Called from various dhcp functions when sending a DHCP message.
+ * This hook is called just before the DHCP message trailer is added, so the
+ * options are at the end of a DHCP message.
+ * Signature:\code{.c}
+ *   void my_hook(struct netif *netif, struct dhcp *dhcp, u8_t state, struct dhcp_msg *msg,
+ *                u8_t msg_type, u16_t *options_len_ptr);
+ * \endcode
+ * Arguments:
+ * - netif: struct netif that the packet will be sent through
+ * - dhcp: struct dhcp on that netif
+ * - state: current dhcp state (dhcp_state_enum_t as an u8_t)
+ * - msg: struct dhcp_msg that will be sent
+ * - msg_type: dhcp message type to be sent (u8_t)
+ * - options_len_ptr: pointer to the current length of options in the dhcp_msg "msg"
+ *                    (must be increased when options are added!)
+ *
+ * Options need to appended like this:
+ *   LWIP_ASSERT("dhcp option overflow", *options_len_ptr + option_len + 2 <= DHCP_OPTIONS_LEN);
+ *   msg->options[(*options_len_ptr)++] = &lt;option_number&gt;;
+ *   msg->options[(*options_len_ptr)++] = &lt;option_len&gt;;
+ *   msg->options[(*options_len_ptr)++] = &lt;option_bytes&gt;;
+ *   [...]
+ */
+/*
+   Add DHCP options to imitate XDK behavior, making DHCP servers treat nxdk the same,
+   i.e. recognizing requests as coming from the same device and assigning the same
+   IP address.
+*/
+#define LWIP_HOOK_DHCP_APPEND_OPTIONS(netif, dhcp, state, msg, msg_type, options_len_ptr) \
+    *options_len_ptr = dhcp_option(*options_len_ptr, msg->options, DHCP_OPTION_CLIENT_ID, 7); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, 0x01); \
+    for (int _i = 0; _i < 6; _i++) { \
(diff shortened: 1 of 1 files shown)


lib/net/nforceif/include/lwipopts.h
```diff
@@ -47,6 +47,24 @@
...
+    \
+    *options_len_ptr = dhcp_option(*options_len_ptr, msg->options, DHCP_OPTION_US, 8); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, 'X'); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, 'B'); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, 'O'); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, 'X'); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, ' '); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, '1'); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, '.'); \
+    *options_len_ptr = dhcp_option_byte(*options_len_ptr, msg->options, '0');
```
JayFoxRox, line 66: I didn't check the upstream file again, but I believe I'd rather add all of this where it is in the upstream `opt.h`, including the comment. We should try to minimize our diff so we can react when upstream changes something dumb (like macro-parameter order).

thrimbor: I did that now. I also added the comment explaining the additions, which I forgot to commit earlier.

JayFoxRox: nit: That comment from the edit could have been part of the code-comment

GXTX:
>There is a third option that the XDK uses, "Requested IP Address", which contains the last IP the Xbox received via DHCP. I did not add this one though, since it isn't required to get the behavior I wanted and would require reading and parsing the Config Sector.

I understand that it isn't required for similar functionality and adds hitting the drive to get the information but could we maybe have an flag or an option *to* do this and not make it a requirement?

thrimbor:
> > There is a third option that the XDK uses, "Requested IP Address", which contains the last IP the Xbox received via DHCP. I did not add this one though, since it isn't required to get the behavior I wanted and would require reading and parsing the Config Sector.
> 
> I understand that it isn't required for similar functionality and adds hitting the drive to get the information but could we maybe have an flag or an option _to_ do this and not make it a requirement?

I'd like to avoid adding it before we have a proper API for setting up the network and querying the config sector. lwIP definitely shouldn't touch the config sector itself imho.

lib/net/nforceif/include/lwipopts.h
```diff
@@ -420,6 +420,59 @@
...
+ *   msg->options[(*options_len_ptr)++] = &lt;option_bytes&gt;;
+ *   [...]
+ */
+/*
+   Add DHCP options to imitate XDK behavior, making DHCP servers treat nxdk the same,
+   i.e. recognizing requests as coming from the same device and assigning the same
+   IP address.
+*/
+#define LWIP_HOOK_DHCP_APPEND_OPTIONS(netif, dhcp, state, msg, msg_type, options_len_ptr) \
+    *options_len_ptr = dhcp_option(*options_len_ptr, msg->options, DHCP_OPTION_CLIENT_ID, 7); \
```
mborgerson, line 460: Generally I'd like to avoid macro code blocks like this. Similar to how there's a `dhcp_option_hostname` function, I think this code could naturally be broken up into two functions, e.g. `dhcp_option_client_id`, `dhcp_option_vendor_class_id`, the vendor string converted to a loop over a `char *dhcp_option_vendor_class_id_str`, and this hook macro would simply call a third, wrapper function to append all the args....or something like that. Upstream LWIP might even want this feature.

thrimbor: Usually I'd prefer to put this into functions, too.
However, `lwipopts.h` is included in lots of different places, of which not all include the necessary headers to have all types available, so I can't just move the code into a static function here, unless I make sure to include all those headers, which I'd like to avoid as `lwipopts.h` isn't meant to do that.
Even if I put the function into a separate C-file (which I'd like to avoid, too), I'd still have to pull in those headers for the declaration.

mborgerson: > However, lwipopts.h is included in lots of different places, of which not all include the necessary headers to have all types available, so I can't just move the code into a static function here, unless I make sure to include all those headers, which I'd like to avoid as lwipopts.h isn't meant to do that.

I agree it's not ideal, but this macro definition is specifically designed to add code, so pulling in the headers in the appropriate place would be acceptable in my opinion. Frankly, it would probably just be better to patch LWIP to add this natively, instead of adding as a hook like this. For now, let's just keep it simple and use the approach you've submitted and we can refactor it later.

PR #388, opened 2020-07-23
Allocate framebuffer when XVideoSetMode is called

### LINKED ISSUE #149: vmCurrent uninitialized before XVideoSetMode call

### LINKED ISSUE #169: `XVideoInit` never allocated the framebuffer memory that it advertises via `XVideoGetFB`

### LINKED ISSUE #170: `XVideoGetFB` can return outdated address

### LINKED ISSUE #269: hal/video.c variables should be static

thrimbor:
This supersedes #355.

The commits are mostly the same - I only addressed the review comments from #355 and modified the newly introduced functions to use `(void)` instead of an empty parameter list.

I tested the changes by running the samples as well as a slightly modified (to remove the `_fb` reference in it) SDLPoPX.

Closes #170.
Closes #169.
Closes #149.
Closes #269.

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -35,15 +35,15 @@
 #define VIDEO_R5G6B5				0x00000011
 #define VIDEO_A8R8G8B8				0x00000012
 
-unsigned char*	framebufferMemory = NULL;
-unsigned char*	_fb = NULL;
-DWORD			dwEncoderSettings 	= 0;
-VIDEO_MODE		vmCurrent = { 0, 0, 0, 0 };
-int			flickerLevel		= 5;
-BOOL			flickerSet		= FALSE;
-BOOL			softenFilter		= TRUE;
-BOOL			softenSet		= FALSE;
-GAMMA_RAMP_ENTRY	gammaRampEntries[256];
+static unsigned char*	framebufferMemory = NULL;
+static unsigned char*	_fb = NULL;
+static DWORD			dwEncoderSettings 	= 0;
+static VIDEO_MODE		vmCurrent = { 0, 0, 0, 0 };
+static int			flickerLevel		= 5;
+static BOOL			flickerSet		= FALSE;
+static BOOL			softenFilter		= TRUE;
+static BOOL			softenSet		= FALSE;
+static GAMMA_RAMP_ENTRY	gammaRampEntries[256];
 
 static KINTERRUPT InterruptObject;
 static KDPC DPCObject;
@@ -62,7 +62,7 @@ typedef struct _VIDEO_MODE_SETTING
 	DWORD dwFlags;
 } VIDEO_MODE_SETTING;
 
-VIDEO_MODE_SETTING vidModes[] =
+static VIDEO_MODE_SETTING vidModes[] =
 {
  {0x44030307,640,480,50,VIDEO_REGION_PAL,AV_PACK_STANDARD}, //640x480 PAL 50Hz
  {0x44040408,720,480,50,VIDEO_REGION_PAL,AV_PACK_STANDARD}, //720x480 PAL 50Hz
@@ -122,7 +122,7 @@ VIDEO_MODE_SETTING vidModes[] =
  {0x04020204,720,480,60,VIDEO_REGION_NTSCJ,AV_PACK_SVIDEO}, //720x480 NTSCJ 60Hz
 };
 
-int iVidModes = sizeof(vidModes) / sizeof(VIDEO_MODE_SETTING);
+static int iVidModes = sizeof(vidModes) / sizeof(VIDEO_MODE_SETTING);
 
 static void __stdcall DPC(PKDPC Dpc,
 PVOID DeferredContext,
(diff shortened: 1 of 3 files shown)


lib/hal/video.c
```diff
@@ -35,15 +35,15 @@
...
-VIDEO_MODE		vmCurrent;
-int			flickerLevel		= 5;
-BOOL			flickerSet		= FALSE;
-BOOL			softenFilter		= TRUE;
-BOOL			softenSet		= FALSE;
-GAMMA_RAMP_ENTRY	gammaRampEntries[256];
+static unsigned char*	framebufferMemory = NULL;
+static unsigned char*	_fb = NULL;
+static DWORD			dwEncoderSettings 	= 0;
+static VIDEO_MODE		vmCurrent = { 0, 0, 0, 0 };
```
GXTX, line 41: nit: add additional spaces to keep aligned

Ryzee119:
Looks good to me! I haven't retested from my previous tests detailed in #355 but that should not have changed.

nits:
Tabs before `=` which is not consistent with rest of document
https://github.com/XboxDev/nxdk/blob/468eab88a60bfc7d58c9725e32f5133f300c7a3c/lib/hal/debug.c#L19-L20

thrimbor:
I don't think we should combine the functional changes with formatting fixes tbh - I'd rather see them done in a separate commit and applied to the whole file.

dracc:
Tested with both NevolutionX and Peanut-GB. Worked great. :+1:

PR #389, opened 2020-07-27
libc: Provide math.h / string.h overloads expected by libc++

thrimbor:
About two hours ago lorecast162 mentioned issues in Discord with using `std::abs` with a parameter of type `std::intmax_t`. We tracked it down to libc++ expecting the C standard library to provide the `abs` overloads for `long` and `long long` when building C++ code.
I went through libc++'s `math.h` and added all expected overloaded functions to our PDCLib extension header for `stdlib.h` (which is the header libc++ expects to find these functions in). The overloarded functions added are basically identical to the ones libc++ provides for other platforms.

I then searched for other occurrences of this, and found similar issues in `string.h`. `_NXDK_PREFERRED_OVERLOAD` is necessary because otherwise the compiler doesn't allow overloading these functions as only the return type differs for some of them.

### DIFF
--- a/lib/xboxrt/libc_extensions/stdlib_ext_.h
+++ b/lib/xboxrt/libc_extensions/stdlib_ext_.h
@@ -35,3 +35,37 @@ int mbtowc (wchar_t *pwc, const char *string, size_t n);
 #ifdef __cplusplus
 }
 #endif
+
+
+#ifdef __cplusplus
+#if (__cplusplus < 201103L)
+#define _NXDK_NOEXCEPT throw()
+#else
+#define _NXDK_NOEXCEPT noexcept
+#endif
+
+// libc++ expects MSVCRT to provide these overloads
+
+inline static long abs (long __x) _NXDK_NOEXCEPT
+{
+    return labs(__x);
+}
+
+inline static long long abs (long long __x) _NXDK_NOEXCEPT
+{
+    return llabs(__x);
+}
+
+inline static ldiv_t div (long __x, long __y) _NXDK_NOEXCEPT
+{
+    return ldiv(__x, __y);
+}
+
+inline static lldiv_t div (long long __x, long long __y) _NXDK_NOEXCEPT
+{
+    return lldiv(__x, __y);
+}
+
+#undef _NXDK_NOEXCEPT
+
+#endif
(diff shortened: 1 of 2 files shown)


lib/xboxrt/libc_extensions/string_ext_.h
```diff
@@ -33,3 +33,88 @@ __attribute__((deprecated)) static int stricmp (const char *s1, const char *s2)
 #ifdef __cplusplus
 }
 #endif
+
+
+#ifdef __cplusplus
+// libc++ expects MSVCRT to provide these overloads
+
+#ifndef _NXDK_PREFERRED_OVERLOAD
+#define _NXDK_PREFERRED_OVERLOAD __attribute__ ((__enable_if__(true, "")))
```
JayFoxRox, line 42: Do we use this `_NXDK` stuff elsewhere too?

I don't see this as being specific to nxdk.

I feel like this should be using a `_PDCLIB` prefix (or whatever upstream PDCLIB would use) and that might even be a good candidate to upstream in the future, or to keep it neutral for re-use in other projects.
If there are collisions, maybe `_PDCLIB_EXT` or similar.

lib/xboxrt/libc_extensions/stdlib_ext_.h
```diff
@@ -35,3 +35,29 @@ int mbtowc (wchar_t *pwc, const char *string, size_t n);
 #ifdef __cplusplus
 }
 #endif
+
+
+#ifdef __cplusplus
+// libc++ expects MSVCRT to provide these overloads
+
+inline static long abs (long __x) noexcept
```
JayFoxRox, line 43: `noexcept` requires C++11. I can see myself compiling C++98 code though.
It's my understanding that these overloads were also added at specific C++ versions. So this needs standard version checks and potentially dropping the noexpect or using a compiler extension for older versions.

thrimbor: I added some preprocessor code to use the old way (`throw()`) if not using at least C++11.
Keep in mind that libc++ headers are only compatible to language version 11 or higher, so if the code actually requires to be built with a lower language version you'll hit incompatibilities.

lib/xboxrt/libc_extensions/string_ext_.h
thrimbor: Not yet afaik.
I'd say these are specific to nxdk, as they're meant to provide the overloads that libc++ expects from the MSVC runtime library.

I don't think we should use the `PDCLib` namespace here - it's not a direct part of PDCLib, and I'd like to avoid potential naming conflicts.

JayFoxRox: I didn't have time to respond to this before this got merged. But I disagree with this.
These functions aren't nxdk specific at all, and they should have been prefixed differently.

PR #391, opened 2020-08-11
vcruntime: Fix and enable RTTI

### LINKED ISSUE #237: RTTI is unavailable in C++
It is currently not possible to use run-time type information in nxdk, which is why it was disabled in #219. Trying to enable RTTI will cause unresolved references to `type_info`'s vftable as soon as virtual functions are used.

thrimbor:
This fixes the `lld: error: undefined symbol: const type_info::'vftable'` linker error that would usually appear when enabling RTTI and linking a program that instantiates a class containing virtual methods, and enables RTTI. This is enough to give it a quick test:
```
class A {
  public:
  virtual ~A() {};
};

A a;
```

This is achieved by the following changes:
* Build the `type_info` class with `-fno-rtti` to prevent it from emitting a recursive reference to its vftable
* Implement the empty virtual destructor which acts as the key function
* Add a hidden constructor, because LLVM will optimize away the vftable if the class is not instantiable.

Removing `const` in `__std_type_info_data` is not necessary to get this working, but is required for when we eventually implement name demangling, which will require this structure to be modifiable at runtime. Some missing `noexcept`'s were also fixed.

These changes are not enough to actually profit from having RTTI (such as using `typeid` and `dynamic_cast`), as these require functional SEH and C++ exception support to implement.

Fixes #237.

### DIFF
--- a/lib/xboxrt/vcruntime/vcruntime_typeinfo.h
+++ b/lib/xboxrt/vcruntime/vcruntime_typeinfo.h
@@ -12,8 +12,8 @@
 
 struct __std_type_info_data
 {
-    char const *_UndecoratedName;
-    char const _DecoratedName[1];
+    char *_UndecoratedName;
+    char _DecoratedName[1];
 };
 
 extern "C"
@@ -25,7 +25,7 @@ extern "C"
 class type_info
 {
 public:
-    virtual ~type_info ();
+    virtual ~type_info () noexcept;
 
     const char *name () const noexcept
     {
@@ -60,7 +60,11 @@ public:
     }
 
 private:
-    type_info (type_info const &) = delete;
+    // According to the standard, this constructor should not exist. However,
+    // not having any constructors causes LLVM to not emit a vftable.
+    type_info () noexcept;
+
+    type_info (type_info const &) noexcept = delete;
     type_info &operator= (type_info const &) = delete;
 
     mutable __std_type_info_data _Data;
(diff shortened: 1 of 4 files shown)


lib/xboxrt/vcruntime/vcruntime_typeinfo.h
```diff
@@ -60,7 +60,11 @@ class type_info
     }
 
 private:
-    type_info (type_info const &) = delete;
+    // According to the standard, this constructor should not exist. However,
+    // not having any constructors causes LLVM to not emit a vftable.
+    type_info () noexcept;
```
JayFoxRox, line 65: Please create an issue about this.

lib/xboxrt/vcruntime/vcruntime_typeinfo.h
```diff
@@ -12,8 +12,8 @@
 
 struct __std_type_info_data
 {
-    char const *_UndecoratedName;
-    char const _DecoratedName[1];
+    char *_UndecoratedName;
+    char _DecoratedName[1];
```
JayFoxRox, line 16: If we demangle at runtime, how will we know when to do it?
Wouldn't it have made more sense to keep it `const` and deal with it when implementing demangling?

lib/xboxrt/vcruntime/vcruntime_typeinfo.h
thrimbor: Done.

lib/xboxrt/vcruntime/vcruntime_typeinfo.h
thrimbor: > If we demangle at runtime, how will we know when to do it?
MS demangles the name when `type_info::name()` is called and `_UndecoratedName` is `NULL`. It allocates a buffer for the string, demangles, and puts the address of the string in `_UndecoratedName`.

> Wouldn't it have made more sense to keep it const and deal with it when implementing demangling?
I don't think there's a significant benefit in keeping or removing `const` at this moment - it just bothered me because it was incorrect.

PR #392, opened 2020-08-11
Implement Structured Exception Handling version 3

thrimbor:
This implements Structured Exception Handling, not to be confused with C++ exception handling. I have already teased this on Discord, [here's](https://gist.github.com/thrimbor/612dd63564f4add3481eb13d8daf0415) some sample code to demonstrate what it can do.

Note that it's a documented flaw of clang/LLVM to not be able to catch asynchronous exceptions in the frame they were thrown in (see https://clang.llvm.org/docs/MSVCCompatibility.html#abi-features) - this is why the above demonstration code uses a separate function for dereferencing the null pointer.

**edit:**
Some material on SEH:
[MSDN on _local_unwind2](https://docs.microsoft.com/en-us/previous-versions/ff798095(v=vs.140))
[Reversing Microsoft Visual C++ Part I: Exception Handling](http://www.openrce.org/articles/full_view/21) (mostly about C++ exceptions, but also some SEH stuff)
[A Crash Course on the Depths of Win32 Structured Exception Handling](https://web.archive.org/web/20170210065433/https://www.microsoft.com/msj/0197/exception/exception.aspx) (very useful, includes some pseudo-code)
[MSDN on _except_handler3](https://docs.microsoft.com/en-us/cpp/c-runtime-library/except-handler3?view=vs-2019)
[LLVM exception handling presentation](https://llvm.org/devmtg/2015-10/slides/KlecknerMajnemer-ExceptionHandling.pdf)


thrimbor:
I updated the PR to fix the merge conflict and added some links about SEH to the PR description.

lib/xboxrt/vcruntime/excpt.cpp
```diff
@@ -0,0 +1,184 @@
...
+    DWORD _ebp;
+};
+
+#define DISPOSITION_DISMISS 0
+#define DISPOSITION_CONTINUE_SEARCH 1
+#define DISPOSITION_NESTED_EXCEPTION 2
+#define DISPOSITION_COLLIDED_UNWIND 3
+#define TRYLEVEL_NONE -1
+
+extern "C" int _nested_unwind_handler (_EXCEPTION_RECORD *pExceptionRecord, EXCEPTION_REGISTRATION_SEH3 *pRegistrationFrame, _CONTEXT *pContextRecord, EXCEPTION_REGISTRATION **pDispatcherContext)
```
dracc, line 31: nit: The other `.cpp` files in the vcruntime directory seems to use a "more global" `extern "C"` rather than per-function.

dracc:
Tested with the provided snippet, works wonders.
Code readthrough looks good.
:+1:

LukeUsher (approved):

PR #394, opened 2020-08-14
fp20compiler: Add more missing texture modes and process texture mode arguments

Voxel9:
This adds most of the missing texture modes into fp20compiler, except for BRDF and rectangle modes, which can be added at a later date.
Futhermore, some modes which take additional arguments (e.g. `offset_2d(tex, float, float, float, float)`) will now be processed, and the compiler will emit pbkit code to set the appropriate GPU registers as required.

This makes some progress on #208.

### DIFF
--- a/tools/fp20compiler/ts1.0_inst_list.cpp
+++ b/tools/fp20compiler/ts1.0_inst_list.cpp
@@ -84,6 +84,129 @@ void InstList::Invoke()
         case TSP_PASS_THROUGH:
             op = "PASS_THROUGH";
             break;
+        case TSP_OFFSET_2D_SCALE:
+            assert(i >= 1);
+            op = "BUMPENVMAP_LUMINANCE";
+            assert(false); /* Untested */
+            break;
+        case TSP_OFFSET_2D:
+            assert(i >= 1);
+            op = "BUMPENVMAP";
+            break;
+        case TSP_DEPENDENT_AR:
+            assert(i >= 1);
+            op = "DEPENDENT_AR";
+            assert(false); /* Untested */
+            break;
+        case TSP_DEPENDENT_GB:
+            assert(i >= 1);
+            op = "DEPENDENT_GB";
+            assert(false); /* Untested */
+            break;
+
+        case TSP_DOT_PRODUCT_2D_1_OF_2:
+            assert(i == 1 || i == 2);
+            op = "DOT_PRODUCT";
+            break;
+        case TSP_DOT_PRODUCT_2D_2_OF_2:
+            assert(i >= 2);
+            op = "DOT_ST";
+            assert(false); /* Untested */
+            break;
+
+        case TSP_DOT_PRODUCT_DEPTH_REPLACE_1_OF_2:
+            assert(i == 1 || i == 2);
+            op = "DOT_PRODUCT";
+            break;
+        case TSP_DOT_PRODUCT_DEPTH_REPLACE_2_OF_2:
+            assert(i >= 2);
+            op = "DOT_ZW";
+            assert(false); /* Untested */
+            break;
+
+        case TSP_DOT_PRODUCT_3D_1_OF_3:
+            assert(i == 1);
+            op = "DOT_PRODUCT";
+            break;
+        case TSP_DOT_PRODUCT_3D_2_OF_3:
+            assert(i == 2);
+            op = "DOT_PRODUCT";
+            break;
+        case TSP_DOT_PRODUCT_3D_3_OF_3:
+            assert(i == 3);
+            op = "DOT_STR_3D";
+            assert(false); /* Untested */
+            break;
+
+        case TSP_DOT_PRODUCT_CUBE_MAP_1_OF_3:
+            assert(i == 1);
+            op = "DOT_PRODUCT";
+            break;
+        case TSP_DOT_PRODUCT_CUBE_MAP_2_OF_3:
+            assert(i == 2);
+            op = "DOT_PRODUCT";
+            break;
+        case TSP_DOT_PRODUCT_CUBE_MAP_3_OF_3:
+            assert(i == 3);
+            op = "DOT_STR_CUBE";
+            assert(false); /* Untested */
+            break;
+
+        case TSP_DOT_PRODUCT_REFLECT_CUBE_MAP_EYE_FROM_QS_1_OF_3:
+            assert(i == 1);
+            op = "DOT_PRODUCT";
+            break;
+        case TSP_DOT_PRODUCT_REFLECT_CUBE_MAP_EYE_FROM_QS_2_OF_3:
+            assert(i == 2);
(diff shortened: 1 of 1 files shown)


tools/fp20compiler/ts1.0_inst_list.cpp
```diff
@@ -219,6 +219,61 @@ void InstList::Invoke()
     printf(");\n");
     printf("p += 2;\n");
 
+    // Process texture stage mode arguments
```
JayFoxRox, line 222: Consider adding a case for `TSP_CULL_FRAGMENT` with `assert(false); /* Unimplemented */` (and then add the `/* Untested */` to the `assert(false)` in the switch above for consistency)

tools/fp20compiler/ts1.0_inst_list.cpp
```diff
@@ -219,6 +219,61 @@ void InstList::Invoke()
...
+
+            printf("pb_push(p++, NV097_SET_EYE_VECTOR, 4);\n");
+            printf("*p++ = 0x%08x; /* NV097_SET_EYE_VECTOR x */\n", *(unsigned int*)&eye_vector_x);
+            printf("*p++ = 0x%08x; /* NV097_SET_EYE_VECTOR y */\n", *(unsigned int*)&eye_vector_y);
+            printf("*p++ = 0x%08x; /* NV097_SET_EYE_VECTOR z */\n", *(unsigned int*)&eye_vector_z);
+            printf("*p++ = 0x%08x; /* NV097_SET_EYE_VECTOR w */\n", *(unsigned int*)&eye_vector_w);
+            assert(false); /* Untested */
+            break;
+        }
+        }
```
JayFoxRox, line 274: nit: I'd prefer having an empty default-case

tools/fp20compiler/ts1.0_inst_list.cpp
```diff
@@ -219,6 +219,61 @@ void InstList::Invoke()
...
+                list[i].args[3], list[i].args[4]
+            };
+
+            printf("/* NV097_SET_TEXTURE_SET_BUMP_ENV_MAT */\n");
+            printf("pb_push(p++, NV097_SET_TEXTURE_SET_BUMP_ENV_MAT + %d * 64, 4);\n", i);
+            printf("*p++ = 0x%08x; /* NV097_SET_TEXTURE_SET_BUMP_ENV_MAT m[0] */\n", *(unsigned int*)&matrix[0]);
+            printf("*p++ = 0x%08x; /* NV097_SET_TEXTURE_SET_BUMP_ENV_MAT m[1] */\n", *(unsigned int*)&matrix[1]);
+            printf("*p++ = 0x%08x; /* NV097_SET_TEXTURE_SET_BUMP_ENV_MAT m[2] */\n", *(unsigned int*)&matrix[2]);
+            printf("*p++ = 0x%08x; /* NV097_SET_TEXTURE_SET_BUMP_ENV_MAT m[3] */\n", *(unsigned int*)&matrix[3]);
+            break;
```
JayFoxRox, line 255: Did you try scaling horizontally / vertically separately, or rotations, to ensure the proper order of matrix elements?

tools/fp20compiler/ts1.0_inst_list.cpp
```diff
@@ -219,6 +219,61 @@ void InstList::Invoke()
...
+        case TSP_OFFSET_2D_SCALE: {
+            float matrix[2*2] = {
+                list[i].args[1], list[i].args[2],
+                list[i].args[3], list[i].args[4]
+            };
+            float offset = list[i].args[5];
+            float scale = list[i].args[6];
+
+            printf("pb_push(p++, NV097_SET_TEXTURE_SET_BUMP_ENV_MAT + %d * 64, 6);\n", i);
+            printf("*p++ = 0x%08x; /* NV097_SET_TEXTURE_SET_BUMP_ENV_MAT m[0] */\n", *(unsigned int*)&matrix[0]);
```
JayFoxRox, line 234: nit: Consider a `static` helper function like `FloatAsInteger` (or a better name). It should also:
```c
assert(sizeof(float) == 4);
assert(sizeof(unsigned int) == 4);
```

alternatively (this needs more work as we have to change a couple of things), we could potentially rewrite:

```c
float args[TSP_MAX_ARGS];
```
to be:
```c
union Arg {
    float value;
    unsigned int value_raw;
};
Arg args[TSP_MAX_ARGS];
```

We probably still want that `sizeof` assert somewhere (potentially in `main()` as a self-check):
```c
assert(sizeof(Arg.value) == 4);
assert(sizeof(Arg.value_raw) == 4);
```

tools/fp20compiler/ts1.0_inst_list.cpp
```diff
@@ -38,6 +38,14 @@ InstList& InstList::operator+=(InstPtr t)
     return *this;
 }
 
+static unsigned int float_to_uint(float value)
```
JayFoxRox, line 41: nit: to me, it appears that the naming scheme for functions is CamelCase in this project. So it should be `FloatToUint` (or similar).

JayFoxRox, line 41: nit:  I'm not 100% sure about `ToUint` because it could imply something like `(float)1.0f` being turned into `(unsigned int)1`. Hence my initial proposal to something like "Raw".

tools/fp20compiler/ts1.0_inst_list.cpp
Voxel9: Yep. I just tested scaling horizontally and the matrix elements seem to be ordered fine.

tools/fp20compiler/ts1.0_inst_list.cpp
Voxel9: Are the new changes acceptable? (Same for below review comment)

thrimbor (approved):
I looked over the code the best I could and tried the xgu samples mentioned in the Discord. I intend to merge this soon-ish - even if I missed some mistake in the details, it's "better than master".

One minor thing - I'd prefer if PRs like this came with a reference to the test code and specified which of the new code paths the tests are supposed to exercise. I know that the situation with xgu not being part of this repo is a bit special, but the PR description doesn't even mention the samples repo.

tools/fp20compiler/ts1.0_inst_list.cpp
Voxel9: It would break any texture modes that aren't handled by the switch statement, because it just goes through the modes that have arguments for them, if that makes any sense.
So if something like `nop()` or `texture_2d(tex)` was processed, it will fall under the default-case here and hit the false assertion.

thrimbor: Ah, yes, that makes sense.

thrimbor:
It's been five days, so I'm merging now. Thanks!

PR #396, opened 2020-08-18
xboxkrnl: Make Handles parameter of NtWaitForMultipleObjectsEx const

thrimbor:
This fixes a warning about implicitly casting away `const` in `WaitForMultipleObjectsEx`.

From what I could see, `NtWaitForMultipleObjectsEx` doesn't try to modify the handles array (it also wouldn't make any sense), so we should be fine.

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -2626,7 +2626,7 @@ XBAPI NTSTATUS NTAPI NtWaitForSingleObject
 XBAPI NTSTATUS NTAPI NtWaitForMultipleObjectsEx
 (
     IN ULONG Count,
-    IN HANDLE Handles[],
+    IN CONST HANDLE Handles[],
     IN WAIT_TYPE WaitType,
     IN KPROCESSOR_MODE WaitMode,
     IN BOOLEAN Alertable,


JayFoxRox:
I'm not sure about this.
It isn't our API, it's MS API. We should check wether this is exposed from the DDK or what other sources document.

thrimbor:
The WDK doesn't contain a declaration of `NtWaitForMultipleObjectsEx`- Xenia does have one, but it doesn't need to care about const because it only cares about a host-side implementation of it.
`NtWaitForMultipleObjectsEx` isn't really documented anywhere else - it's essentially like `NtWaitForMultipleObjects`, but with a `WaitMode` parameter - so I don't really see an issue with making this parameter const.

JayFoxRox:
Fair enough; we'll consider that `WaitForMultipleObjectsEx` and friends use it internally and that it's probably `const` because that would propagate.

(I was unable to find information about `NtWaitForMultipleObjects` / `ZwWaitForMultipleObjects`?)

> I don't really see an issue with making this parameter const.

It's just about correctness and avoiding warnings elsewhere (function pointers in kernel-hooks for example, which also won't have the `const`).

But given that this is undocumented, it's probably not an issue anyway.
So I can also understand the perception that my comment was bike-shedding.

PR #397, opened 2020-09-03
pdclib: Update submodule to b3207ef

thrimbor:
Updates PDCLib to the current version in `nxdk-pdclib`, which includes the math fixes.


dracc:
Looks good to me. :+1:

PR #398, opened 2020-09-13
samples: Fix extern screen variables

Ryzee119:
b8d162b0aa9f5da5ed7a2af0cd7fb084a394ad89 changed the default values of SCREEN_WIDTH	and SCREEN_HEIGHT to zero. 

Now these are only synchronised to the actual framebuffer size when using the debug API.
Unfortunately this broke the sdl, sdl_tff and sdl_image samples as they relied on using these variables by `extern int`. 

This fix declares them locally in each sample.

Not sure if this is the best solution going forward, but it gets the samples working again.

Issue reported on xboxdev discord by a user.

### DIFF
--- a/samples/sdl/main.c
+++ b/samples/sdl/main.c
@@ -174,8 +174,8 @@ static const struct {
 };
 
 //Screen dimension constants
-const extern int SCREEN_WIDTH;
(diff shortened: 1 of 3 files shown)


dracc:
Looks good to me. 👍🏻

Issue #324, opened 2020-04-25
nxdk creates invalid XBE under some circumstances


JayFoxRox:
I had a weird issue while porting HODE: https://github.com/JayFoxRox/hode/pull/1#discussion_r403417028 ; [related Makefile section](https://github.com/JayFoxRox/hode/blob/ab70deb82aa50b5d2edf18713233db7f8a0a2bfd/Makefile.nxdk#L15):

Whenever I'd link the compiled object files for the XBR scaler, the resulting XBE was not loadable by the Xbox kernel within XQEMU (it boots into dashboard instead of launching into the nxdk binary). I was using LLVM 9 at the time.
I was also getting some linker warnings, but I'm not sure if that is related to loading failure of the XBE.

I suspect this is an issue with cxbe (more likely) or LLD (less likely). It's also possible that this is a bug in extract-xiso (potentially https://github.com/XboxDev/extract-xiso/issues/29), but I doubt that.

Ideally someone would create a minimal test case to reproduce it.

JayFoxRox:
Thanks to @thrimbor who investigated this, we know that the XBE and EXE have a huge `.data` section (64MiB):

```
Section Name                     : 0x00010479 (".data")
Flags                            : 0x00000003 (Writable) (Preload) 
Virtual Address                  : 0x001A0000
Virtual Size                     : 0x04010000
```

This is a "bug" in libxbr that is documented here: https://github.com/Treeki/libxbr-standalone/issues/3.
\- It uses a 24-bit index lookup table with each element at 32-bit = 64MiB lookup table.

So this isn't a bug with nxdk, but a problem with libxbr.

I'll remove the "critical" label, mark this as "invalid" and close this issue.

PR #400, opened 2020-10-01
cxbe: Warn on excessively large image sizes

thrimbor:
This makes cxbe spit out warnings on excessive image sizes, to better diagnose issues such as #324.


Ryzee119:
Tested this by modifying the hello sample and adding large arrays.

Was able to trigger the 32MB warning, and 64Mb warnings as expected.

image

PR #406, opened 2020-10-18
debug: Support setting and resetting cursor position

mikezila:
This makes it easy to use debugPrint() to make a simple persistent readout. Call debugPrint() with your readout text after you have drawn your screen, then debugResetCursor() to prevent the next frames prints from running down the screen on the next frames.

It's not perfect because the readout will flicker a bit depending on how fast the screen is being updated, but it's very easy to use this way and perfectly readable for quick debug output.

### DIFF
--- a/lib/hal/debug.c
+++ b/lib/hal/debug.c
@@ -236,6 +236,25 @@ void debugClearScreen( void )
 	XVideoFlushFB();
 }
 
+void debugResetCursor ( void )
+{
+	nextRow = MARGIN;
+	nextCol = MARGIN;
+}
+
+void debugMoveCursor (int x, int y)
+{
+	if ( x < MARGIN || x > SCREEN_WIDTH-MARGIN-FONT_WIDTH ) {
+		return;
+	}
+	if ( y < MARGIN || y > SCREEN_HEIGHT-MARGIN-FONT_HEIGHT ) {
+		return;
+	}
+	
+	nextRow = y;
+	nextCol = x;
+}
+
 void debugPrintHex(const char *buffer, int length)
 {
 	char tmp[10];
--- a/lib/hal/debug.h
+++ b/lib/hal/debug.h
@@ -29,7 +29,8 @@ void debugPrintBinary( int num );
 void debugPrintHex(const char *buffer, int length);
 void debugClearScreen( void );
 void debugAdvanceScreen( void );
-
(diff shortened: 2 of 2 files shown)


lib/hal/debug.c
```diff
@@ -236,6 +236,11 @@ void debugClearScreen( void )
 	XVideoFlushFB();
 }
 
+void debugResetCursor( void ) {
```
thrimbor, line 239: Formatting doesn't follow the pattern the other functions use

lib/hal/debug.c
```diff
@@ -236,6 +236,11 @@ void debugClearScreen( void )
...
+	nextRow = MARGIN;
+	nextCol = MARGIN;
```
thrimbor, line 241: It would probably be better to just make this a generic "set cursor to x,y" function instead, and have resetting the cursor to 0,0 just be one of the cases handled by it.

thrimbor:
The commit title is very unwieldy and doesn't use the prefix pattern we usually use - I'd suggest something like `debug: Add debugResetCursor() to allow resetting the cursors instead of scrolling`

mikezila:
I agree it would make more sense to allow an arbitrary x and y position. Would it be better it add that arbitrary position to the normal margin that debugPrint() uses, or no?

On one hand, adding it to the default margins makes sense since debugPrint() interacts with them anyways, but on the other, it makes it so that the position you specify isn't the position that you actually get.

If you do not respect the margin then resetting to the default debugPrint() position requires you to specify a magic number.

Perhaps a `debugResetCursor()` and `debugMoveCursor(int x, int y)` so that you can easily set a fully arbitrary position, but also easily wrangle `debugPrint()` without worrying about the position?

mborgerson:
Thank you for this PR, @mikezila! This helps solve an annoying limitation.

Some minor points of feedback before merge: 
* The arbitrary cursor placement does no bounds checking; please make sure that this function does not allow you to write outside the framebuffer.
* Would you please squash the commits to one commit and change the commit message to have a shorter title (~50 chars or so). Suggestion: `debug: Support setting cursor position`
* If you're feeling motivated, would you please refactor surrounding code to use your new functions? This is a stretch goal: I won't block merge for this, we can follow up later if you don't get to it.

mikezila:
I'll squad the commits and implement your suggestions for sure. Should be able to look into this more tomorrow evening.

PR #409, opened 2020-10-25
ci: Build, test and push Docker image

### LINKED ISSUE #37: Add Dockerfile for easy builds
Using containers for builds will help standardize the toolchain and build process across platforms. Let's create a basic Dockerfile that can be used as a baseline ("it works on my machine").

thrimbor:
This adds a Dockerfile plus two actions to build and test a Docker image containing Alpine Linux 3.12, the requirements of nxdk (llvm etc.) and the nxdk source code. Pushes to master will trigger the second action, which builds the image and pushes it to DockerHub: https://hub.docker.com/repository/docker/xboxdev/nxdk. The images are tagged with their git short hash, so we we can correlate commits and images.
I also added a `.dockerignore` to exclude the git directories and some other large unnecessary directories and files from the image to reduce the size.

For the push to DockerHub to work, we'll have to configure the username and token secrets on the nxdk repo before merging.

To test, you can pull the image from https://hub.docker.com/repository/docker/thrimbor/nxdk and run this from a directory containing an nxdk Makefile:
``docker run --rm -v `pwd`:/usr/src/app -it nxdk make``
It'll automatically set `NXDK_DIR` correctly, too.

/edit:
Fixes #37.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+**/.git
+tools/cg/mac
+tools/cg/win
+lib/sdl/SDL2/test
+lib/libcxx/test
--- a/.github/workflows/build_docker_image.yml
+++ b/.github/workflows/build_docker_image.yml
@@ -1,0 +1,47 @@
+name: Build Docker image
+
+on:
+  push:
+    branches: master
+
+jobs:
+  main:
+    name: Docker
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Clone Tree
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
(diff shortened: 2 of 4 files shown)


thrimbor:
I recently learned that Docker Hub will rate-limit its users. To avoid hitting this limit when doing CI runs, I've modified nxdk-buildbase to push its images not only to Docker Hub, but also to the GitHub container registry. I then updated this PR to pull from GHCR and also push its images to both registries.

The secrets for both registries are already set for this repo, so after merging the nxdk image should appear on Docker Hub as well as under the "Packages" tab on the XboxDev org on GitHub.

mborgerson:
Awesome. Thanks for doing this!

PR #411, opened 2020-10-31
winapi: Implement ZeroMemory()

thrimbor:
This adds `ZeroMemory`, which is just a wrapper for `RtlZeroMemory`. I initially used this in a feature branch I'm working on, and even though that branch now doesn't require it anymore, I still think it's worthy of getting added just for completeness.


dracc (approved):

Ryzee119:
All works fine for me

PR #413, opened 2020-11-25
Makefile: Fix lld-link -threads argument

Voxel9:
Argument expects an integer on the latest version of lld-link (11.0.0-1), rather than the string "no".
This fixes compilation of nxdk programs on windows msys2, which would otherwise have shown the following error at link time:
`lld-link: error: -threads:: expected a positive integer, but got 'no'`

Tested and working also with clang/llvm 9, for compatibility. It just shows up as an unknown argument warning anyway.


dracc (approved):

thrimbor:
Have you checked whether the workaround is still necessary? If not, we could drop it altogether.

GXTX:
Tested and doesn't seem to be required any longer, however someone else should verify.

Apfel:
Tried this out using a clean installation of msys2, nxdk and whatnot on the `hello` example and this worked just fine.
It made no difference whether it's set to 1 or any other value or entirely removed.

Voxel9:
Works fine for me without the workaround altogether as well. I've updated this PR, so it should be good to go for a merge.

PR #416, opened 2020-12-21
xboxkrnl: Document string and event functions, fix OBJECT_ATTRIBUTES doc

thrimbor:
The documentation for OBJECT_ATTRIBUTES's ObjectName was wrongfully stating that it's a Unicode string, while it is actually an ANSI string.

Also documented a few more functions that I used in another commit today.


lib/xboxkrnl/xboxkrnl.h
```diff
@@ -2087,11 +2104,19 @@ XBAPI VOID NTAPI RtlGetCallersAddress
     OUT PVOID *CallersCaller
 );
 
+/**
+ * Frees the string buffer allocated by RtlAnsiStringToUnicodeString.
+ * @param AnsiString A pointer to an ANSI string whose buffer was previously allocated by RtlAnsiStringToUnicodeString.
```
mborgerson, line 2109: Update to match param

thrimbor: Oops, fixed.

PR #417, opened 2020-12-21
Clean up stray whitespaces in Makefile, add .editorconfig

thrimbor:
This removes stray whitespaces, and introduces a .editorconfig file to avoid re-introducing them.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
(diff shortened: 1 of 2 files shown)


mborgerson:
I'm not a huge fan of these in-repo editor config files but this appears to be a pretty common config format, so if it helps keep things in order I'm for it. Possibly a .clang-format in the future

PR #419, opened 2020-12-29
Implement lookup of functions export by the xbe, add PE and XBE header structs

thrimbor:
This grew a bit larger than anticipated. It implements the ability to look up functions marked with `__declspec(dllexport)` by calling `GetProcAddress` with NULL as the handle. JFR and I briefly talked about this in #335, but we wrongfully assumed that this would look up the exports of the module calling `GetProcAddress`, while it actually looks up the exports of the main module, i.e. the currently running xbe in this case. I have confirmed that this is the correct behavior by testing on Windows 7 with VS 2017.

8df5cb2 is required to be able to do this at all. To look up exports, we need the `IMAGE_EXPORT_DIRECTORY` header. This header is referenced in the PE optional header, but the xbe format does not have such a reference. The struct gets its own section called `.edata`, but modern versions of MSVC and lld merge this section into `.rdata` by default. This commit overrides this behavior by adding a merge rule that merges `.edata` into a new section called `.edataxb`, so we can look up that section name during runtime and gain access to `IMAGE_EXPORT_DIRECTORY` this way. Unfortunately, this currently makes lld emit a warning - this is a quirk of lld, overriding the default behavior should not result in a warning, nor does it on MS's linker.

1905ebf is there to make dealing with XBE headers during runtime at least somewhat sane. The structs are loosely based on caustik's spec, but I tried to name the fields similar to their PE counterparts, with one notable exception being `SizeOfStack` - caustik's spec names this field `PE Stack Commit` and claims it can be directly copied from the similarly named PE field, which is incorrect (see #86). I'm not fully sure `winnt.h` is the best place for these structs, as they're not usually found there, but the PE headers are there, too, and the XBE headers reuse some of their structs.

A simple demo that show what this can do now can be found here: https://gist.github.com/thrimbor/8d48c81dd2ba6585c6e4f3a98f567e15

### DIFF
--- a/lib/winapi/libloaderapi.c
+++ b/lib/winapi/libloaderapi.c
@@ -6,6 +6,7 @@
 
 #include <assert.h>
 #include <stdlib.h>
+#include <string.h>
 
 HMODULE LoadLibraryExA (LPCSTR lpLibFileName, HANDLE hFile, DWORD dwFlags)
 {
@@ -30,9 +31,46 @@ BOOL FreeLibrary (HMODULE hLibModule)
     return TRUE;
 }
 
+static PIMAGE_EXPORT_DIRECTORY find_edataxb (void)
+{
+    DWORD num_sections = CURRENT_XBE_HEADER->NumberOfSections;
+    PXBE_SECTION_HEADER section_header_addr = CURRENT_XBE_HEADER->PointerToSectionTable;
+
+    for (DWORD i = 0; i < num_sections; i++) {
+        if (memcmp(section_header_addr[i].SectionName, ".edataxb", 8) == 0) {
+            return (PIMAGE_EXPORT_DIRECTORY)section_header_addr[i].VirtualAddress;
+        }
+    }
+
+    return NULL;
+}
+
 FARPROC GetProcAddress (HMODULE hModule, LPCSTR lpProcName)
 {
-    // FIXME: If hModule is NULL, the symbol is looked up in the current module
+    if (hModule == NULL) {
+        // When no dll handle is given, the symbol gets looked up in the main module
+
+        PIMAGE_EXPORT_DIRECTORY exportdir = find_edataxb();
+        if (!exportdir) {
+            SetLastError(ERROR_PROC_NOT_FOUND);
+            return NULL;
+        }
+
+        for (DWORD i = 0; i < exportdir->NumberOfNames; i++) {
+            const char **nametable = (const char **)(exportdir->AddressOfNames + XBE_DEFAULT_BASE);
+            const char *name_addr = (const char *)(nametable[i] + XBE_DEFAULT_BASE);
+
+            if (strcmp(lpProcName, name_addr) == 0) {
+                // Found a matching name and its index. This index is not valid for the address table, that index needs to be looked up in the ordinal table!
+                WORD *ordtable = (WORD *)(exportdir->AddressOfNameOrdinals + XBE_DEFAULT_BASE);
+                BYTE **proctable = (BYTE **)(exportdir->AddressOfFunctions + XBE_DEFAULT_BASE);
+                return (FARPROC)proctable[ordtable[i]] + XBE_DEFAULT_BASE;
+            }
+        }
+
+        SetLastError(ERROR_PROC_NOT_FOUND);
+        return NULL;
+    }
 
     // FIXME: If the module handle is invalid, fail with ERROR_MOD_NOT_FOUND
 
--- a/lib/winapi/winnt.h
+++ b/lib/winapi/winnt.h
@@ -226,4 +226,85 @@ typedef struct _IMAGE_TLS_DIRECTORY32
     };
 } IMAGE_TLS_DIRECTORY_32, *PIMAGE_TLS_DIRECTORY_32;
 
+
+// The XBE_ prefixed functionality is an nxdk-specific extension, and not usually found in winnt.h
+
+typedef struct _XBE_CERTIFICATE_HEADER
+{
+    DWORD SizeOfHeader;
+    DWORD TimeDateStamp;
+    DWORD TitleID;
+    WCHAR TitleName[40];
+    DWORD AlternateTitleIDs[16];
+    DWORD AllowedMedia;
+    DWORD GameRegion;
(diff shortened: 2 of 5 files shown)


lib/winapi/libloaderapi.c
```diff
@@ -30,9 +31,46 @@ BOOL FreeLibrary (HMODULE hLibModule)
...
+    }
+
+    return NULL;
+}
+
 FARPROC GetProcAddress (HMODULE hModule, LPCSTR lpProcName)
 {
-    // FIXME: If hModule is NULL, the symbol is looked up in the current module
+    if (hModule == NULL) {
+        // If hModule is NULL, the symbol is looked up in the main module
```
Teufelchen1, line 51: This comment is odd. It does not proper describe why this is happening but how(which I can read myself). Maybe change it to something like: 
"When the dll handle is not given, retrieve the address of an exported function in the calling module"

thrimbor: `When the dll handle is not given` is logically equivalent to `If hModule is NULL`, but I went ahead and changed it anyway. The second part of the comment is the more important one.

`retrieve the address of an exported function in the calling module` is not correct, the module in which the symbol is looked up is the main module (the xbe, in our case), not the calling module (which could be a dll in the future). It's a method used on Windows to look up exe symbols from a dll that was loaded by the exe.

thrimbor:
I looked over it once more and fixed some formatting. I'll merge once the checks are completed.

lib/winapi/libloaderapi.c
```diff
@@ -30,9 +31,46 @@ BOOL FreeLibrary (HMODULE hLibModule)
     return TRUE;
 }
 
+static PIMAGE_EXPORT_DIRECTORY find_edataxb (void)
+{
+    DWORD num_sections = CURRENT_XBE_HEADER->NumberOfSections;
+    PXBE_SECTION_HEADER section_header_addr = CURRENT_XBE_HEADER->PointerToSectionTable;
+
+    for (DWORD i = 0; i < num_sections; i++) {
+        if (memcmp(section_header_addr[i].SectionName, ".edataxb", 8) == 0) {
```
JayFoxRox, line 40: I consider this name (".edataxb") extremly ugly and hacky, and I hope it will be renamed to something better, before XBE/DXT/DLL with it start shipping.

I don't have ideas for a better name though.

Ideally it should just remain `.edata` in my opinion. Could we have a rule like:
- `-merge:.edata=.edata`
- `-merge:.edata=.edata$`
- `-merge:.edata$=.edata`.

Would those still merge to `.rdata` (or even MSVC / Clang behaviour differences maybe)?

If nothing else works, `.exports` is also a slightly better section name to me.
*(Although in a year or so, I'll probably start complaining that it is not closely related to `.edata`, which makes it harder to find the docs)*

thrimbor:
`-merge:.edata=.edata` works, but only with lld. It crashes the MSVC linker. The other two don't work, they cause the name to be mangled with following linker parameters.
I originally didn't want to use something that was incompatible with MSVC, because LLVM/lld might change their behavior, but I guess if this ends up being a problem, we could submit a patch upstream to introduce an xbox subsystem that will have the [default merge rule](https://github.com/llvm/llvm-project/blob/184377da5c7ce624adf09341360c6cf8f31ebe8f/lld/COFF/Driver.cpp#L1654) disabled, so I submitted #462 to change it.

lib/winapi/libloaderapi.c
thrimbor: `-merge:.edata=.edata` works, but only with lld. It crashes the MSVC linker. The other two don't work, they cause the name to be mangled with following linker parameters.
I originally didn't want to use something that was incompatible with MSVC, because LLVM/lld might change their behavior, but I guess if this ends up being a problem, we could submit a patch upstream to introduce an xbox subsystem that will have the [default merge rule](https://github.com/llvm/llvm-project/blob/184377da5c7ce624adf09341360c6cf8f31ebe8f/lld/COFF/Driver.cpp#L1654) disabled, so I submitted #462 to change it.

lib/winapi/winnt.h
```diff
@@ -13,4 +13,298 @@ typedef LPCWSTR LPCTSTR;
...
+        struct {
+            DWORD Reserved0 : 20;
+            DWORD Alignment : 4;
+            DWORD Reserved1 : 8;
+        };
+    };
+} IMAGE_TLS_DIRECTORY_32, *PIMAGE_TLS_DIRECTORY_32;
+
+
+// The XBE_ prefixed functionality is an nxdk-specific extension, and not usually found in winnt.h
```
GXTX, line 230: Should this perhaps be behind a `ifdef` since we're notating this as being nxdk specific?

lib/winapi/libloaderapi.c
JayFoxRox: >  we could submit a patch upstream to introduce an xbox subsystem that will have the default merge rule disabled

I'd rather avoid such massive changes in a huge project for a small project like nxdk.
Disabling merge rules using a generic switch (not Xbox-specific) could be a good option though.

I wonder if there isn't a better way, like recreating this section as part of Cxbe from the EXE header (probably issues with overlapping sections or alignment when splitting into a section?).

We could also rename from `.edata` to `xbe:.edata` (= make the command part of the name) and then execute the rename in Cxbe (which runs into many section-name length conflicts anyway, so it makes sense to provide a step for it). Might trip debuggers though.

I also had the same issue with xbe-loader and to this day, I believe I hated all solutions that I've tried so far.

lib/winapi/winnt.h
JayFoxRox: No, because nxdk is Xbox specific, so our winnt.h is exclusively meant for original Xbox. Forks of it are extremely unlikely due to how specific our winapi is to the NT kernel (and Xbox and nxdk ecosystem in particular).
However, this XBE stuff might deserve to be in a different header; hard to tell.

If it follows MS (ideally XDK) conventions it's fine here.
If it follows other conventions it should probably be an nxdk lib with nxdk codestyle?

No need to bikeshed over it for now though.
For now it's enough that this is marked as nxdk specific extension.

Issue #381, opened 2020-07-15
Samples don't demonstrate how to unmount drives


kosmas12:
The winapi_filefind and winapi_drivelist samples don't show how to unmount drives, even though it would probably be a good idea to show how, even though it's simple.

kosmas12:
If I recall correctly this has been fixed, can we close this issue or is there something missing?

PR #382, opened 2020-07-16
samples: Make winapi samples unmount drives when exiting

kosmas12:
Fixes #381

PR #421, opened 2021-01-14
samples: Make winapi samples unmount drives before finishing execution

kosmas12:
This PR fixes issue #381. While I had previously opened a PR for this (#382), it was getting bloated, so I am opening this new one after I improved upon all the suggestions in the other PR.


samples/winapi_filefind/main.c
```diff
@@ -27,24 +30,23 @@ int main(void)
...
-            debugPrint("Directory: ");
-        } else {
-            debugPrint("File     : ");
-        }
-
-        debugPrint("%s\n", findFileData.cFileName);
+      if (findFileData.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) {
+        debugPrint("Directory: ");
+      }
+      else {
```
thrimbor, line 41: Weird formatting.

samples/winapi_filefind/main.c
```diff
@@ -10,12 +10,15 @@ int main(void)
 {
     XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
 
+    // Create a variable for WinAPI error checking
+    DWORD error = GetLastError();
```
Teufelchen1, line 14: Is there a specific reason why the variable gets immediately assigned the "LastError" without it being used/checked? Wouldn't it be possible to just write `DWORD error;`? (Maybe I'm missing something, in that case, enlighten me please 🤓)

samples/winapi_filefind/main.c
```diff
@@ -53,9 +55,15 @@ int main(void)
...
-        Sleep(2000);
+    ret = nxUnmountDrive('C');
+    // If there was an error while unmounting
+    if (!ret) {
+      error = GetLastError();
+      debugPrint("Couldn't unmount C: drive! Reason: %x", error);
     }
 
+    // Give some time to the user to read the output
+    Sleep(10000);
```
Teufelchen1, line 66: The old code looped infinite. Why the change to a timeout?

samples/winapi_filefind/main.c
```diff
@@ -53,9 +54,16 @@ int main(void)
...
+    // If there was an error while unmounting
+    if (!ret) {
+      error = GetLastError();
+      debugPrint("Couldn't unmount C: drive! Reason: %x", error);
+    }
+
     while (1) {
         Sleep(2000);
-    }
+    };
```
thrimbor, line 66: Accidental change?

samples/winapi_drivelist/main.c
```diff
@@ -9,25 +9,24 @@ int main(void)
...
     BOOL ret;
+    DWORD error;
     ret = nxMountDrive('C', "\\Device\\Harddisk0\\Partition2\\");
     if (!ret) {
-        debugPrint("Failed to mount C: drive!\n");
-        Sleep(5000);
-        return 1;
+        // Additional error info can be retrieved with GetLastError()
+        error = GetLastError();
+        debugPrint("Failed to mount C: drive! Error code: %x\n", error);
```
thrimbor, line 17: The old code simply quit the program (which makes us unable to properly unmount),  but this code now just continues as if the error didn't happen.

I think this is a case where using `goto` is actually beneficial. I'd recommend to add a `goto end;` or something here, that jumps to a label placed just after unmounting the two drives, and a `goto unmount_c;` in the if statement below, which jumps to a label placed after unmounting E and before unmounting C (note that the unmounting order will have to change for that).

It's probably also a good idea to introduce an `unmount_e` label, too, and jump to that in case `GetLogicalDrives()` or `GetLogicalDriveStringsA()` fail.

This similarly applies to the other sample.

samples/winapi_drivelist/main.c
```diff
@@ -56,9 +53,23 @@ int main(void)
...
         while(*drive++);
     }
-    debugPrint("\ndone");
 
-    while(1) {
+    ret = nxUnmountDrive('C');
+    // If there was an error while unmounting
+    if (!ret) {
+      error = GetLastError();
+      debugPrint("\nCouldn't unmount C: drive! Error code: %x\n. Trying to unmount E: drive...\n", error);
```
thrimbor, line 61: Indentation doesn't match the rest of the file.

samples/winapi_drivelist/main.c
```diff
@@ -56,9 +53,23 @@ int main(void)
...
+    }
+
+    ret = nxUnmountDrive('E');
+    if (!ret) {
+      error = GetLastError();
+      debugPrint("\nCouldn't unmount E: drive! Error code: %x\n", error);
```
thrimbor, line 67: Indentation doesn't match the rest of the file.

samples/winapi_filefind/main.c
```diff
@@ -27,24 +30,22 @@ int main(void)
     // no matter whether they contain a dot or not
     hFind = FindFirstFile("C:\\*.*", &findFileData);
     if (hFind == INVALID_HANDLE_VALUE) {
-        debugPrint("FindFirstHandle() failed!\n");
-        Sleep(5000);
-        return 1;
+      error = GetLastError();
+      debugPrint("FindFirstHandle() failed! Reason: %x\n", error);
```
thrimbor, line 34: Indentation doesn't match the rest of the file.

samples/winapi_filefind/main.c
```diff
@@ -27,24 +30,22 @@ int main(void)
...
-            debugPrint("File     : ");
-        }
-
-        debugPrint("%s\n", findFileData.cFileName);
+      if (findFileData.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) {
+        debugPrint("Directory: ");
+      } else {
+        debugPrint("File     : ");
+      }
+      debugPrint("%s\n", findFileData.cFileName);
```
thrimbor, line 43: Accidental indentation change?

samples/winapi_filefind/main.c
```diff
@@ -53,9 +54,16 @@ int main(void)
 
     FindClose(hFind);
 
+    ret = nxUnmountDrive('C');
+    // If there was an error while unmounting
+    if (!ret) {
+      error = GetLastError();
+      debugPrint("Couldn't unmount C: drive! Reason: %x", error);
```
thrimbor, line 61: Indentation doesn't match the rest of the file.

thrimbor (changes requested):
Left a few more review comments. When you make changes to your PR, I recommend to also leave a comment or respond to the feedback comments, it makes it easier to see that you changed something and which feedback you addressed. When you just quietly push changes, they might go unnoticed.

kosmas12:
Fixed the indentation and added gotos and labels

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,54 @@ int main(void)
...
+    // Retrieve drive bitmasks. Every bit represents one drive letter
     DWORD driveBits = GetLogicalDrives();
     if (driveBits == 0 && GetLastError() != ERROR_SUCCESS) {
-        debugPrint("Failed to retrieve drive bitmask!\n");
-        Sleep(5000);
+        error = GetLastError();
+        debugPrint("Failed to retrieve drive bitmask! Error code: %x\n", error);
+        goto unmount_c;
+        goto unmount_e;
         return 1;
```
thrimbor, line 37: This `return` is dead code and should be removed.

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,54 @@ int main(void)
...
-    // Retrieve drive bitmaks. Every bit represents one drive letter
+    // Retrieve drive bitmasks. Every bit represents one drive letter
     DWORD driveBits = GetLogicalDrives();
     if (driveBits == 0 && GetLastError() != ERROR_SUCCESS) {
-        debugPrint("Failed to retrieve drive bitmask!\n");
-        Sleep(5000);
+        error = GetLastError();
+        debugPrint("Failed to retrieve drive bitmask! Error code: %x\n", error);
+        goto unmount_c;
+        goto unmount_e;
```
thrimbor, line 36: This should just be a single `goto unmount_e;`. The second goto is dead code currently.

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,54 @@ int main(void)
...
     // IMPORTANT: The size passed to GetLogicalDriveStringsA is WITHOUT the null-terminator, even though it gets written
     DWORD charsWritten = GetLogicalDriveStringsA(sizeof(buffer)-1, buffer);
     if (charsWritten == 0) {
-        // Additional error info can be retrieved with GetLastError()
-        debugPrint("Failed to retrieve drive strings!\n");
-        Sleep(5000);
+        error = GetLastError();
+        debugPrint("Failed to retrieve drive strings! Error code: %x\n", error);
+        goto unmount_c;
+        goto unmount_e;
```
thrimbor, line 49: This should just be a single `goto unmount_e;`.

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,54 @@ int main(void)
...
         return 1;
     }
 
     if (charsWritten > sizeof(buffer) - 1) {
         // Can't happen here as our buffer is large enough to cover all possibilities
         debugPrint("Buffer for GetLogicalDriveStringsA too small!\n");
-        Sleep(5000);
+        goto unmount_c;
+        goto unmount_e;
```
thrimbor, line 57: This should just be a single `goto unmount_e;`.

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,54 @@ int main(void)
...
         return 1;
+        Sleep(5000);
```
thrimbor, line 59: This is dead code now and should be removed.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +65,30 @@ int main(void)
         debugPrint("%s\n", drive);
         while(*drive++);
     }
-    debugPrint("\ndone");
 
-    while(1) {
+    goto unmount_c;
+    goto unmount_e;
```
thrimbor, line 70: This should be removed, you're bypassing the `Sleep`-loop here and the second goto is dead code.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +65,30 @@ int main(void)
...
+    debugPrint("\nDone!");
+
+    while (1) {
         Sleep(2000);
     }
 
     return 0;
-}
+
+    unmount_c:
```
thrimbor, line 80: This is the wrong order, `unmount_e` needs to come before `unmount_c, otherwise a failure when mounting E would cause a goto to `unmount_c`, which would unmount C, and subsequently it would try to unmount E, which was never mounted.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +65,30 @@ int main(void)
...
+            error = GetLastError();
+            debugPrint("\nCouldn't unmount C: drive! Error code: %x\n. Trying to unmount E: drive...\n", error);
+        }
+
+    unmount_e:
+        ret = nxUnmountDrive('E');
+        if (!ret) {
+            error = GetLastError();
+            debugPrint("\nCouldn't unmount E: drive! Error code: %x\n", error);
+        }
```
thrimbor, line 93: I find this indentation misleading. Indenting a block of code usually signals a scope level or that the block belongs together, but this is not the case here, it's just normal code at the same scope as the rest of the function.

I'd recommend to use the same indentation for this code as the rest of the function, but use no indentation for the label. Small example:
```c
int main(void) { 
    goto end;

end:
    return 0
}
```

samples/winapi_filefind/main.c
```diff
@@ -53,9 +58,19 @@ int main(void)
 
     FindClose(hFind);
 
+    goto unmount_c;
```
thrimbor, line 61: Should be removed, bypasses the `Sleep`-loop.

samples/winapi_filefind/main.c
```diff
@@ -53,9 +58,19 @@ int main(void)
...
 
     return 0;
+
+    unmount_c:
+        ret = nxUnmountDrive('C');
+        // If there was an error while unmounting
+        if (!ret) {
+            error = GetLastError();
+            debugPrint("Couldn't unmount C: drive! Reason: %x", error);
+        }
```
thrimbor, line 75: Same thing regarding indentation applies here.

thrimbor (changes requested):
You seem to have had a bit of trouble understanding how exactly `goto` works or how my review comment was meant, so let me try to explain a bit further.

`goto` is like a jump instruction, whenever it executes, it simply jumps to the place where the label is located, it's not like function call, there is no returning back to where the goto is.

My idea of handling the errors is something like this (pseudo-code):
```c
    if(mount C fails) {
        print error;
        goto done;
    }

    if(mount E fails) {
        print error;
        goto unmount_c;
    }

    if(something else fails) {
        print error;
        goto unmount_e;
    }

    do whatever the sample is intended to do

unmount_e:
    unmount E here
unmount_c:
    unmount C here

    while(1) {
        Sleep(2000);
    }

   return 0;
```

Notice how C is mounted first, but unmounted last. This is important because a `goto unmount_e;` will first unmount E, and execution will then "fall through" to `unmount_c`, so both drives will get unmounted, but a `goto unmount_c` will only unmount C, not E.
When no error happens, execution will simply continue normally, never executing a goto, but still running the code after the labels, so the drives will get unmounted, too.

kosmas12:
Changes made, hope it's all good now

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +58,25 @@ int main(void)
...
+        debugPrint("\nCouldn't unmount E: drive! Error code: %x\n", error);
     }
 
+unmount_c:
+    ret = nxUnmountDrive('C');
+    if (!ret) {
+        error = GetLastError();
+        debugPrint("\nCouldn't unmount C: drive! Error code: %x\n. Trying to unmount E: drive...\n", error);
+    }
+    Sleep(5000);
```
thrimbor, line 80: There should be a `Sleep()` loop here, it's not clear to me why it was removed.

samples/winapi_filefind/main.c
```diff
@@ -53,9 +57,15 @@ int main(void)
...
+    goto unmount_c;
 
+unmount_c:
+    ret = nxUnmountDrive('C');
+    // If there was an error while unmounting
+    if (!ret) {
+        error = GetLastError();
+        debugPrint("Couldn't unmount C: drive! Reason: %x", error);
+    }
+    Sleep(5000);
```
thrimbor, line 69: There should be a `Sleep()` loop here, it's not clear to me why it was removed.

samples/winapi_filefind/main.c
```diff
@@ -53,9 +57,15 @@ int main(void)
 
     FindClose(hFind);
 
-    while (1) {
-        Sleep(2000);
-    }
+    goto unmount_c;
```
thrimbor, line 60: This line is unnecessary, it has no effect on the actual program behavior.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +58,25 @@ int main(void)
         debugPrint("%s\n", drive);
         while(*drive++);
     }
-    debugPrint("\ndone");
 
-    while(1) {
-        Sleep(2000);
+    debugPrint("\nDone!");
+
+    goto unmount_e;
```
thrimbor, line 64: This line is unnecessary, it has no effect on the actual program behavior.

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,47 @@ int main(void)
...
     BOOL ret;
+    DWORD error;
     ret = nxMountDrive('C', "\\Device\\Harddisk0\\Partition2\\");
     if (!ret) {
-        debugPrint("Failed to mount C: drive!\n");
+        // Additional error info can be retrieved with GetLastError()
+        error = GetLastError();
+        debugPrint("Failed to mount C: drive! Error code: %x\n", error);
         Sleep(5000);
         return 1;
```
thrimbor, line 19: This return value is now different compared to the other handling handling. I suggest to remove these two lines and replace them with a goto to the `Sleep()`-loop (you can name the label `end` or something).

samples/winapi_filefind/main.c
```diff
@@ -10,10 +10,15 @@ int main(void)
...
+
     // Mount C:
     BOOL ret = nxMountDrive('C', "\\Device\\Harddisk0\\Partition2\\");
     if (!ret) {
-        debugPrint("Failed to mount C: drive!\n");
+        // There was an error. We can get more information about an error from WinAPI code using GetLastError()
+        error = GetLastError();
+        debugPrint("Failed to mount C: drive! Reason: %x\n", error);
         Sleep(5000);
         return 1;
```
thrimbor, line 23: This return value is now different compared to the other handling handling. I suggest to remove these two lines and replace them with a goto to the `Sleep()`-loop (you can name the label `end` or something).

thrimbor (changes requested):

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +58,25 @@ int main(void)
...
+    if (!ret) {
+        error = GetLastError();
+        debugPrint("\nCouldn't unmount E: drive! Error code: %x\n", error);
     }
 
+unmount_c:
+    ret = nxUnmountDrive('C');
+    if (!ret) {
+        error = GetLastError();
+        debugPrint("\nCouldn't unmount C: drive! Error code: %x\n. Trying to unmount E: drive...\n", error);
```
JayFoxRox, line 78: Trying to unmount E: has nothing to do with unmount of C: having failed.
Therefore these should at least be separate messages (different `debugPrint` to differentiate them more clearly).

Also, the second message isn't even true, as @thrimbor pointed out, so it should be removed.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +58,25 @@ int main(void)
...
 
-    while(1) {
-        Sleep(2000);
+    debugPrint("\nDone!");
+
+    goto unmount_e;
+
+unmount_e:
+    ret = nxUnmountDrive('E');
+    // If there was an error while unmounting
```
JayFoxRox, line 68: This comment adds nothing in my opinion.
The "GetLastError" in combination with `!ret` strongly implies that this is error-checking.

samples/winapi_filefind/main.c
```diff
@@ -10,10 +10,15 @@ int main(void)
 {
     XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
 
+    // Create a variable for WinAPI error checking
+    DWORD error;
```
JayFoxRox, line 14: This can be a highly local variable, I don't see a benefit for putting it at function scope?
If you declare it where it's used it also won't require any additional comment.

If the variable is redeclared, you should use different names to differentiate them; in this particular sample this only seems to happen after `FindNextFile`, so you could probably do `DWORD findFileError = GetLastError()` in that instance.

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,47 @@ int main(void)
 
     // Mount some drives for demonstration purposes
     BOOL ret;
+    DWORD error;
     ret = nxMountDrive('C', "\\Device\\Harddisk0\\Partition2\\");
     if (!ret) {
-        debugPrint("Failed to mount C: drive!\n");
+        // Additional error info can be retrieved with GetLastError()
+        error = GetLastError();
+        debugPrint("Failed to mount C: drive! Error code: %x\n", error);
```
JayFoxRox, line 17: Here you use:
```
         debugPrint("Failed to mount C: drive! Error code: %x\n", error); 
```

In other instances you use:

```
         debugPrint("Couldn't unmount C: drive! Reason: %x", error); 
```

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,47 @@ int main(void)
...
     }
 
-    // Retrieve drive bitmaks. Every bit represents one drive letter
+    // Retrieve drive bitmasks. Every bit represents one drive letter
     DWORD driveBits = GetLogicalDrives();
     if (driveBits == 0 && GetLastError() != ERROR_SUCCESS) {
-        debugPrint("Failed to retrieve drive bitmask!\n");
-        Sleep(5000);
-        return 1;
+        error = GetLastError();
```
JayFoxRox, line 32: Doing `GetLastError()` twice feels like bad practice that shouldn't appear in samples.

kosmas12:
I just applied all the suggested changes, ready to make more if needed

kosmas12:
Building failed previously, tired me forgot to test before pushing haha.

Errors fixed and building of both samples tested to work fine.

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,46 @@ int main(void)
...
     }
+
     ret = nxMountDrive('E', "\\Device\\Harddisk0\\Partition1\\");
     if (!ret) {
-        debugPrint("Failed to mount E: drive!\n");
-        Sleep(5000);
-        return 1;
+        error = GetLastError();
+        debugPrint("Failed to mount E: drive! Error code: %x\n", error);
+        goto unmount_c;
```
JayFoxRox, line 25: This code makes little sense to me; mounting E failed, so it jumps to unmounting C?
The labels should be named more appropriately (like `handleError` or `end`)

samples/winapi_drivelist/main.c
```diff
@@ -9,45 +9,46 @@ int main(void)
...
+    // Retrieve drive bitmasks. Every bit represents one drive letter
     DWORD driveBits = GetLogicalDrives();
-    if (driveBits == 0 && GetLastError() != ERROR_SUCCESS) {
-        debugPrint("Failed to retrieve drive bitmask!\n");
-        Sleep(5000);
-        return 1;
+    error = GetLastError();
+    if (driveBits == 0 && error != ERROR_SUCCESS) {
+        debugPrint("Failed to retrieve drive bitmask! Error code: %x\n", error);
+        goto unmount_e;
```
JayFoxRox, line 33: Here you jump to `unmount_e`, but in https://github.com/XboxDev/nxdk/pull/421/commits/f9e1ed24b6617932311ba1a90e216bde05eadc25#r611092579 you jump to `unmount_c`

Both of these jump targets are confusing.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +57,25 @@ int main(void)
         debugPrint("%s\n", drive);
         while(*drive++);
     }
-    debugPrint("\ndone");
 
-    while(1) {
-        Sleep(2000);
+    debugPrint("\nDone!");
+
+unmount_e:
```
JayFoxRox, line 63: It's not clear from `unmount_e` that it also does `unmount_c` and then stops execution.
A more appropriate name would be `unloadAfterError` or something

samples/winapi_filefind/main.c
```diff
@@ -38,24 +39,29 @@ int main(void)
...
-
         debugPrint("%s\n", findFileData.cFileName);
     } while (FindNextFile(hFind, &findFileData) != 0);
 
     debugPrint("\n");
 
-    DWORD error = GetLastError();
-    if (error == ERROR_NO_MORE_FILES) {
+    DWORD noMoreFilesError = GetLastError();
+    if (noMoreFilesError == ERROR_NO_MORE_FILES) {
```
JayFoxRox, line 48: This variable name is bad, because it's not always `ERROR_NO_MORE_FILES`, which the name implies.

Instead, `findFileError` is a better name, because it reflects that it's sort-of synchronized with `findFileData`

samples/winapi_filefind/main.c
```diff
@@ -38,24 +39,29 @@ int main(void)
...
+        debugPrint("error: %x\n", noMoreFilesError);
     }
 
     FindClose(hFind);
 
+unmount_c:
+    ret = nxUnmountDrive('C');
+    // If there was an error while unmounting
+    if (!ret) {
+        DWORD unmountError = GetLastError();
```
JayFoxRox, line 60: nit: `DWORD error` is fine, because there's no other variable to confuse this with, the scope is also tiny, so it's very clear what this is referring to.
Giving it an explicit name is just inviting copy/paste errors.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +57,25 @@ int main(void)
...
+
+unmount_c:
+    ret = nxUnmountDrive('C');
+    if (!ret) {
+        error = GetLastError();
+        debugPrint("\nFailed to unmount C: drive! Error code: %x\n", error);
     }
 
-    return 0;
+end:
```
JayFoxRox, line 77: nit: `end` is fine, but I'd suggest `waitIndefinitely` probably: it's most descriptive

samples/winapi_filefind/main.c
JayFoxRox: This wasn't done everywhere

JayFoxRox:
> I just applied all the suggested changes, ready to make more if needed

No, you didn't. And it's very hard to keep track of what you did or how you resolved things.

I suggest to leave a comment on each review comment, telling the reviewer what you did to resolve it.
That way it's much easier to confirm if you understood the feedback + to spot potential logic errors in the new solution.
This is especially important for beginners like you, but should also be expected of experienced coders.

samples/winapi_drivelist/main.c
JayFoxRox: This was not done. In one file you use "Reason" in the other one you use "Error code".
I didn't check if "Couldn't" vs. "Failed to" is also still an issue.

samples/winapi_drivelist/main.c
kosmas12: I believe that this specific label should stay as unmount_c, it's a good enough name IMO.

samples/winapi_drivelist/main.c
kosmas12: `cleanup` would indeed be a better name for that label, using that.

samples/winapi_drivelist/main.c
kosmas12: Will name it `sleepForever`, should be good enough.

samples/winapi_filefind/main.c
kosmas12: Using `findFileError` for this

samples/winapi_drivelist/main.c
kosmas12: I believe that with the change to `cleanup` this will make more sense.

samples/winapi_filefind/main.c
kosmas12: There are many types of errors in the sample, like mountError/unmountError and fileFindError, each with a different scope, I wouldn't be so sure that it's not confusing.

kosmas12:
Changed what I could and tested. If anything more is left I'd also gladly move it to a new PR (even though I know that I should have got it right a looooong time ago and that opening even this 2nd PR is a bit absurd)

kosmas12:
I believe this is ready for a review and hopefully it's (at least one of the) final iteration(s).

samples/winapi_drivelist/main.c
thrimbor: This is still the case.

samples/winapi_drivelist/main.c
thrimbor: @JayFoxRox What would you suggest? Personally, it doesn't bother me, imho it looks just like when a function is freeing a previous allocation before returning due to an error.

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +57,25 @@ int main(void)
...
+    }
+
+unmount_c:
+    ret = nxUnmountDrive('C');
+    if (!ret) {
+        error = GetLastError();
+        debugPrint("\nFailed to unmount C: drive! Error code: %x\n", error);
     }
 
-    return 0;
```
thrimbor, line 65: Why the removal? `main` returns an int, so there should be a return statement, even if it's never reached.

samples/winapi_filefind/main.c
```diff
@@ -38,24 +40,29 @@ int main(void)
...
+        debugPrint("Couldn't unmount C: drive! Error code: %x", unmountError);
+    }
+sleepForever:
     while (1) {
         Sleep(2000);
     }
-
-    return 0;
```
thrimbor, line 60: Why the removal? `main` returns an int, so there should be a return statement, even if it's never reached.

samples/winapi_filefind/main.c
JayFoxRox: If they are in a different scope, then they don't need special name prefixes / suffixes.
We also don't call this `nxdkWinapiFileFindSampleMainUnmountError` (for example) because we just don't have to be that specific - longer names don't help with readability, but make it harder to adapt existing code for different situations.
Names in a narrow scope can (and should) have generic names in my opinion.

samples/winapi_drivelist/main.c
JayFoxRox: I'm not sure; I'd probably cause a fatal error if anything fails to init (to avoid having to do conditional cleanup).

If conditional cleanup is mandatory I'd probably have a boolean which stores which drives were mounted correctly (`mountedC` / `mountedE`), I'd then have a generic cleanup function at the end of the scope which unmounts it (a single label called `cleanup`, which would conditionally deal with cleanup).

In a real application I'd probably have something like `atexit` / destructors which deal with this (= no labels); also ensures proper order of destruction without having to think about it.
Probably not practical for a sample because it adds too much complexity.

kosmas12:
Removal of `return` was indeed stupid of me, added it back now and ready for a new round of reviews if needed

kosmas12:
I believe I have solved the merge conflicts, so this might be in the clear.

lib/usb/libusbohci_xbox/xid_driver.c
```diff
@@ -63,8 +63,16 @@ static xid_dev_t *alloc_xid_device(void) {
...
+
+    //If the xid is at the head of the list, remove now.
+    if (xid_dev == head)
+    {
+        pxid_list = head->next;
+        head = NULL;
+    }
+
+    //Find the device head in the linked list
     while (head != NULL && head->next != xid_dev)
```
thrimbor, line 76: What is this change doing here? (https://github.com/XboxDev/nxdk/commit/f2c634f8774b2de2b958326f5d9e78a27ce827bf)

samples/winapi_drivelist/main.c
```diff
@@ -56,11 +65,30 @@ int main(void)
...
+            debugPrint("\nCouldn't unmount C: drive! Error code: %x\n. Trying to unmount E: drive...\n", error);
+        }
+
+    unmount_e:
+        ret = nxUnmountDrive('E');
+        if (!ret) {
+            error = GetLastError();
+            debugPrint("\nCouldn't unmount E: drive! Error code: %x\n", error);
+        }
+}
```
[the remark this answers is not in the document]
thrimbor: I'm almost certain this was fixed before, but now it's removed again.

samples/winapi_filefind/main.c
```diff
@@ -38,24 +42,31 @@ int main(void)
...
+        debugPrint("Failed to unmount C: drive! Error code: %x", error);
+    }
+sleepForever:
     while (1) {
         Sleep(2000);
     }
 
     return 0;
-}
+}
```
thrimbor, line 72: This newline should not get removed.

thrimbor (changes requested):
There are some review comments still unaddressed, and a couple of issues that seem to be new - newlines are getting removed again, the commit author is incorrect, and somehow https://github.com/XboxDev/nxdk/commit/f2c634f8774b2de2b958326f5d9e78a27ce827bf got mixed in.

This is also branching off a rather old version of nxdk now and imho should be rebased onto current master to make sure there aren't any conflicts.

samples/winapi_drivelist/main.c
JayFoxRox: Especially with everything else jumping to `cleanup` now, I think this should use booleans / keep track of it and handle everything in `cleanup`

PR #422, opened 2021-01-14
samples: Change titles

### LINKED ISSUE #372: Improve `XBE_TITLE` of nxdk samples
I think we should consider doing `XBE_TITLE = "nxdk sample: gamma-fade"` instead of `XBE_TITLE = gamma-fade`.
If this doesn't work (spaces in title), then we should fix it so that it does work.

Adding quotes should help to clarify that leading whitespaces are not part of the name (are they?). It also shows how to include whitespaces in the XBE title.

kosmas12:
On a side note, as I mentioned in the related issue, XBE_TITLE requires just quotation marks to use spaces and special characters, but GEN_XISO requires quotation marks *and* the \ escape character, thus I wasn't able to make GEN_XISO use the XBE_TITLE. Theoretically, it should have been possible to use \ in XBE_TITLE, but, when used, a simple hex view of the xbe shows that \ is also written to the XBE's title, and doesn't actually escape anything.

The samples have also been tested to load and run fine with this fix, so it should cause no issues.

If anyone has a way to fix this or a better way to do the naming, I'll gladly accept it.

Closes #372.

### DIFF
--- a/samples/gamma-fade/Makefile
+++ b/samples/gamma-fade/Makefile
@@ -1,4 +1,4 @@
-XBE_TITLE = gamma-fade
+XBE_TITLE = nxdk\ sample\ -\ gamma-fade
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.c
 NXDK_DIR = $(CURDIR)/../..
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,7 @@ $(OUTPUT_DIR):
 ifneq ($(GEN_XISO),)
 $(GEN_XISO): $(OUTPUT_DIR)/default.xbe $(EXTRACT_XISO)
 	@echo "[ XISO     ] $@"
-	$(VE) $(EXTRACT_XISO) -c $(OUTPUT_DIR) $(XISO_FLAGS) $@ $(QUIET)
+	$(VE) $(EXTRACT_XISO) -c $(OUTPUT_DIR) $(XISO_FLAGS) "$@" $(QUIET)
 endif
 
 $(SRCS): $(SHADER_OBJS)
--- a/samples/gamma/Makefile
+++ b/samples/gamma/Makefile
@@ -1,4 +1,4 @@
-XBE_TITLE = gamma
+XBE_TITLE = nxdk\ sample\ -\ gamma
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.c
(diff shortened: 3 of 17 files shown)


Apfel:
Are you really sure on using colons in file names? Windows' Explorer cries if I try that myself.

kosmas12:
@Apfel This has been fixed now by using dashes instead.

Should have no problem on any platform.

Apfel:
There's no need to escape a dash; it even breaks on Windows.

kosmas12:
It should work now.

Checks have passed.

samples/gamma-fade/Makefile
```diff
@@ -1,5 +1,5 @@
-XBE_TITLE = gamma-fade
-GEN_XISO = $(XBE_TITLE).iso
+XBE_TITLE = "nxdk sample - gamma-fade"
```
thrimbor, line 1: While the quotes are necessary when passing the parameter through the shell to extract-xiso, this is not a good place to add them. People might want to use the `XBE_TITLE` variable for additional stuff, and not realize that the quotation marks are actually part of the string here (they're not in shell scripts, but they are in Makefiles).

Use escaping instead: `nxdk\ sample\ -\ gamma-fade`.

samples/gamma-fade/Makefile
```diff
@@ -1,5 +1,5 @@
...
+GEN_XISO = "nxdk\ sample\ -\ gamma-fade.iso"
```
thrimbor, line 2: This is redundant, and should be `GEN_XISO = $(XBE_TITLE).iso` instead.

thrimbor (changes requested):
The review comments above apply to all the changed Makefiles ofc.

The title changes are fine, but with your quotation mark workaround you missed the actual core issue, the xiso rule in the Makefile:

https://github.com/XboxDev/nxdk/blob/535e1c25d6f4ae033e5820e15d595789d45d5897/Makefile#L126

This needs to be changed to `$(VE) $(EXTRACT_XISO) -c $(OUTPUT_DIR) $(XISO_FLAGS) "$@" $(QUIET)` (note the quotation marks).

That way, the name will be properly quoted when passing the parameter, without the quotes ending up in the variable itself.

kosmas12:
I added `"$@"` in the nxdk Makefile for XISO gen and added escaping in all the XBE titles

thrimbor:
I've taken the liberty to slightly improve your commit title and fix the merge conflict. Merging now.

PR #423, opened 2021-01-14
ci: Explicitly install llvm package to get llvm-lib

thrimbor:
GitHub is currently rolling out a change that makes ubuntu-latest equivalent to ubuntu-20.04 instead of 18.04. On Ubuntu 20.04, the llvm package doesn't seem to be installed out of the box, so for users already affected by the change, Ubuntu CI currently fails.

This commit fixes that by explicitly installing llvm.


dracc:
LGTM :+1:

PR #424, opened 2021-01-18
video: Free framebufer of previous running app

### LINKED ISSUE #357: `AvGetSavedDataAddress` / `AvSetSavedDataAddress` is never free'd or used
`AvGetSavedDataAddress` returns the address of a persistent allocation which contains the framebuffer contents of the previous application (as a D3D surface). At 1080i, this buffer would be about 8MiB. That memory will be unavailable to the application.
This is mostly a documentation issue, because it should probably be the applications responsibility (or that of some API) to read it out / release it.

As a quick workaround, we could potentially integrate this with XVideo by freeing the buffer during `XVideoSetMode`?

kosmas12:
This has been tested to cause no problems on xemu and a 1.1 Xbox. The project I used to test with is [https://github.com/Ryzee119/SDLPoPX](url).

Thanks to @kl0wn for looking into how this can be achieved!

Closes #357

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -296,6 +296,11 @@ void XVideoSetFB(unsigned char *fb)
 {
 	assert(((unsigned int)fb & ~0x7FFFFFFF) == 0);
 	_fb = fb;
+
+	SIZE_T fbSize = MmQueryAllocationSize(_fb);
+	MmPersistContiguousMemory(_fb, fbSize, TRUE);
+	AvSetSavedDataAddress(_fb);
+
 	VIDEOREG(PCRTC_START) = (unsigned int)_fb & 0x7FFFFFFF;
 }
 
@@ -337,6 +342,7 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 	XVideoSetVideoEnable(FALSE);
 
 	if (framebufferMemory != NULL) {
+		MmPersistContiguousMemory(framebufferMemory, screenSize, FALSE);
 		MmFreeContiguousMemory(framebufferMemory);
 	}
 	framebufferMemory = MmAllocateContiguousMemoryEx(screenSize,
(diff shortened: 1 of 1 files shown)


thrimbor (changes requested):
The indentation is different in comparison to the surrounding code (tabs vs. spaces).

I don't think `XVideoSetMode` is the best place for this - imho it makes more sense to put it into `XVideoInit`, so the framebuffer doesn't get freed when failing to set a new mode. There, we could also compare the size of the old framebuffer to the framebuffer size of the requested mode, and reuse the old framebuffer instead of allocating a new one.

To fully resolve the linked issue, we should also mark newly allocated framebuffers as persistent, and store their address with `AvSetSavedDataAddress`, so other applications can reuse our framebuffer.

kosmas12:
I applied all the suggested improvements and tested them on both xemu and a 1.1 Xbox with [https://github.com/dracc/NevolutionX](url) and [https://github.com/Ryzee119/SDLPoPX](url), all seems to work fine.

thrimbor (approved):

thrimbor:
I took the liberty of fixing some smaller remaining issues myself (whitespaces, formatting, commit title too long), merging now.

lib/hal/video.c
```diff
@@ -357,7 +363,28 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 	/* Store the framebuffer that we set; normally this is done in XVideoSetFB.
 	   However, we're using the kernel's AvSetDisplayMode here instead, so we
 	   set it here, too. */
-	_fb = framebufferMemory;
+
+	// Get previous framebuffer
+	PVOID previousFB = AvGetSavedDataAddress();
+
+	if (previousFB != 0) {
```
GXTX, line 370: Shouldn't this be `NULL`?

lib/hal/video.c
```diff
@@ -357,7 +363,28 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
...
+		SIZE_T previousFBSize = MmQueryAllocationSize(previousFB);
+		// If the previous framebuffer is the same size as the one we want to create, don't create a new one and use that instead
+		if (previousFBSize == screenSize) {
+			memset(previousFB, 0x00, screenSize);
```
GXTX, line 374: Shouldn't this use `RtlZeroMemory` / `ZeroMemory()`?

lib/hal/video.c
```diff
@@ -357,7 +363,28 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
...
+			_fb = (unsigned char *)previousFB;
+		} else {
+			MmPersistContiguousMemory(previousFB, previousFBSize, FALSE);
+			MmFreeContiguousMemory(previousFB);
+			AvSetSavedDataAddress(NULL);
```
GXTX, line 379: Is this required? We set a new address once we're out of the if/else block.

Issue #321, opened 2020-04-14
OpenAL-soft port


JayFoxRox:
Independent of #20 it would be good to have a OpenAL software implementation for portability. OpenAL-soft is the obvious choice.

Having OpenAL as API would allow us to recommend it, and then later add a hardware backend for performance.

---

I have prototyped OpenAL-soft ports, but have various issues:

- [OpenAL-soft 1.20](https://github.com/JayFoxRox/nxdk/pull/71) (C++, I quickly gave up on this)
- [OpenAL-soft 1.19](https://github.com/JayFoxRox/nxdk/pull/72) (C, I also gave up on this due to the amount of required underlying features)

I've also been looking at OpenAL-soft 1.1 (to minimize dependencies) and OpenAL-soft 1.15 (compromise in terms of features and dependencies). However, OpenAL-soft 1.19 was the first version to have an SDL2 audio backend; so we'd have to backport a backend, or implement something else.

---

Some missing features to compile 1.19 (after replacing many functions with macros, stubbing some functions, and only from the first couple of error messages):

Event related and threading related:

- GetExitCodeThread
- WaitForSingleObject
- CreateEventW
- SetEvent
- WaitForMultipleObjects
- ResetEvent

Time related:

- GetSystemTimeAsFileTime

Debug related:

- _wgetenv
- DebugBreak

Library related:

- LoadLibraryW
- GetProcAddress
- FreeLibrary

Exception related:

- _except_handler3
- EXCEPTION_CONTINUE_EXECUTION

Widestring conversion related:

- WideCharToMultiByte
- MultiByteToWideChar

Windows basic types:

- HINSTANCE
- HMODULE
- APIENTRY

Potentially avoidable constants (until we implement a dynamic linker for DLLs):

- DLL_PROCESS_ATTACH
- DLL_THREAD_ATTACH
- DLL_THREAD_DETACH
- DLL_PROCESS_DETACH

Atomic operations:

- InterlockedExchange64
- InterlockedExchangePointer
- InterlockedCompareExchangePointer

Much of the required functionality can be stubbed or easily implemented. However, it should be obvious that porting a modern variant of OpenAL-soft is a more massive undertaking than originally anticipated.

With older versions of OpenAL we'll likely be able to get rid of most of these requirements (at the loss of functionality and additional work for adding a new backend).

PR #425, opened 2021-01-26
winapi: Implement more interlocked functions

thrimbor:
Implements `InterlockedExchangePointer`, `InterlockedCompareExchangePointer` (both as xboxkrnl wrappers) and `InterlockedExchange64`, all of which are requirements of #321. Sample code for testing the latter is below.

<details>
  <summary>main.c</summary>

  ```c
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>

int main(void)
{

    LONG64 target = 0xDEADBEEFBEEFC0DE;
    LONG64 value =  0x0123456789123456;
    LONG64 result = InterlockedExchange64(&target, value);

    debugPrint("target: 0x%08X%08X\n", ((int)(target >> 32)), ((int)(target & 0xFFFFFFFF)));
    debugPrint("result: 0x%08X%08X\n", ((int)(result >> 32)), ((int)(result & 0xFFFFFFFF)));
    debugPrint("value: 0x%08X%08X\n", ((int)(value >> 32)), ((int)(value & 0xFFFFFFFF)));

    // Expected output:
    // target: 0x0123456789123456
    // result: 0xDEADBEEFBEEFC0DE
    // value: 0x0123456789123456


    while(1) {
        Sleep(2000);
    }

    return 0;
}

  ```
  
</details>

### DIFF
--- a/lib/winapi/winnt.c
+++ b/lib/winapi/winnt.c
@@ -1,0 +1,29 @@
+#include <winnt.h>
+#include <xboxkrnl/xboxkrnl.h>
+
+LONG64 InterlockedExchange64 (LONG64 volatile *Target, LONG64 Value)
+{
+    // There's no xchg8b in x86, so we implement swapping atomically with cmpxchg8b
+    LARGE_INTEGER old_target_value, new_value;
+    old_target_value.QuadPart = *Target;
+    new_value.QuadPart = Value;
+
+    __asm__ __volatile__("1:\n"
+                         "cmpxchg8b (%2)\n"
+                         "jnz 1b\n" // If the comparison failed, try again with updated edx:eax
+                         : "+a"(old_target_value.LowPart), "+d"(old_target_value.HighPart)
+                         : "r"(Target), "b"(new_value.LowPart), "c"(new_value.HighPart)
+                         : "memory");
+
+    return old_target_value.QuadPart;
+}
+
+PVOID InterlockedExchangePointer (PVOID volatile *Target, PVOID Value)
+{
+    return (PVOID)InterlockedExchange((PLONG)Target, (LONG)Value);
+}
+
+PVOID InterlockedCompareExchangePointer (PVOID volatile *Destination, PVOID Exchange, PVOID Comperand)
+{
+    return (PVOID)InterlockedCompareExchange((PLONG)Destination, (LONG)Exchange, (LONG)Comperand);
+}
--- a/lib/winapi/Makefile
+++ b/lib/winapi/Makefile
@@ -17,7 +17,8 @@ WINAPI_SRCS := \
 	$(NXDK_DIR)/lib/winapi/sync.c \
 	$(NXDK_DIR)/lib/winapi/sysinfo.c \
 	$(NXDK_DIR)/lib/winapi/thread.c \
-	$(NXDK_DIR)/lib/winapi/tls.c
+	$(NXDK_DIR)/lib/winapi/tls.c \
+	$(NXDK_DIR)/lib/winapi/winnt.c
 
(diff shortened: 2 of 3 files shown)


lib/winapi/winnt.c
```diff
@@ -0,0 +1,26 @@
...
+#include <xboxkrnl/xboxkrnl.h>
+
+LONG64 InterlockedExchange64 (LONG64 volatile *Target, LONG64 Value)
+{
+    // There's no xchg8b in x86, so we implement it with cmpxchg8b
+    LARGE_INTEGER li;
+    li.QuadPart = *Target;
+    __asm__ __volatile__("1:\n"
+                         "cmpxchg8b (%2)\n"
+                         "jnz 1b\n" // If the comparison failed, try again with updated edx:eax
```
Teufelchen1, line 11: The comment make it look like "there can go something wrong". But from my understanding, it has to fail the first round by design. I would like to see a different/extended documentation of the overall asm code. It took me way to long to figure out whats going on here. (And now: Hoping that I didn't make myself a fool for this 😅)

thrimbor: Nah it does usually work first try. But it may fail if the thread gets interrupted between the read from `Target`, from which edx and eax get initialized when entering the asm block, and `cmpxchg8b`, and the other thread modifies the target.

Teufelchen1: To wrap my head around it:
- eax:edx will contain the Target-value(as in *not* the address)
- ebx:ecx will contain the Value that should be placed in the Target soon
- r contains the address of Target

Your code does:
1. load up the register as stated above
2. `cmpxchg8b` compares the value at the address of r(the value of Target) with the content of eax:eax
3. If equal: copy ebx:ecx into the memory pointed to by r(setting new value of Target)
4. If not equal: jump back to step 2.

So I would agree now, that it won't fail most of the time. But, I don't get the point of this function in that case. 😟

lib/winapi/winnt.c
```diff
@@ -0,0 +1,26 @@
...
+                         : "+a"(li.LowPart), "+d"(li.HighPart)
+                         : "r"(Target), "b"((int)(Value & 0xFFFFFFFF)), "c"((int)(Value >> 32))
```
Teufelchen1, line 13: Why are you using `LARGE_INTEGER` for `Target` in `eax:edx` but not for `Value`?

lib/winapi/winnt.c
thrimbor: > But, I don't get the point of this function in that case.

Wdym? It's a winapi function to atomically exchange two 8-byte values.

Teufelchen1: [after clarifying via discord]
Instead of my original question, to further document the asm instruction, I would propose to add your thoughts to your first comment in Line 6. 
- Want to swap bytes atomic
- No instruction available 
- could do it via dis- & reenable interrupts
- opted for "abusing" `cmpxchg8b` due to guessing its faster(and simpler)

Anyway, I think thats up to you. Keep in mind that I can't resolve this, as it is just a comment.

lib/winapi/winnt.c
thrimbor: No particular reason. I may end up changing `Value`.

thrimbor: I modified the code to use `LARGE_INTEGER` for the value, too, and slightly tweaked my comments.

Teufelchen1:
☺️ No complaints from my side anymore.

thrimbor:
Nobody merged or added another review, so I just quickly looked over the code again and am merging it now.

PR #431, opened 2021-03-11
debug: Ensure unsigned char when printing hex bytes

Ryzee119:
Currently when printing a hex value using `debugPrintHex` with a sign bit, it would print the 32bit two's complement hex version

i.e printing `0x80` would print `0xffffff80`

Before:
image

After:
image


thrimbor:
Looks good, merging. Thanks!

PR #433, opened 2021-03-16
libc: Add Win32 byteswapping functions

thrimbor:
This implements `_byteswap_ushort`, `_byteswap_ulong` and `_byteswap_uint64`, as they're required by zstd. #418 also implemented them, but this PR implements them as normal functions, which is the way MS does it in their crt.

Also, as laid out in the Discord, I plan to create a new repo for library ports, and put zstd there instead.


dracc:
LGTM 👍

PR #435, opened 2021-03-17
build: Remove NXDK_STACKSIZE variable

thrimbor:
This removes the `NXDK_STACKSIZE` variable in the Makefile. Overriding the nxdk default stack size is still possible by simply appending another `-stack:` parameter to the linker flags.


Ryzee119:
Simple change, nothing broke in my testing. 
Looks good to me!

Issue #172, opened 2019-09-09
Debug library discussion


JayFoxRox:
The current debug library has various issues:

- Only allows sending messages to the framebuffer.
- It doesn't support many display modes and needs to guess the internal framebuffer state (see #149).
- Considers `XVideoGetFB` the framebuffer; this is bad. See #169 and #170.
- It is called from various parts of nxdk (audio driver, network driver, pbkit, legacy-fileio) and can't be disabled; given the `XVideoGetFB` issues, this causes random side-effects and random writes to memory that the user doesn't expect.
- It has dead code, such as constants for colored output. See #171.
- Existing code goes out of its way to support debug.h, like pbkit having explicit functions to switch framebuffers between GPU (self-allocated) and debug.h (`XVideoGetFB`); even if done properly (2 allocated framebuffers) that just means a higher memory consumption and complexity.
- It's not a debug library, but a more generic "console-output" library.
- It ships 2 fonts (terminal + sasteroids2), but only 1 of them is used (terminal).
- The fonts can be hard to read on small / bad monitors.
- The paging sometimes hides valuable information and can be confusing.

All of this makes debug.h hacky and error prone.

---

Ideally we'll replace it with a dedicated debugging lib: debug/debug.h (or similar).

This new debug lib would provide tools for the user to install a local debugger (as supported by the kernel); potentially:

- Crash handlers (SIGSEGV handling).
- Debug input / output handler ( `DbgPrompt` / `DbgPrint`).
- Memory status watchdog (Kernel hooks for guarding allocations or warning about low-memory condition).
- Explicit API for debug input / output, in case the kernel debugger is not used.

Debugger output could be forwarded to callbacks, another kernel debugger, explicit serial connection, a file, or a character buffer.

If the new API also has a framebuffer implementation, an explicit function to redraw the framebuffer would probably be used instead (over implicitly updating); in a callback based API, the user could still decide to explicitly redraw on each message.

However, I'd argue that we don't even need a framebuffer implementation. If it's easy enough to retrieve information from the debug API, then the debug log could be rendered using pbkit, SDL or other code, from within the application.
\- There could still be an independent "console-application" lib which replaces the existing framebuffer-based debug.h

JayFoxRox:
For exposure and because it's slightly related: I found this catch handler for `OutputDebugString` on thrimbors repository: https://github.com/thrimbor/nxdk/commit/098e2fbf45d07e4eae238f1e37f876c69e050def

PR #438, opened 2021-03-24
USB: Replace OHCI USB host library

Ryzee119:
This PR replaces the legacy 'hacked' together linux usb stack currently in nxdk with a new library.

~~Note: CI will fail on the sdl samples because I havent update the `hal/input.c` code yet~~ It should work now, but I have submoduled in own SDL2 fork. https://github.com/XboxDev/nxdk-sdl/pull/38 will need to be merged and SDL2 submodule update prior to merge.

This library was developed by OpenNuvoton (https://github.com/OpenNuvoton) who are a large manufacturer of ARM based embedded microcontrollers. This library was intended for use on their range of chips with OHCI hardware. I specifically imported it from https://github.com/OpenNuvoton/N9H30_emWin_NonOS/tree/master/Library/UsbHostLib with some modifications to make it xbox friendly. The largest change is the separation between virtual and physical memory which is not really a concept on microcontrollers.

The library has some advantages over the current implementation:
1. A more permissive license (Apache 2.0)
2. A full range of inbuilt USB class drivers. I have tested a range of these on my personal repo https://github.com/Ryzee119/nxdk_ohci_test. These samples are just quick hacks for testing. Not considered 'production ready'.
3. Proper OHCI interrupt management so that USB hot-plugging should work properly.
4. The XID driver can properly distinguish between various Xinput devices (Duke, Steel Battalion, Xremote)
5. Proper interrupt out writes, so rumble should work with no problem.
6. Some very useful features for developers is support for CDC class USB-Serial adaptors for improved 'printf debugging'.

**I have thoroughly tested this on my own repo projects. Some outstanding aspects: **

1. ~~Untested on 1.0 xboxes with daughterboard. I would expect this to have no issues though.~~ Has been tested ok!
2. ~~PR does not include changes to `hal/input.c`, consequently sdl2 joystick/gamecontroller will break. I am happy to do this work but would like some feedback on the preferred APIs going forward.~~ sdl_joystick driver rewritten. All old features plus more. (Rumble, hotplugging etc)
3. ~~Currently the ohci library is included as a [submodule to my personal repo](https://github.com/Ryzee119/ohci_usbhostlib). This should either be migrated here or commited directly into nxdk (or it can stay linked to my repo, I dont mind)~~ Its here https://github.com/XboxDev/libusbohci.
4. ~~The mass storage class driver has [been tested](https://github.com/Ryzee119/nxdk_ohci_test/tree/xbox), but disabled in this PR due to the fact it is hardcoded to use [FATFS](http://elm-chan.org/fsw/ff/00index_e.html). I think this requirement should be removed to allow other filesystem types (FATX etc). I can do this in a follow up PR.~~ FATFS code removed. This class driver just supports standard SCSI needed for MSC. User filesystem drivers need to be linked in.


dracc:
Amazing! :tada:
I'll test this tonight.

samples/usb/main.c
```diff
@@ -0,0 +1,38 @@
...
+    usbh_install_conn_callback(&device_connection_callback, &device_disconnect_callback);
+    debugPrint("== ochi usb test ==\n");
+    debugPrint("== Plug and unplug USB devices to test ==\n");
+    while (1) {
+        usbh_pooling_hubs();
+        Sleep(1);
+    }
+    usbh_core_deinit();
+    return 0;
+}
```
Teufelchen1, line 38: Misses newline at end of file

Teufelchen1 (changes requested):

Teufelchen1:
Misses adjustment of the README.md

samples/usb/main.c
```diff
@@ -0,0 +1,38 @@
...
+{
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+
+    usbh_core_init();
+    usbh_install_conn_callback(&device_connection_callback, &device_disconnect_callback);
+    debugPrint("== ochi usb test ==\n");
+    debugPrint("== Plug and unplug USB devices to test ==\n");
+    while (1) {
+        usbh_pooling_hubs();
+        Sleep(1);
```
Teufelchen1, line 34: nit: unusual short sleeping period. Other samples either wait for vblank or wait i.e. 1000ms.

Ryzee119: Although the low level events happen via interrupt, the callbacks and overall library updates occur when usbh_pooling_hubs is called so this should be called reasonably often in this sample to illustrate quick updates when hot plugging.
The 1ms is just based on USB frame size.

I'd rather remove it all together which I can do if it causes some confusion. Thanks for your review!

lib/usb/Makefile
```diff
@@ -1,32 +1,59 @@
...
+NXDK_CFLAGS += -I$(USBH_DIR)/inc
+NXDK_CXXFLAGS += -I$(USBH_DIR)/inc
+
+main.exe: $(NXDK_DIR)/lib/libnxdk_usb.lib
+
+CLEANRULES += clean-usb
+
+.PHONY: clean-usb
+clean-usb:
+	$(VE)rm -f $(USB_OBJS) $(NXDK_DIR)/lib/libnxdk_usb.lib
```
Teufelchen1, line 59: Misses newline 😚

Teufelchen1 (changes requested):

Ryzee119:
Ive made a sdl_joystick driver using this new stack. It does not use hal/input.c at all and instead directly calls the usb HID APIs

I've tested this on SDLPopX, OpenTyrian, Commander keen, Xenium-Tools, sdl_gamecontroller sample and NevolutionX
https://github.com/Ryzee119/nxdk-sdl/tree/usbh

It also supports native Xbox360, Xbone controllers too!

samples/sdl_gamecontroller/main.c
```diff
@@ -97,4 +121,4 @@ int main(void)
   }
   
   return 0;
-}
+}
```
Teufelchen1, line 124: Misses Newline 😏

samples/usb/main.c
```diff
@@ -0,0 +1,37 @@
...
+#include <windows.h>
+#include <usbh_lib.h>
+
+/*
+ * This demo will initialise the usb stack, detect events on the usb ports
+ * and print the device PID and VID.
+*/
+void device_connection_callback(UDEV_T *udev, int status)
+{
+    debugPrint("Device connected on port %u (PID: %04x VID: %04x)\n", udev->port_num,
```
JayFoxRox, line 12: What do these ports look like? If they are USB ports and not gamepad ports, then that's relevant for the message, in my opinion.

samples/usb/main.c
```diff
@@ -0,0 +1,37 @@
...
+                                                                          udev->descriptor.idVendor);
+}
+
+int main(void)
+{
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+
+    usbh_core_init();
+    usbh_install_conn_callback(&device_connection_callback, &device_disconnect_callback);
+    debugPrint("Plug and unplug USB devices to test\n");
```
JayFoxRox, line 30: This should be a bit more explicit.
People might assume that this will test their gamepad like that XDK sample that everybody loves (with the 3D gamepad).
This, however, merely seems to show plug and unplug events?

samples/usb/main.c
```diff
@@ -0,0 +1,37 @@
...
+void device_connection_callback(UDEV_T *udev, int status)
+{
+    debugPrint("Device connected on port %u (PID: %04x VID: %04x)\n", udev->port_num,
+                                                                      udev->descriptor.idProduct,
+                                                                      udev->descriptor.idVendor);
+}
+
+void device_disconnect_callback(UDEV_T *udev, int status)
+{
+     debugPrint("Device disconnected on port %u (PID: %04x VID: %04x)\n", udev->port_num,
```
JayFoxRox, line 19: nit: I think I might prefer order 1. VID, 2. PID; but I'm not sure what other common USB tools use. (Idea is to visually group them by manufacturer first.. so that should come first)

samples/usb/Makefile
```diff
@@ -0,0 +1,5 @@
+XBE_TITLE = nxdk\ sample\ -\ usb
```
JayFoxRox, line 1: nit: I think just calling this sample "usb" is not ideal; but I have trouble coming up with my own name.

*(I'd assume there'll be different USB samples eventually, at least some of which will likely deserve a usb prefix)*

kosmas12: Perhaps call it `hotplugging`?

lib/usb/Makefile
```diff
@@ -1,32 +1,59 @@
...
+NXDK_CFLAGS += -I$(USBH_DIR)/inc
+NXDK_CXXFLAGS += -I$(USBH_DIR)/inc
+
+main.exe: $(NXDK_DIR)/lib/libnxdk_usb.lib
+
+CLEANRULES += clean-usb
+
+.PHONY: clean-usb
+clean-usb:
+	$(VE)rm -f $(USBH_OBJS) $(NXDK_DIR)/lib/libnxdk_usb.lib
```
JayFoxRox, line 59: I think this Makefile should be part of the USB submodule, unless that submodule remains platform unspecific (which, to be honest, sounds like a good idea to me; but only if we can attract other stakeholders / convince upstream to submodule it)

Other maintainers might have different thoughts on this matter though.

.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "lib/sdl/SDL2_image"]
 	path = lib/sdl/SDL2_image
 	url = https://github.com/libsdl-org/SDL_image.git
+[submodule "lib/usb/ohci_usbhostlib"]
+	path = lib/usb/ohci_usbhostlib
+	url = https://github.com/Ryzee119/ohci_usbhostlib.git
```
JayFoxRox, line 36: - Where to discuss https://github.com/Ryzee119/ohci_usbhostlib/compare/master...xbox?
- What's https://github.com/Ryzee119/ohci_usbhostlib/compare/xbox...timing?
- What are your plans for https://github.com/Ryzee119/nxdk_ohci_test/tree/master?

JayFoxRox, line 36: > This library was developed by OpenNuvoton (https://github.com/OpenNuvoton) who are a large manufacturer of ARM based embedded microcontrollers. This library was intended for use on their range of chips with OHCI hardware. I specifically imported it from https://github.com/OpenNuvoton/N9H30_emWin_NonOS/tree/master/Library/UsbHostLib with some modifications to make it xbox friendly.


With the linux stack we had major issues were our codebase is heavily outdated, leading to bugs that were already fixed upstream. But because nobody did anything for years, it all had a lot of bitrot.
So co-maintaining (or even just using it) is generally better than doing it alone.

- How "bad" would it be to submodule the entire original repository and only use the USB portion?
- Could upstream be convinced to submodule their USB subsystem?
- Are we the only ones who took the submodule? What are the plans: make it highly Xbox specific, or make it re-usable for all sorts of platforms?
- Could the Xbox specific code be moved out-of-tree of the USB stack? How hard would it be (hooks?)

I'd also appreciate if we weren't the only users, so it might be easier to have shared projects, like a libusb for it (does one already exist maybe?).

JayFoxRox, line 36: An issue with the linux stack was that it wasn't easily possible to support new devices without upstreaming support, because the driver got built as part of nxdk. So if my app needed a new USB driver I always needed people to build a custom nxdk (example: Xbox 360 gamepads, but also potentially niche custom hardware for debugging which we might not want upstream).

So, can we register drivers out-of-tree somehow? (also why I brought up libusb in another comment)

I'm already not sure if mixing a generic HID driver + XID driver is good. I mean.. it works, but it just feels dirty and when more people want to add other devices it might explode in size.
If we want to share our code with others (which, ideally, we should, as mentioned in another comment), then it would help to have a scalable solution.

lib/usb/Makefile
```diff
@@ -1,32 +1,59 @@
...
+USB_UAC_SRCS := \
+	$(USBH_DIR)/src_uac/uac_core.c \
+	$(USBH_DIR)/src_uac/uac_driver.c \
+	$(USBH_DIR)/src_uac/uac_parser.c
+
+# USB Video class driver
+USB_UVC_SRCS := \
+	$(USBH_DIR)/src_uvc/uvc_core.c \
+	$(USBH_DIR)/src_uvc/uvc_driver.c \
+	$(USBH_DIR)/src_uvc/uvc_parser.c
```
JayFoxRox, line 32: - Did you test these?
- Can we make these optional? (For smaller binaries + faster compilation + promoting the idea of having modular drivers)

lib/usb/Makefile
```diff
@@ -1,32 +1,59 @@
...
+
+# Mass Storage Class driver
+# FIXME: This driver is currently hardcoded so that it requires FATFS
+# This requirement should be removed and abstracted out.
+# USB_MSC_SRCS := \
+#	$(USBH_DIR)/src_msc/msc_driver.c \
+#	$(USBH_DIR)/src_msc/msc_xfer.c
```
JayFoxRox, line 39: - What would happen if you enabled this right now? Does it auto-format devices or something?
- If this was more modular (nxdk_usb_msc.lib) then we could already build this, but disable default inclusion in default.xbe

lib/usb/Makefile
```diff
@@ -1,32 +1,59 @@
...
+	$(USBH_DIR)/src_cdc/cdc_driver.c \
+	$(USBH_DIR)/src_cdc/cdc_parser.c
+	
+#USB Audio class driver
+USB_UAC_SRCS := \
+	$(USBH_DIR)/src_uac/uac_core.c \
+	$(USBH_DIR)/src_uac/uac_driver.c \
+	$(USBH_DIR)/src_uac/uac_parser.c
+
+# USB Video class driver
```
JayFoxRox, line 28: nit: Other comments in this file do not use leading space. I prefer it with space.

samples/usb/main.c
```diff
@@ -0,0 +1,37 @@
...
+                                                                          udev->descriptor.idProduct,
+                                                                          udev->descriptor.idVendor);
+}
+
+int main(void)
+{
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+
+    usbh_core_init();
+    usbh_install_conn_callback(&device_connection_callback, &device_disconnect_callback);
```
JayFoxRox, line 29: nit: I prefer function pointers without `&`. No particular reason.

samples/usb/main.c
```diff
@@ -0,0 +1,37 @@
...
+    debugPrint("Plug and unplug USB devices to test\n");
+
+    while (1) {
+        usbh_pooling_hubs();
+    }
+    usbh_core_deinit();
```
JayFoxRox, line 35: I believe in other samples we had discussed to put a comment for such cases, to say that this is never reached.
Ignore this comment if the other samples don't have such a comment.

.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "lib/sdl/SDL2_image"]
 	path = lib/sdl/SDL2_image
 	url = https://github.com/libsdl-org/SDL_image.git
+[submodule "lib/usb/ohci_usbhostlib"]
```
JayFoxRox, line 34: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> However is still considered draft for these reasons:

Is it still draft? On GitHub it's no longer a draft PR

JayFoxRox, line 34: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> The HID driver can properly distinguish between various Xinput devices (Duke, Steel Battalion, Xremote, Xbox360, Xbone etc)

Personally, I'm not a big fan of supporting non-original devices (like Xbox 360 and Xbox One) as part of core nxdk.
Feel free to ignore this feedback, but as a purist I think it's weird to support these.
I'm also slightly concerned with people adding PS4 gamepad drivers etc., leading to an explosion of complexity.
While I'd like those features *for some projects* myself, I don't think it's ideal for upstream inclusion in the nxdk core (which, I feel, should be a very minimalistic C toolchain.. ideally a package manager would bring these independent parts together).

JayFoxRox, line 34: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> Some very useful features for developers is support for CDC class USB-Serial adaptors for improved 'printf debugging'.

I think this deserves a sample and a replacement hal/debug.h in a follow-up.

JayFoxRox, line 34: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> Some very useful features for developers is support for CDC class USB-Serial adaptors for improved 'printf debugging'.

I think this deserves a sample and a replacement hal/debug.h in a follow-up PR.
Consider adding this as a todo for #449 or #172

JayFoxRox, line 34: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> 1. ~~Untested on 1.0 xboxes with daughterboard. I would expect this to have no issues though.~~ Has been tested ok!

So.. what kind of hardware has it been tested with? What hosts / what devices?
xpad in OpenXDK always had the issue that some 3rd party gamepads weren't supported, which was super frustrating.

JayFoxRox:
Most of my feedback so far is overly critical, because it's about long-term effects of using this codebase / design that we'll inherit.
Take it with a grain of salt.

I'll just use this top level comment to express my excitement about this changeset! Very good to see this done.
I have tried porting another USB stack in the past, so this is highly respected.

Also there's a large amount of surrounding code like tests, comments about broken features, custom made drivers which is way beyond what anybody could expect for a first revision.
*(That said, it also makes review much harder, because there's so much ecosystem to learn about)*

Some more comments abour PR description:

> A more permissive license (Apache 2.0)

I'd prefer CC0, but this is already much much much better. Thanks for looking into it :heart: 

> Proper interrupt out writes, so haptic feedback should work with no problem.

Finally! I had a bunch of trouble with the Linux driver.

---

I also have some feedback about the submodule itself, but I'd like to leave code comments for that, so I'll do that after https://github.com/XboxDev/nxdk/pull/438#discussion_r614292560
I think it might be the best option if you sent a PR to yourself, so everything is on record.

.gitmodules
JayFoxRox, line 36: On your master branch, could you rename your commit about original import of OpenNuvoton/N9H30_emWin_NonOS.git so it has a full URL + a commit hash for the latest version?

The commit hash is critical so we can later retrace where the fork began.

The URL change would be good because I'm concerned it will confuse users when the repository migrates elsewhere + it will actually make it easier to find it quickly.

lib/usb/Makefile
```diff
@@ -1,32 +1,59 @@
...
+USB_CDC_SRCS := \
+	$(USBH_DIR)/src_cdc/cdc_core.c \
+	$(USBH_DIR)/src_cdc/cdc_driver.c \
+	$(USBH_DIR)/src_cdc/cdc_parser.c
+	
+#USB Audio class driver
+USB_UAC_SRCS := \
+	$(USBH_DIR)/src_uac/uac_core.c \
+	$(USBH_DIR)/src_uac/uac_driver.c \
+	$(USBH_DIR)/src_uac/uac_parser.c
```
JayFoxRox, line 26: Does this support the communicator?

lib/usb/Makefile
```diff
@@ -1,32 +1,59 @@
...
+
+# USB Video class driver
+USB_UVC_SRCS := \
+	$(USBH_DIR)/src_uvc/uvc_core.c \
+	$(USBH_DIR)/src_uvc/uvc_driver.c \
+	$(USBH_DIR)/src_uvc/uvc_parser.c
+
+# Mass Storage Class driver
+# FIXME: This driver is currently hardcoded so that it requires FATFS
+# This requirement should be removed and abstracted out.
```
JayFoxRox, line 36: How hard will it be to wire this into the kernel driver, so we can use all filesystems that the kernel already supports?

samples/usb/Makefile
Ryzee119: Agree. This sample really just shows the absolute basic core init so it atleast builds during ci tests.
I am working on an 'all-in-one` usb sample. https://github.com/Ryzee119/nxdk_ohci_test/blob/xbox/main.c, advantage being it includes all source files in ci runs to catch issues better.

.gitmodules
Ryzee119: * I created a self PR, happy to discuss stack specifics there. https://github.com/Ryzee119/ohci_usbhostlib/pull/1
* Timing branch was a initial attempt to minimise/remove some of the blocking delays in the stack. Have not spent any time on this yet. afair the whole stack is non-block except during device connection/removal/suspend stuff
* Master branch is/was intended to be a clean copy of upsteam to track diff.

Ryzee119: > How "bad" would it be to submodule the entire original repository and only use the USB portion?

Not a huge fan. But can see the benefits. It creates alot of unused code that might confuse people. Afair there's still no good way to submodule a subdirectory?

> Could upstream be convinced to submodule their USB subsystem?

This is ideal, I have created an issue upsteam https://github.com/OpenNuvoton/N9H30_emWin_NonOS/issues/4. They have been quite active so I would expect a response within a few dayts.

> ... highly Xbox specific, or make it re-usable for all sorts of platforms?.
> Could the Xbox specific code be moved out-of-tree of the USB stack? How hard would it be (hooks?)

Being OHCI its very generic already. The only "xbox" specific parts are some kernal calls for memory allocation and interrupt handling. This could be wrapped in a platform/xbox file. Its very platform independant. I dont think it would be hard.

I quick scan of the internet, I can really see any open source projects using this stack.  Its probably used in many thousands of embedded products that we would never know about :)
OHCI is very old, and there is likely more popular EHCI implementations that people would use.

Ryzee119: Registering a new driver is very easy and can be done out of tree.
As simple as creating a `UDEV_DRV_T` struct
```
typedef struct
{
    int   (*probe) (struct iface_t *iface);  //what criteria should this driver be applied/init
    void  (*disconnect) (struct iface_t *iface);
    void  (*suspend) (struct iface_t *iface);
    void  (*resume) (struct iface_t *iface);
} UDEV_DRV_T;
```
`Then calling `usbh_register_driver(my_driver)`.

There's a bit of boilerplate code needed for each unique driver but not too bad.
HID and Xbox controller shared alot code, so just adding a simple class/subclass check in the `hid_probe` function seemed the easiest to me. This also has a benefit of just treating all HID devices in a similar way. I hope to  expose the steelbattalion via the joystick api. The xremote could be exposed in the same way.

lib/usb/Makefile
Ryzee119: It has some things in the driver such as specific header includes, `#include "ff.h"`, and calls to fatfs specific functions like `f_mount`, 
This is specific to http://elm-chan.org/fsw/ff/00index_e.html. So the only way to compile it, is to include the whole fatfs lib. Ive done this here for testing purposes https://github.com/Ryzee119/nxdk_ohci_test.

lib/usb/Makefile
Ryzee119: Audio is tested, video has not. I dont have a webcam or similar at all. It has some similaries to audio devices and uses isochronous endpoints too.

.gitmodules
Ryzee119: It was tested by some users in discord. It just testing the internal daughter which indeed just acts like a normal hub. 
Hub is connected to OHCI port 1:
Hub numbering is 3,4,2,1 as per other revs.

I have tested myself using downstream hubs and the internal daughterboard is just a commerical chip which nothing special. Without having a 1.0 myself, this is the most confident I can be.

>xpad in OpenXDK always had the issue that some 3rd party gamepads weren't supported, which was super frustrating.
Im not too sure how I can resolve this. As far as I can tell, I am doing it as correct as we understand it should be done. More exposure and users would help.

The only issue I can see is that I read/write to the appropriate interrupt pipes. Genuine controllers respect this and more inline with how the USB spec should be interpreted in my opinion, But you can also achieve this by doing control transfers. 3rd party may only respect control transfers. Im not sure what XDK does, I have a feeling it does control transfers for rumble atleast.

lib/usb/Makefile
Ryzee119: A quick review of the [wiki](https://xboxdevwiki.net/Xbox_Live_Communicator), and currently it will not. But likely only due to Microsoft changing the interface class descriptor to prevent it automatically being detected.

A review of the chipset linked in the wiki, indicates that its a audio class compliant `Both codec and headset firmware support all audio class requests.` so a solid probably will work ok with some minor xbox specific tweaks.

lib/usb/Makefile
Ryzee119: Good question, and my preferred solution but I dont know what this would involve. Some kind of wrapper around https://github.com/Ryzee119/ohci_usbhostlib/blob/xbox/src_msc/msc_xfer.c? or mapping to a memory region somehow.

.gitmodules
Ryzee119: >Most of my feedback so far is overly critical, because it's about long-term effects of using this codebase / design that we'll inherit. Take it with a grain of salt.

I can appreciate this. Thanks for taking the time to review.

> I'll just use this top level comment to express my excitement about this changeset! Very good to see this done.

Very greatful to hear this. Thank you.

JayFoxRox: > I created a self PR, happy to discuss stack specifics there. Ryzee119/ohci_usbhostlib#1

Please add this information to the PR description - I assume it could be more important than the discussion on this repository.

>> What are your plans for https://github.com/Ryzee119/nxdk_ohci_test/tree/master?
>
> Master branch is/was intended to be a clean copy of upsteam to track diff.

I should probably have made a separate comment, but I was asking about "nxdk_ohci_test" and not "ohci_usbhostlib"; trying to get an idea of the existing ecosystem.

JayFoxRox: > [...]  Afair there's still no good way to submodule a subdirectory?

No, unfortunately there isn't.
We have the same issue with https://github.com/XboxDev/nxdk-libcxx/issues/3; so we might require a solution for this anyway.

Having a kind of "synchronized GitHub mirror (of a subdirectory of monorepo + overlay of other changes)" looks like a useful tech to have, so I'm sure that somewhere, some solution exists (like https://github.com/kubernetes/git-sync + `git filter-branch` hook + generate README which documents what original commit was synchronized last).

But maybe I'm overcomplicating it?

>> How "bad" would it be to submodule the entire original repository and only use the USB portion?
>
> Not a huge fan. But can see the benefits. It creates alot of unused code that might confuse people. [...]

What else is part of upstream? Is it too large like the libcxx repository?

>> Could upstream be convinced to submodule their USB subsystem?
>
> This is ideal, I have created an issue upsteam. They have been quite active so I would expect a response within a few dayts.

Link?

> Being OHCI its very generic already. The only "xbox" specific parts are some kernal calls for memory allocation and interrupt handling. This could be wrapped in a platform/xbox file. Its very platform independant. I dont think it would be hard.

Okay, I'd imagine it like this:
```
lib/usb/libusbohci/             < upstream mirror submodule
lib/usb/libusbohci_xbox/port.c  < only the usb driver
lib/usb/Makefile                < bring upstream + our fork together
```

So we can independently update the upstream and have a clean history for our Xbox stuff.

>  OHCI is very old, and there is likely more popular EHCI implementations that people would use.

Interesting. I didn't even consider that.

So probably not many stakeholders, so might not worth it to put too much effort into making a re-usable fork.
Maybe the ps2dev people (or similar) might still be interested?

JayFoxRox: I still have to look into the other repository, but if a standard interface exists for adding USB drivers, then it might be worth it to add a similar interface for registering hid drivers (out-of-tree).

> I hope to expose the steelbattalion via the joystick api.

Are we talking about implementing the entire low-level driver in nxdk-sdl (SDL joystick), or are we talking about having a driver in the USB libs, with SDL using it?

I'd still prefer if the essential core drivers of nxdk were kept very small, so my idea of this would be:

* nxdk/lib/nxdk_usb.lib
* nxdk/lib/nxdk_usb_hid.lib
* nxdk/lib/nxdk_usb_hid_xid.lib
* nxdk/lib/nxdk_usb_hid_xid_steelbattalion.lib (if necessary)
* nxdk/lib/nxdk_sdl2.lib

Again: I'm probably overthinking it.

lib/usb/Makefile
JayFoxRox: Okay. I didn't understand that from the existing comment (although it's quite clear, I thought that it requires use of the "FatFs" filesystem type, not some libraries).

 I think we should standardize an API so that the msc driver calls elsewhere.
Because USB drivers can already be out-of-tree, it would make sense to register filesystem handling out-of-tree as well.
However, I didn't look at that code yet.

That sounds like an upstreamable solution, too.

Temporarily disabling msc is fine with me.

lib/usb/Makefile
JayFoxRox: I think you responded on the wrong comment?

lib/usb/Makefile
Ryzee119: Audio is tested, video has not. I dont have a webcam or similar at all. It has some similaries to audio devices and uses isochronous endpoints too.

.gitmodules
JayFoxRox: >  Im not sure what XDK does, I have a feeling it does control transfers for rumble atleast.

We should be finding out asap and document in XboxDevWiki.
It's ridiculous that we still don't have these facts after almost 20 years.

lib/usb/Makefile
JayFoxRox: There's https://github.com/JayFoxRox/xbox-tools/tree/4bc808e187311010f850d7fbd9af4b76bed90727/communicator-tool at the bottom of the page, which is a "working" proof-of-concept driver that I wrote from scratch (a long time ago).
On the topic of drivers; I also wrote a driver for the Xbox Remote: https://github.com/JayFoxRox/lirc-xboxlibusb/commit/67a8cc4288e58c124c4b32e50ef41d2d85992623 (I actually have another one in Python that I used instead, but I'm not sure where it is / if it's even public)

lib/usb/Makefile
JayFoxRox: This implements a very dirty kernel filesystem driver: https://github.com/JayFoxRox/xbe-loader/blob/master/xiso_driver.c (you can register a raw device, and then mount it with either a specific driver, or have the kernel probe it - so you only have to respond to block requests). Unfortunately, still somewhat unstable. Similar code exists in nkpatcher (which was unknown to me, so I ended up reverse engineering a lot of the kernel stuff independently, particularily noticing that the structs and magic numbers for constants changed between Xbox and Windows; namely https://github.com/JayFoxRox/xbe-loader/blob/master/cheat.h)

.gitmodules
JayFoxRox, line 34: *(This is a generic comment about commit style, but made on code so we can resolve it more easily)*

Your commit prefix is "USB: ", but nxdk typically uses lowercase prefixes.

lib/usb/Makefile
Ryzee119: I modified the makefile so that all class drivers are optional. HID are enabled by default, all others are enabled in a similar way to how SDL2 is enabled in nxdk.

`NXDK_USB_ENABLE_CDC=y` in makefile will enable CDC class driver for example.

I would image the user expects HID to work out of the box so this is enabled by default but can be disabled by `NXDK_USB_DISABLE_HID=y`

I wasnt really sure if this is a good implementation. If you had anything in mind let me know.

samples/usb/main.c
Ryzee119: I added a comment to advise it is never reached. mesh sample has a similar comment.

Just checked a few samples, they seem to use `while(1)`, so will keep it as is.

.gitmodules
Ryzee119: All should be lowercase now

samples/usb/main.c
Ryzee119: They are the number of the USB hub port. I have updated the wording here.

samples/usb/main.c
Ryzee119: updated the debug print here to make it clearer what it is doing

samples/usb/main.c
Ryzee119: Makes sense to me. Updated this

.gitmodules
Ryzee119: > What are your plans for https://github.com/Ryzee119/nxdk_ohci_test/tree/master?
I think it could make a good nxdk sample if/when i clean it up to it suitable quality. Probably something I will do in the near future.

Ryzee119: So update on this side of this. The issue upsteam was raised here regarding a separate usb host lib repo to allow a clean fork https://github.com/OpenNuvoton/N9H30_emWin_NonOS/issues/4

The response wasnt negative but not a clear positive either; (or a timeline), but I'm happy to wait a week or two incase they do this as this is really my preferred solution.

Secondly,

>lib/usb/libusbohci/             < upstream mirror submodule
lib/usb/libusbohci_xbox/port.c  < only the usb driver
lib/usb/Makefile                < bring upstream + our fork together

I like this idea, and will probably make it like this. The "upstream mirror submodule" will likely be my own forked submodule of "upstream" with various commits to make it xbox friendlier but not platform specific.

This will take me a bit of work, so any preferrences or further details on this let me know.

Ryzee119: I make the neccessary changes so now the repo/submodule cofiguration looks like:

>lib/usb/libusbohci (submoduled lib)
lib/usb/libusbohci_xbox/port.c (Xbox specific functions/config)
lib/usb/Makefile

.gitmodules
dracc: FTR, I've tested on my 1.0 using a Duke, controller S, a Logitech wheel ("Precision Vibration Feedback Wheel") and the Logitech wireless controller ("Cordless Precision Controller").
All worked great, rumble and all.

espes:
This is fantastic!

> Currently the ohci library is included as a submodule to my personal repo. This should either be migrated here or commited directly into nxdk (or it can stay linked to my repo, I dont mind)

I've invited you to the xboxdev org if you want to move it

.gitmodules
Ryzee119: xid xbox specifc driver added to lib/usb/libusbohci_xbox. Upstream HID class is just generic HID class now.

Now all xbox specific parts are in `lib/usb/libusbohci_xbox`.

I think this makes a cleaner way for adding xbox specific drivers in the future (like communicator, camera etc) without polluting the ohci lib too much.

JayFoxRox:
This PR was merged in a bad state - https://github.com/XboxDev/nxdk-sdl/commit/60ff4108911ed4028c27fd910e476d29ab679525 is not part of the XboxDev git repo. GitHub only finds it because it's part of the same fork tree.

PR #440, opened 2021-03-27
build: Remove -Wno-ignored-attributes clang parameter

thrimbor:
This removes the `-Wno-ignored-attributes` parameter from `NXDK_CFLAGS`. It doesn't appear to produce any difference when building nxdk code, not even when also using `-Wall -pedantic`.


Ryzee119:
Ran this though a few nxdk homebrew apps, no differences for me either. Looks good.

PR #441, opened 2021-03-31
vcruntime: Implement stack smashing detection

thrimbor:
This implements stack smashing detection support. When stack protection is enabled, this will catch attempts to overwrite the return address on the stack by overflowing stack buffers.
At the moment, a static default value is used as the canary value, but a call to `__security_init_cookie()` will be added to PDCLib soon.

To test, add the `-fstack-protector-all` clang parameter and try the following code:
```c
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>

void tfunc ()
{
    char buf[3];
    memcpy(buf, "abcdefghiuaoiegboegboub", 16); // Buffer overflow
}

int main(void)
{

    while(1) {
        debugPrint("Hello nxdk!\n");
        tfunc();
        Sleep(2000);
    }

    return 0;
}
```

### DIFF
--- a/lib/xboxrt/Makefile
+++ b/lib/xboxrt/Makefile
@@ -19,6 +19,7 @@ XBOXRT_SRCS := \
 	$(NXDK_DIR)/lib/xboxrt/c_runtime/check_stack.c \
 	$(NXDK_DIR)/lib/xboxrt/c_runtime/chkstk.s \
 	$(NXDK_DIR)/lib/xboxrt/vcruntime/purecall.c \
+	$(NXDK_DIR)/lib/xboxrt/vcruntime/stack_protection.c \
 	$(NXDK_DIR)/lib/xboxrt/vcruntime/threadsafe_statics.c \
 	$(NXDK_DIR)/lib/xboxrt/vcruntime/vcruntime_exception.cpp \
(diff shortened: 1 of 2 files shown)


Ryzee119:
Ive tried this in xemu and my 1.4 xbox. Works as expected.
I modified your sample to remove the overflow and it didnt assert as expected. Increased to 4 (1 byte overflow and it triggerd successfully)

I went stupid and caused an overflow of 1000 bytes and it no longer worked, but it worked at 950 bytes. May be just too much damage to the memory at that point.

image

Ryzee119:
I figured that was the case.
Can confirm it did hang in the case of the  excessive 1000byte overflow

thrimbor:
Ah, that's fine then. You're getting an exception for that memory access, but you can't catch it because the handler registration also resides on the stack, so it's overwritten too. When you don't have a handler, the kernel will catch the exception and stop the program.
SafeSEH, once I implement it, will detect overwritten handler registrations, so that it won't be possible for an attacker to overwritte them, and then trigger an exception on purpose to get RCE (which would currently be possible, because triggering an exception bypasses the normal function return and thus skips the stack cookie check).

thrimbor:
I'll merge this early now, so I can merge all the PDCLib PRs (https://github.com/XboxDev/nxdk-pdclib/pull/42 needs to be merged after this).

PR #445, opened 2021-04-04
xboxrt: Make assembly files SafeSEH-compatible

thrimbor:
This modifies the assembly files to include the special symbol and value that marks them as SafeSEH compatible. It also removes the linker parameter that was necessary previously.

This does not implement or enable SafeSEH, it only makes the code compatible.

### DIFF
--- a/lib/xboxrt/c_runtime/_alldiv.s
+++ b/lib/xboxrt/c_runtime/_alldiv.s
@@ -16,6 +16,9 @@
 
 // Modified for stdcall calling convention for use in nxdk
 
+.include "prelude.s.inc"
+safeseh_prelude
+
 .text
 .balign 4
 .globl __alldiv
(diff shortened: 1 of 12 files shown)


lib/xboxrt/c_runtime/_alldiv.s
```diff
@@ -16,6 +16,16 @@
...
 
+// This creates a special symbol with the least significant bit set.
+// It signals to the linker that this code is SafeSEH-compatible.
+.text
+.def @feat.00;
+.scl  3; // IMAGE_SYM_CLASS_STATIC
+.type 0; // IMAGE_SYM_TYPE_NULL
+.endef
+.globl @feat.00
+.set @feat.00, 1
```
JayFoxRox, line 27: Can't this be included from a `prelude.s` or be a macro or something?

thrimbor: Done.

JayFoxRox: Cheers. Much better.

*(Minor complaint: Filename could be a generic `prelude.s` if you still have to apply the macro afterwards anyway; but I don't care too much either)*

JayFoxRox: Will merge if you don't plan to rename file (as in my minor complaint)
Although, I feel more strongly by now, because I don't want to add more and more files to contain re-usable macros, if we can just have one `prelude.s` which contains all the common 3-5 line macros.

thrimbor: I've renamed the file now, but kept the `.s.inc` extension, I want to avoid people thinking that this is a normal assembly code file.

PR #446, opened 2021-04-05
samples: Allow overriding NXDK_DIR

thrimbor:
This allows overriding the `NXDK_DIR` variable in the samples Makefiles by using the `?=` operator, which only assigns a value if the variable is not defined already. It allows to build the samples with the Docker image and sets a better example.

### DIFF
--- a/samples/hello/Makefile
+++ b/samples/hello/Makefile
@@ -1,5 +1,6 @@
 XBE_TITLE = nxdk\ sample\ -\ hello
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.c
-NXDK_DIR = $(CURDIR)/../..
+NXDK_DIR ?= $(CURDIR)/../..
+
 include $(NXDK_DIR)/Makefile
--- a/samples/mesh/Makefile
+++ b/samples/mesh/Makefile
@@ -2,6 +2,6 @@ XBE_TITLE = nxdk\ sample\ -\ mesh
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/math3d.c $(CURDIR)/main.c
 SHADER_OBJS = ps.inl vs.inl
-NXDK_DIR = $(CURDIR)/../..
+NXDK_DIR ?= $(CURDIR)/../..
 
 include $(NXDK_DIR)/Makefile
--- a/samples/sdl_gamecontroller/Makefile
+++ b/samples/sdl_gamecontroller/Makefile
@@ -1,7 +1,7 @@
 XBE_TITLE = nxdk\ sample\ -\ SDL2_GameController
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.c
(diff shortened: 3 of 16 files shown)


samples/hello/Makefile
```diff
@@ -1,5 +1,6 @@
 XBE_TITLE = nxdk\ sample\ -\ hello
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.c
-NXDK_DIR = $(CURDIR)/../..
+NXDK_DIR ?= $(CURDIR)/../..
+CFLAGS = -fstack-protector-all
```
dracc, line 5: Was this line supposed to be added here?

thrimbor: Nope, left that in accidentally. It's fixed now.

dracc (approved):
Looks good to me. 👍

samples/mesh/Makefile
```diff
@@ -2,6 +2,6 @@ XBE_TITLE = nxdk\ sample\ -\ mesh
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/math3d.c $(CURDIR)/main.c
 SHADER_OBJS = ps.inl vs.inl
-NXDK_DIR = $(CURDIR)/../..
+NXDK_DIR ?= $(CURDIR)/../..
```
Teufelchen1, line 6: Might want to delete this empty line in order to unify this Makefile with the majority.

samples/sdl_gamecontroller/Makefile
```diff
@@ -1,7 +1,7 @@
 XBE_TITLE = nxdk\ sample\ -\ SDL2_GameController
 GEN_XISO = $(XBE_TITLE).iso
 SRCS = $(CURDIR)/main.c
-NXDK_DIR = $(CURDIR)/../..
+NXDK_DIR ?= $(CURDIR)/../..
 NXDK_SDL = y
```
Teufelchen1, line 6: Might want to delete this empty line in order to unify this Makefile with the majority.

thrimbor:
I actually prefer the visual separation the newline provides, so instead of removing those two, I added them to the other Makefiles.

JayFoxRox:
> [...] and sets a better example.

Agreed for real-world applications; but slightly disagree for samples.

I'm torn on this; when the user clones a temporary copy of nxdk for testing, they might run samples in that fork and expect different results.
I consider this a bigger use-case than the docker situation.

However, I don't feel strongly about this, so I'm also fine with accepting it. I just wanted to raise my concerns about this use-case.
I'll merge this on the weekend if there's no veto.

thrimbor:
My main concern was that people will base their own Makefiles on the ones from the samples, and then hit issues or hardcode the paths from their development machine, causing issues for others that want to build it. But I just noticed it's also a problem for how I want to test the Docker image - I'm currently working on images that contained prebuilt nxdk versions without the samples, but I still want to test them by building the samples with them.

Due to the latter I'm self-merging this now.

PR #450, opened 2021-04-10
.gitmodules: Make submodules shallow, fix url

Margen67:


thrimbor:
I don't think we should make submodules shallow by default.

Margen67:
> 
> 
> I don't think we should make submodules shallow by default.

Why?

thrimbor:
Because it requires the user to undertake additional steps before they can use tools that I think are essential, such as bisection. From my perspective, shallow cloning serves one of two purposes, either saving bandwidth or disk space. Imho these are better addressed in different ways, like shallow cloning in CI, manual shallow cloning by the user if disk space or bandwidth is tight, reducing the overall amount of submodules we carry in the main nxdk repo etc.

JayFoxRox:
I agree with @thrimbor.

Users are already told to shallow-clone here: https://github.com/XboxDev/nxdk/wiki/Install-the-Toolchain.
The README is just outdated *again*. therefore I decided to reopen https://github.com/XboxDev/nxdk/issues/73#issuecomment-612397420 (due to a lack of counter-arguments).

PR #451, opened 2021-04-10
ci: Update existing GitHub workflows

Margen67:
```
.ci_build_samples.sh:
 Use one less cd command.
 chmod +x
Upgrade checkout to v2.
Use -e when invoking shell(s).
Run .ci_build_samples.sh directly.
build_samples.yml:
 Don't use multi-line syntax for one command.
 Windows:
  Opt out of PowerShell telemetry.
  Use curl instead of Invoke-WebRequest.
 macOS:
  Move exported variables to env:.
  Opt out of brew analytics.
 Simplify ci_build_samples commands.
 Ubuntu:
  Remove unneeded packages.
  Remove unneeded PATH addition.
```

### DIFF
--- a/.github/workflows/build_docker_image.yml
+++ b/.github/workflows/build_docker_image.yml
@@ -12,9 +12,8 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Clone Tree
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
-          fetch-depth: 1
           submodules: recursive
       - name: Generate tags
         id: tags
--- a/.github/workflows/build_samples.yml
+++ b/.github/workflows/build_samples.yml
@@ -9,70 +9,67 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Clone Tree
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
-          fetch-depth: 1
           submodules: recursive
       - name: Build Docker image
-        run: |
-          docker build -t nxdk ./
+        run: docker build -t nxdk ./
       - name: Test Docker image
-        run: |
-          docker run nxdk sh -c "cd /usr/src/nxdk && sh ./.ci_build_samples.sh"
+        run: docker run nxdk sh -ec "cd /usr/src/nxdk && ./.ci_build_samples.sh"
   windows:
     name: Windows
     runs-on: windows-latest
     timeout-minutes: 20
+    env:
+      POWERSHELL_TELEMETRY_OPTOUT: 1
     steps:
       - name: Clone Tree
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
-          fetch-depth: 1
           submodules: recursive
       - name: Install MSYS2 & Dependencies
         run: |
           echo "Downloading MSYS2 environment..."
-          Invoke-WebRequest -Uri "https://github.com/XboxDev/nxdk-ci-environment-msys2/releases/latest/download/msys64.7z" -OutFile "msys64.7z"
+          curl -fLO "https://github.com/XboxDev/nxdk-ci-environment-msys2/releases/latest/download/msys64.7z"
           echo "Extracting MSYS2 environment..."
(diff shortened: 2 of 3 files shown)


JayFoxRox:
While you are at it, can you also do https://dev.to/github/running-github-actions-ci-cd-triggers-on-specific-branches-5e6m (or similar)?

I trigger 8 different builds when I push to my own repository while having a PR to my own repository: 4x for `push` and 4x for `pull_request`. That push event should probably only happen when pushing to master.

Feel free to ignore this request at your discretion, it can also be done at a different time.

.github/workflows/build_samples.yml
```diff
@@ -9,70 +9,67 @@ jobs:
...
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
-          fetch-depth: 1
           submodules: recursive
       - name: Install Dependencies
         run: |
-          sudo apt-get update -y && sudo apt-get install -y flex bison clang llvm lld
+          sudo apt-get -y update
+          sudo apt-get -y install llvm lld
```
thrimbor, line 73: What about the other dependencies?

.github/workflows/build_docker_image.yml
```diff
@@ -12,9 +12,8 @@ jobs:
     timeout-minutes: 20
     steps:
       - name: Clone Tree
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
-          fetch-depth: 1
```
thrimbor, line 17: Why the removal?

thrimbor:
> I trigger 8 different builds when I push to my own repository while having a PR to my own repository: 4x for `push` and 4x for `pull_request`. That push event should probably only happen when pushing to master.

I would like to keep the push event, I rarely do self-PRs but still want to have my code built. Maybe introducing [a branch filter](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags) would be an option (something like skipping all branches with a `-noci` suffix)?

JayFoxRox:
> I would like to keep the push event, I rarely do self-PRs but still want to have my code built.

Fair argument. I didn't think this through.

> Maybe introducing a branch filter would be an option

Not necessary; It just complicates stuff and I'd not use it.
I'll just ignore the redundant CI builds.

.github/workflows/build_docker_image.yml
Margen67: checkout v2 is fetch-depth 1 by default.

.github/workflows/build_samples.yml
Margen67: This is the minimum needed to build. flex and bison are already installed: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#installed-apt-packages

thrimbor: My concern is that these are subject to change with newer runner releases, but I guess we cannot expect to be able to upgrade without breakage anyway.

JayFoxRox:
I think this can be merged, but it needs the commit to be cleaned up:
- Prefix style
- Avoid long commit description (= we usually prefer a single line commit description)

Margen67:
> 
> 
> I think this can be merged, but it needs the commit to be cleaned up:
> 
>     * Prefix style
> 
>     * Avoid long commit description (= we usually prefer a single line commit description)

Should I split each change so it only takes up 1-2 lines in the commit description?

JayFoxRox:
Sorry for not getting back earlier. I try to minimize my involvement with Xbox stuff and was hoping someone else would respond.

> Should I split each change so it only takes up 1-2 lines in the commit description?

However you like. I don't mind a single commit for small changes like this, but it should follow the existing commit style (= prefix, no long description, commit title is punctual which is short enough to not cause a break on GitHub).

Renaming the existing commit to "ci: Update existing GitHub workflows" is totally fine *with me*.

JayFoxRox:
Commit still has the long multi-line description. When I find time for it and it isn't removed I'll force push it away and merge this PR (unless there's a veto).

thrimbor:
Pushed away the long description, merging now.

PR #455, opened 2021-04-11
submodules: Use new SDL_ttf and SDL2_image repository

Margen67:


.gitmodules
```diff
@@ -28,6 +25,9 @@
 [submodule "lib/libjpeg/libjpeg-turbo"]
 	path = lib/libjpeg/libjpeg-turbo
 	url = https://github.com/libjpeg-turbo/libjpeg-turbo.git
-[submodule "lib/sdl/SDL_image"]
+[submodule "lib/sdl/SDL_ttf"]
+	path = lib/sdl/SDL_ttf
+	url = https://github.com/libsdl-org/SDL_ttf.git
```
JayFoxRox, line 30: For easier review:

- Old: https://github.com/SDL-mirror/SDL_ttf/tree/8f8753c4288948d1fd483d7de23e36c7b30a7807
- New: https://github.com/libsdl-org/SDL_ttf/tree/200901a8f55bcf7aa031157f53a575c942d2b7aa

Looks good.

Worth noting: this is between version 2.0.14 and 2.0.15
We should consider updating to 2.0.15 in a separate step later.

.gitmodules
```diff
@@ -28,6 +25,9 @@
...
+[submodule "lib/sdl/SDL2_image"]
 	path = lib/sdl/SDL2_image
-	url = https://github.com/SDL-mirror/SDL_image.git
+	url = https://github.com/libsdl-org/SDL_image.git
```
JayFoxRox, line 33: For easier review:

- Old: https://github.com/SDL-mirror/SDL_image/tree/b4f28c1fc9ab2f91ae594fc3e86622c5c13ffbbb
- New: https://github.com/libsdl-org/SDL_image/tree/ab2a9c602623193d61827ccd395302d92d90fc38

Looks good.

Worth noting: this is version https://github.com/libsdl-org/SDL_image/tree/release-2.0.5
This is the latest version available, so no further action is necessary.

JayFoxRox:
Thanks again for all the housekeeping. 

Commit title must be rewritten to match prefix style.
I also don't like the "SDL_ttf+SDL2_image" look and would prefer "SDL_ttf and SDL2_Image".

Issue #454, opened 2021-04-11
Network stack should be a static library


JayFoxRox:
Most parts of nxdk have been lib-ified.
However, the network stack is only setting variables to add sources and flags to the application: https://github.com/XboxDev/nxdk/blob/72485f402b7f2b486430af9236a26fdef5ef7bd7/lib/net/Makefile#L16-L24

This is a problem when someone wants to use the network stack without the nxdk makefile, because they have to bring their own build-system for pktdrv, nforceif and lwip.

Instead, the network stack should be build as a library like other libs:
- USB: https://github.com/XboxDev/nxdk/commit/089401fa7ad366b4fc5fc61589644d1e4ee3a4ae
- HAL: https://github.com/XboxDev/nxdk/commit/1bab3a78e6cb62eb282d1d659d8570a08ed01b6b
- ...

PR #457, opened 2021-04-16
net: Use own Makefile, build as library

Teufelchen1:
Hi 🙋‍♀️ 

This PR addresses #454 by adjusting the `nxdk/lib/net/Makefile` to yield a `nxdk/lib/libnxdk_net.lib`.

I'm unsure about my work:

- I could have moved the inclusion of `nxdk/lib/net/Makefile` from `nxdk/Makefile` to `nxdk/lib/Makefile`
- The include `CFLAGS` for net are in the net-Makefile, I could have moved them into `nxdk/Makefile`(following the other libs examples(hal, pdclib, winapi, etc.). But that feels wrong - so I wonder, why are the other libs included there and not in their own Makefile(i.e. like libpng does)?
- Why does the net-Makefile has extra `CXXFLAGS` while other libraries don't? Can't find the reason 🙈😅
- Why can we do `#include <hal/debug.h>` but do `#include "lwip/debug.h"` in the samples? ("" vs <>) I switched it to <> and it still works. Note: I know the general difference between these two include methods, I just don't understand it in this context(nxdk, complex build systems, libs, etc. etc.).

___

I currently have no methods of running xbes, so I could only test by compiling.

### DIFF
--- a/lib/net/Makefile
+++ b/lib/net/Makefile
@@ -5,20 +5,33 @@ NETIFFILES := $(filter-out $(LWIPDIR)/netif/slipif.c,$(NETIFFILES))
 LWIPSRCS := $(COREFILES) \
             $(CORE4FILES) \
             $(CORE6FILES) \
-            $(NETIFFILES) \
-            $(APIFILES)
+            $(APIFILES) \
+            $(NETIFFILES)
 
 # Include driver sources
 DRIVERSRCS := $(NXDK_DIR)/lib/net/pktdrv/pktdrv.c \
               $(NXDK_DIR)/lib/net/nforceif/src/driver.c \
               $(NXDK_DIR)/lib/net/nforceif/src/sys_arch.c
 
-SRCS += $(LWIPSRCS) $(DRIVERSRCS)
+NET_SRCS := \
+	$(LWIPSRCS) \
+	$(DRIVERSRCS)
 
-# Add includes
-CFLAGS += -I $(LWIPDIR)/include
-CFLAGS += -I $(NXDK_DIR)/lib/net/nforceif/include
-CFLAGS += -I $(NXDK_DIR)/lib/net/pktdrv
-CXXFLAGS += -I $(LWIPDIR)/include
-CXXFLAGS += -I $(NXDK_DIR)/lib/net/nforceif/include
-CXXFLAGS += -I $(NXDK_DIR)/lib/net/pktdrv
+NET_OBJS = $(addsuffix .obj, $(basename $(NET_SRCS)))
+
+NXDK_CFLAGS += -I$(LWIPDIR)/include
+NXDK_CFLAGS += -I$(NXDK_DIR)/lib/net/nforceif/include
+NXDK_CFLAGS += -I$(NXDK_DIR)/lib/net/pktdrv
+NXDK_CXXFLAGS += -I$(LWIPDIR)/include
+NXDK_CXXFLAGS += -I$(NXDK_DIR)/lib/net/nforceif/include
+NXDK_CXXFLAGS += -I$(NXDK_DIR)/lib/net/pktdrv
+
+$(NXDK_DIR)/lib/libnxdk_net.lib: $(NET_OBJS)
+
+main.exe: $(NXDK_DIR)/lib/libnxdk_net.lib
+
+CLEANRULES += clean-net
+
+.PHONY: clean-net
+clean-net:
+	$(VE)rm -f $(NET_OBJS) $(NXDK_DIR)/lib/libnxdk_net.lib


lib/net/Makefile
```diff
@@ -5,20 +5,33 @@ NETIFFILES := $(filter-out $(LWIPDIR)/netif/slipif.c,$(NETIFFILES))
...
               $(NXDK_DIR)/lib/net/nforceif/src/sys_arch.c
 
-SRCS += $(LWIPSRCS) $(DRIVERSRCS)
+NET_SRCS := \
+	$(LWIPSRCS) \
+	$(DRIVERSRCS)
+
+NET_OBJS = $(addsuffix .obj, $(basename $(NET_SRCS)))
 
-# Add includes
```
Teufelchen1, line 18: Removed that by accident, but wasn't that useful anyways.

lib/net/Makefile
```diff
@@ -5,20 +5,33 @@ NETIFFILES := $(filter-out $(LWIPDIR)/netif/slipif.c,$(NETIFFILES))
 LWIPSRCS := $(COREFILES) \
             $(CORE4FILES) \
             $(CORE6FILES) \
-            $(NETIFFILES) \
-            $(APIFILES)
+            $(APIFILES) \
+            $(NETIFFILES)
```
Teufelchen1, line 9: This switch was made to represent the order given in `Filelist.mk` of lwip.

lib/net/Makefile
```diff
@@ -5,20 +5,33 @@ NETIFFILES := $(filter-out $(LWIPDIR)/netif/slipif.c,$(NETIFFILES))
...
 
 # Include driver sources
 DRIVERSRCS := $(NXDK_DIR)/lib/net/pktdrv/pktdrv.c \
               $(NXDK_DIR)/lib/net/nforceif/src/driver.c \
               $(NXDK_DIR)/lib/net/nforceif/src/sys_arch.c
 
-SRCS += $(LWIPSRCS) $(DRIVERSRCS)
+NET_SRCS := \
```
JayFoxRox, line 16: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> - The include `CFLAGS` for net are in the net-Makefile, I could have moved them into `nxdk/Makefile`(following the other libs examples(hal, pdclib, winapi, etc.). But that feels wrong - so I wonder, why are the other libs included there and not in their own Makefile(i.e. like libpng does)?

The network stuff are optional includes and only necessary if the user actually needs network access.
So adding them here is better, because this Makefile is conditionally included.

The others are part of making a C/Windows toolchain, so they are always mandatory (although we might eventually want a `-freestanding` option for something like my bin-wrappers).

JayFoxRox, line 16: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> - Why does the net-Makefile has extra `CXXFLAGS` while other libraries don't? Can't find the reason 🙈😅

Because someone made a C++ project and wanted to use these libs + these headers are probably C++ aware already.

JayFoxRox, line 16: *(This is a generic comment about PR description, but made on code so we can resolve it more easily)*

> - Why can we do `#include <hal/debug.h>` but do `#include "lwip/debug.h"` in the samples? ("" vs <>) I switched it to <> and it still works. Note: I know the general difference between these two include methods, I just don't understand it in this context(nxdk, complex build systems, libs, etc. etc.).

I believe it often comes down to different conventions.

A library that expects to be bundled with the app will often use `""` and a library which is meant for system-wide install (or toolchain installation) is typically `<>`.
I assume lwip, as an embedded lib, is often bundled with the project (like a firmware), so `""` might make more sense then.

Since pkg-config and CMake came along those lines got blurred, because some people might have SDL systemwide, and others don't. Also some will have their SDL header in a path with other stuff, so now you have to listen to pkg-config or CMake (which recommend `#include "SDL.h"`, even if some platforms claim that it should be `#include <SDL2/SDL.h>`)

Different tools also handle it differently; here are the GCC rules: https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html
Some interesting notes:
- MSVC++ also likes to look in parent directories or something weird; so it finds files in different order. I believe you can even trip it by putting stuff in some root directory like `C:/`
- Some compilers ignore warnings when you use `<>`, others will not make this decision at time of use, but at time of include dir addition (like GCC with `-isystem` vs. `-i`).

I'm sure that @thrimbor will correct me on all of this though. He's usually more standard aware than I am.

---

Random rant of the day: With generic names like `debug.h`, `input.h`, `audio.h` etc. there can also be conflicts, which is why I absolutely hate this (which allows `#include "debug.h"` or `#include <debug.h>` when the only right thing should probably be `#include <hal/debug.h>`):
https://github.com/XboxDev/nxdk/blob/9519fbd8c259792e0fc56244c69b1f13a86fe97e/Makefile#L57

Ryzee119: This has caught me out a few times that 
I started changing it. 
My code had its own "debug.h" but was including the nxdk's one instead causing a confusimg set of compilation errors.

Maybe ill rebase and raise a PR.
https://github.com/Ryzee119/nxdk/pull/1/commits

JayFoxRox: > Maybe ill rebase and raise a PR.
> https://github.com/Ryzee119/nxdk/pull/1/commits

Please upstream. I also had https://github.com/JayFoxRox/nxdk/pull/61 for a long time (also see failed https://github.com/XboxDev/nxdk/pull/258 for another approach).
I keep running into this and it gets even worse with autoconf and cmake (which is how I primarily use nxdk).

thrimbor: > I'm sure that @thrimbor will correct me on all of this though. He's usually more standard aware than I am.

What you said sounds about right. The most notable difference afaik is that `<>` only searches the include directories given to the compiler while `""` searches relative to the code file first.

For nxdk, I think it's best to use `<>` everywhere when including a library's header, except for special cases, e.g. when including a header within a library that is meant to be private to the library or something.

> Random rant of the day: With generic names like `debug.h`, `input.h`, `audio.h` etc. there can also be conflicts, which is why I absolutely hate this (which allows `#include "debug.h"` or `#include <debug.h>` when the only right thing should probably be `#include <hal/debug.h>`)

Agreed. I recall adding that include directory because we moved the header and it conflicted with the way a submodule (SDL I believe) included it, but I don't remember whether we fixed that issue in the meantime. We should check and remove that include path if we can.

JayFoxRox: Resolving after #461 has been merged

lib/net/Makefile
```diff
@@ -5,20 +5,33 @@ NETIFFILES := $(filter-out $(LWIPDIR)/netif/slipif.c,$(NETIFFILES))
...
+
+NET_OBJS = $(addsuffix .obj, $(basename $(NET_SRCS)))
 
-# Add includes
 CFLAGS += -I $(LWIPDIR)/include
 CFLAGS += -I $(NXDK_DIR)/lib/net/nforceif/include
 CFLAGS += -I $(NXDK_DIR)/lib/net/pktdrv
 CXXFLAGS += -I $(LWIPDIR)/include
 CXXFLAGS += -I $(NXDK_DIR)/lib/net/nforceif/include
 CXXFLAGS += -I $(NXDK_DIR)/lib/net/pktdrv
```
JayFoxRox, line 27: I believe this must be `NXDK_` prefixed; we do this, so the user can overwrite `CFLAGS` and friends with their own stuff, without breaking compilation.
Also relevant for people who only run the makefile to extract the NXDK variables, so they can hook their own build-system into it.

JayFoxRox: These days, we also do these without space between `-I` and the path

Issue #449, opened 2021-04-10
Reimplementing Hal code


DobaMuffin:
Currently most of the code in the lib/hal folder is licensed under the GPL license due to coming from OpenXDK. To allow for closed sourced programs to be created using the nxdk, all GPL'd code needs to be replaced. Since a replacement USB stack is being worked on, this leaves the hal code. At some point in the future, the hal code would need to be re-implemented and licensed under a different license.

JayFoxRox:
There's barely any HAL code of interest left I think.
Also most of it is known to have been stolen elsewhere (so GPL is probably not even the right license) or it's unused or largely rewritten.

So for the HAL code I suggest to slim it down first: deprecating and removing code.
I don't think many people need it for compatibility, and if they do, it's probably possible to replace deprecated functions with wrappers around winapi.

Once it's smaller, review it with the authors who have reworked it and consider if it's legally and ethically acceptable to just slap another license on it.
If relicensing is not an option: rewrite the affected parts.

JayFoxRox:
I've gone through these alphabetically:

## audio.c

For what it's worth, the algorithms in https://github.com/XboxDev/nxdk/blame/231336bfba1cbc66ae7259f8a27a81448ef5421d/lib/hal/audio.c are largely by me (see #364).
The remaining code is probably factual (constants, memory addresses, variable names), so I don't think it can actually be copyrighted (Disclaimer: not-a-lawyer).

Therefore, consider my code in https://github.com/XboxDev/nxdk/blame/231336bfba1cbc66ae7259f8a27a81448ef5421d/lib/hal/audio.c as public domain; feel free to relicense my code under *whatever* license you choose, claim copyright or whatever else is necessary.

## debug.c

This can be rewritten in an hour or so; it's trivial and would likely end up being better than the existing code.
There might also be existing libs which could be used for this, so we should potentially just deprecate this and remove it eventually.

The worst offender in this is probably `debugPrint` which is cancerously spreading throughout the nxdk codebase.
Also see #172, to discuss a more permanent solution.

## fileio.c

This had been deprecated for 2 years already: https://github.com/XboxDev/nxdk/commit/8e5a12fe49ef6c1f6ede6f26f8797b35c5940fd3
Replacements exist in winapi.

I think it can (and should be) removed.

## fonts

*(PD non-issue)*

Exception: A deprecation `#warning` should be added for "font_sasteroids2.h" asap.
I'm not sure what license it has, but I don't think it has any users either.

[...]

PR #458, opened 2021-04-16
hal: Mark outdated functions as deprecated and remove unused font

Teufelchen1:
Hi 🙋‍♀️ 

Following JFRs analysis in issue #449, I marked a bunch of functions as deprecated. 
I also added messages to commonly used function on how to replace them. 

As I wasn't sure how to deprecate a font, I just added the deprecated attribute to the top of the file `font_sasteroids2.h`

Confusingly, JFRs analysis included functions that have already been removed in the past. I took no further actions regarding them. 

After this PR(or in this PR), some samples need to be adjusted as they use now deprecated functions 😅

___
Edit:
For the interested reader, the clang deprecated attribute is documented [here](https://clang.llvm.org/docs/AttributeReference.html#deprecated).


lib/hal/debug.h
```diff
@@ -23,13 +23,21 @@ extern "C"
  * Prints a message to whatever debug facilities might
  * be available.
  */
+__attribute__((deprecated("deprecated as of 05/2021", "Use SDL2 for generic text output")))
```
JayFoxRox, line 26: The SDL2 message is a bit vague / misleading. There are 2 primary uses for hal/debug.h:

- Debugging:  also used within nxdk, no replacement exists. SDL2 is not suitable here.
- Dumb UI: also used in nxdk samples. Replacements do exist, in the form of other rendering libraries. SDL2 is suitable, but also typically overkill.

I'd probably also not deprecate `debugPrint` just yet.
Rest should be deprecated though.

I believe there are also variables in hal/debug.c that people still like to use (to move the cursor around). That has always been a hack though.

JayFoxRox, line 26: Surprised you can have comma separated messages? GNU gcc man-pages say:

``` 
deprecated
deprecated (msg)
```

Also not sure if the message should contain the date like that.

lib/hal/audio.h
```diff
@@ -34,9 +34,13 @@ typedef struct
 // note that I currently ignore sampleSizeInBits and numChannels.  They
 // are provided to cope with future enhancements. Currently supported samples
 // are 16 bit, 2 channels (stereo)
+__attribute__((deprecated("deprecated as of 05/2021", "Use SDL2 for audio output")))
```
JayFoxRox, line 37: *(This is a generic comment about the PR description, that I'm leaving here, so it can be resolved / hidden more easily)*

> Confusingly, JFRs analysis included functions that have already been removed in the past. I took no further actions regarding them.

Such as?

I believe I started looking at master and only used GitHub search rather passively, so I'm not sure while I'd have looked at a wrong revision.

lib/hal/font_sasteroids2.h
```diff
@@ -3,6 +3,7 @@
 // www.svgalib.org site. Graciously provided
 // by John Remyn, originally distributed
 // with sasteroids.
+__attribute__((deprecated("deprecated as of 05/2021", "Use font_terminal.h or font_unscii_16.h instead")))
```
JayFoxRox, line 6: `font_unscii_16.h` is not a drop-in replacment, because it has different height.

lib/hal/debug.h
GXTX: So instead of just using `debugPrint` i'll have to initiate an entire SDL instance (wasting memory) just to display text on the screen? I don't like this. What's the replacement for `debugPrint`.

JayFoxRox: > - an entire SDL instance (wasting memory)
> [...]
> Not just SDL, I'd have to init SDL_TTF which is more overhead,

You have a fixed 64MiB anyway and it's not like you are stealing it from another process.
Chances are you won't be hitting memory bounds when doing a console application.

I'd be more concerned with amount of boilerplate code.

> TTF is quite slow at drawing text to a surface

Unless you are making an ASCII renderer (but even then the hal/debug stuff is not suitable for you) your text rendering performance won't matter - no user can read as fast as you dump out pixels.
I'm probably the worst offender when it comes to debug message spam during game init on neverball and SWE1R; but I can also disable those prints if they are too slow.

> I don't like this. What's the replacement for `debugPrint`.

I'd prefer to remove it sooner than later.
So even if it temporarily stays (and isn't "officially" deprecated yet), consider it a stupid move to use it.
Nobody prevents you from taking hal/debug.c into your own project, it's a single file to copy.
So the maintenance overhead, unclean design and license shenanigans are not worth to keep it long-term.

Personally, I'd recommend to make your own lib with an interface like this:

```c
typedef struct {
  struct {
    void* address;
    unsigned int width,
    unsigned int height,
    unsigned int stride,
    SDL_PixelFormatEnum format;
  } buffer;

  struct {
    int x;
    int y;
  } cursor;
  
  struct {
    const void* address;
    unsigned int width;
    unsigned int height;
    unsigned int stride;
    SDL_PixelFormatEnum format;
  } font;
} SDL_FB_LOG;

#define SDL_FB_LOG_DEFAULTS {...}

void*  sdl_fb_log_allocate(SDL_FB* fb);
sdl_fb_log_printf(SDL_FB* fb, const char* fmt, ...);
```

..so basically a glorified `memcpy` to move stuff around. It will be more stable than hal/debug, more flexible, potentially faster etc.
Most importantly: You can blit the resulting framebuffer with SDL2, XGU or PVIDEO, and show and hide the terminal whenever you want (unlike hal/debug which overwrites your memory which quickly corrupts everything).

A custom lib can also be platform independent, so you can avoid adding it to nxdk, which should always be preferred for random libs that don't need explicit Xbox support.
The only thing that *would* require Xbox support for a lib like this would probably be cache flushing, but I assume debug.h also won't do it properly. If you present through SDL2 or similar, then you don't have to worry about it either.

Also writing these functions will take like.. 30 minutes? Bonus points for adding some sort of color mapping.
You could also steal similar libs from some other platforms standard library (like PSP, Gamecube, 3DS, ...).

I assume the samples will eventually use an SDL2 UI framework of some sort.

lib/hal/font_sasteroids2.h
thrimbor, line 6: This looks strange to me (the attribute being in the middle of a definition, instead of pre- or post-fix), are you sure it's syntactically correct?

lib/hal/debug.h
thrimbor: Personally, I'm very torn on this one. On one hand, it isn't good to have to carry this code around, and we risk people relying on it for too much, but on the other hand, it is *the* most used debugging tool for me, and probably others as well.
I read something a while back that described how on the PS3, stdout and stderr are essentially tunneled over the network, and output on the developer's system. This is something I had in mind for nxdm (another thing I can't seem to find the time for), along with intercepting `DbgPrint` and friends. This would at least reduce the need for `debugPrint` (together with providing a gdb stub), but even then, I see myself needing the opportunity to get something output without adding a whole pile of complexity quite often.

Are there any standard solutions we could add instead? I would like to not recommend users to roll their own, I suspect it would just lead to people copying old versions of `debugPrint` from project to project. Some lib maybe that gives you relatively simple text output capabilities, but can be pointed to an arbitrary framebuffer (so it could be adapted to SDL or PVIDEO trivially)? Not to mention that we currently *also* have `pb_print`...

JayFoxRox: - `pb_print` is an absolute nightmare to me; it wastes RAM, is very slow and works against the architecture (it does per-"pixel" drawing on the GPU, so it sends thousands of individual commands). Overall, I hope that pbkit will get major refactors (slimming down) in the future. It's easy enough to set up PVIDEO (needs a lib / code) or to send a draw call for a surface which contains the framebuffer graphic.
- I'd like to see nxdk split into the kernel-space section (CXX / C support + low-level drivers) and the user-space facing APIs (winapi, SDL, ...). kernel-space should never rely on anything in user-space.
- I'd like to not rely on `SDL` or `winapi` in kernel-space, so printing with those isn't an option.
- I don't want large font graphics as part of the kernel-space section, so it can't do visual drawing or it must communicate to optional parts (&rarr; `DbgPrint`, `std::cerr`, `fprintf(stderr, ...)` are probably good standard APIs).
- There should probably be:
    - Optional `libnxdk_print_screen.lib` which catches some of that and renders them to some framebuffer
    - Optional `libnxdk_print_usb.lib` which would send to CDC after #438 
    - Optional `libnxdk_print_network.lib` which sends data via network somehow
    - These don't have to be compatible with each other (in my opinion), but if you provide none, the app should still work.
- That `libnxdk_print_screen.lib` would use another independent, simple, drawing lib + `DbgPrint` hooking lib, so it's also possible for people to diverge from `libnxdk_print_screen.lib` without writing a lot of code.
- An additional `libnxdk_initialize_screen.lib` could be used to also set up XVideo, for the laziest ports (similar to the `automount_d`). Standard C console apps should work out of the box then.
- `debugPrint` should probably be replaced by `DbgPrint` (kernel-space like-code) or `OutputDebugString` (samples which are equivalent to user-space level); in some instances `SDL_Log` might also be an option, which can be re-routed to `OutputDebugString`. `printf` should probably be routed to `DbgPrint`?
- The use of internal `DbgPrint` should be `#if` guarded, so we can compile it away for performance builds / avoid spamming the users production build about internal events. Spurious printing (= something from nxdk prints, even if you didn't request it) has always been an annoying issue with nxdk for me; with debug.h it can easily corrupt your application memory, but it's also annoying in general.

MasonT8198: `Pb_Print` is definitely something I would agree with being a nightmare, especially on a lower end console, I could see it not being a big deal if it was being utilized on a newer console generation, however this is just personal opinion, not to be taken as logistical advice

JayFoxRox: @MasonT8198 there are enough options for more complicated UIs. This discussion focuses on `debugPrint` which is often used for CLI-like applications and simple tools (like reporting status of an EEPROM dumper, which doesn't need any UI, or a unit-test or minimalistic sample for an API).

Other things XboxDev people got involved with:
- https://github.com/Tyyppi77/imgui_sdl
- https://github.com/Ryzee119/lvgl-sdl

However, the only case that imgui and other high level libraries cover, is actual applications, designed for Xbox or framebuffer based systems.
But they all need additional code and don't support:
```c
#include <stdio.h>
int main() {
  printf("Hello world from a tiny sample!");
}
```
or:
```c
void functionInSomeLibOrLowLevelDriver() {
#if DEBUG_BUILD
  debugPrint("Some note about stuff that happened near the kernel");
#endif
}
```
So it doesn't support libraries / core functionality which don't have control over UI, and it also doesn't cover the case for "very-little-code" project.
The last point is critical for nxdk adoption, because you can't expect some beginners to write a fully fledged UI (not even with toolkits like imgui).
Even experienced developers like thrimbor, ryzee, or I will want a simple option for quick PoCs. [Most of my small projects eventually went into xbox-tools](https://github.com/JayFoxRox/xbox-tools/tree/master/python-scripts) / xboxpy, because I couldn't even be bothered to write C boilerplate or to transfer my code manually - my threshold for "too much boilerplate work" is *that* low.

MasonT8198: Thank you for the Clarification and Input, I rather appreciate it (I’m rather new to the scene so getting input and stuff is helpful for me learning all this) 

> This discussion focuses on debugPrint which is often used for CLI-like applications and simple tools (like reporting status of an EEPROM dumper, which doesn't need any UI, or a unit-test or minimalistic sample for an API).

I do agree that passing `debugPrint` as an argument called `dbgPrint` or `XdbgPrint` (This one is just attributing to the X theme) would be better as it is rather more associated to kernel-level abbreviation styles.

> But they all need additional code and don't support:

#include <stdio.h>
int main() {
  printf("Hello world from a tiny sample!");
}
or:

void functionInSomeLibOrLowLevelDriver() {
#if DEBUG_BUILD
  debugPrint("Some note about stuff that happened near the kernel");
#endif
}

I had overlooked this in the moment of writing the previous statement, I apologize. I do agree SDL2 UI is rather overkill for something like this.

lib/hal/audio.h
```diff
@@ -34,9 +34,13 @@ typedef struct
...
 void XAudioInit(int sampleSizeInBits, int numChannels, XAudioCallback callback, void *data);
+__attribute__((deprecated))
 void XAudioPlay();
+__attribute__((deprecated))
 void XAudioPause();
+__attribute__((deprecated))
 void XAudioProvideSamples(unsigned char *buffer, unsigned short bufferLength, int isFinal);
```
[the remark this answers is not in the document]
JayFoxRox: *(I'm sure I already left a comment about this *somewhere*, but I can't seem to find it anymore?)*

---

Yes, the API is deprecated, but the implementation isn't.
It depends on how you look at the deprecation attribute.

I think there should be a guard in the beginning, like this:
*(Macro name and message should probably be refined)*
```c
#ifndef INCLUDED_FROM_SDL2
#warning XAudio API is deprecated. Use SDL2 audio instead
#endif
```

*(same thing could also be done for deprecated function attributes, not sure what's better / worse)*

lib/hal/debug.h
JayFoxRox: More clarification, so it's not just me and thrimbor who are on the same page:
- [`DbgPrint` is an existing export from the Xbox Kernel, and triggers an interrupt that's meant for the debug service (so a kernel debugger would catch and display it)](https://github.com/XboxDev/nxdk/blob/d47653bea0d018e90b244ab7027b36f476db1735/lib/xboxkrnl/xboxkrnl.h#L4298-L4308) (fmt string supported).
- [`OutputDebugString` is an existing winapi function which is basically doing the same as `DbgPrint`](https://github.com/XboxDev/nxdk/blob/d47653bea0d018e90b244ab7027b36f476db1735/lib/winapi/debug.c) (no fmt string supported).

There's an existing proof of concept by thrimbor, which registers a handler for the debug service / interrupt (also the XDK would do it): https://github.com/XboxDev/nxdk/issues/172#issuecomment-821411999

Because those APIs have to exist anyway (also `printf` and other C friends), it makes sense to use them.

lib/hal/font_sasteroids2.h
Teufelchen1: No I'm not sure. I especially noted that unclearness in the PR description. 
> As I wasn't sure how to deprecate a font, I just added the deprecated attribute to the top of the file font_sasteroids2.h

I hoped you might have a better idea :p

lib/hal/debug.h
MasonT8198: Got it, makes more sense to me now

lib/hal/font_sasteroids2.h
JayFoxRox: Good catch! I missed this.

This is definitely incorrect as the preprocessor runs independently.
So if you did
```c
#include <hal/font_sasteroids2.h>
void shinyNewFunction() {}
```
it would deprecate your `shinyNewFunction`

---

Just add a `#warning` or we should skip deprecation and remove it directly.
- There'll only be a slight visual change when switching to the new font, so it's an acceptable trade-off.
- Nothing prevents users to copy the file into their own project if they require old behaviour.
- Does anybody even use this file? Especially with unclear legal state it's unlikely.
- Migrating to another file is as easy as renaming the include (which is less work, than us bikeshedding over it).

Teufelchen1: Resolved by removing `font_sasteroids2.h`. Not sure if having multiple(or so many) fonts should be part of "hal" anyway, imo.

Teufelchen1:
This PR can probably not advance any further without an agreement on the replacement for `hal/debugPrint()` (if any). 
We can either deprecate everything else first and exclude `debugPrint()` from this PR or just wait until this PR is ripen.
JFR also raised the question referring to my "deprecated as of 05/2021":
> Also not sure if the message should contain the date like that.

I added that date with the thought that I would enable people to not only search for a fix(when encountering that message) at the place of github/nxdk but also in time(git history etc.). 🤷‍♀️ Maybe I over thought it a little. Anyway, we need an agreement on that message as well. 

I resolved some conversations, against the usual workflow, prematurely as I found the topic of discussion mostly unrelated to this PR and the discussion should have been moved to the referenced issue instead(imo). So to keep things clear and productive I resolved them. 

My second(and maybe following) commit will be squashed together with the first one when the time is right, for now they will stay separated to keep clean about what is decided and implemented - and what is still on debate.

lib/hal/debug.h
JayFoxRox: Unresolved, because the question about clang docs remain: is `replacement` meant as an auto-replacement hint that must be machine readable?

lib/hal/font_sasteroids2.h
JayFoxRox: Unresolved, because there's no comment saying how this was resolved. We should not resolve discussions unless there's also a comment which concludes the discussion.

- If you changed the code to resolve this, mention what change you did.
- If you think the comment doesn't deserve further discussion (why not?) then state your reasons and let the original commenter decide if they consider it resolved.

JayFoxRox: Resolved, as per @Teufelchen1 comment in https://github.com/XboxDev/nxdk/pull/458#discussion_r618488929:

> Resolved by removing font_sasteroids2.h

*(Please make sure each discussion has a copy of such information when resolving them, otherwise it's very hard to figure out what happened)*

JayFoxRox:
> I resolved some conversations, against the usual workflow, prematurely as I found the topic of discussion mostly unrelated to this PR and the discussion should have been moved to the referenced issue instead(imo). So to keep things clear and productive I resolved them.

I unresolved the `debugPrint` discussion:

- Even though the long discussion might not affect this PR, the knowledge contained in the discussion (which involved multiple people taking time to share their opinion) should not go to waste. The essence of that discussion should be moved elsewhere.
    Until it's moved, it's not resolved.
- The main point that `debugPrint` should not be deprecated was not decided.
    The unresolved discussion is an important marker to be reminded that there's some TODO.

I also unresolved some others and left comments as to why I consider them unresolved.

> This PR can probably not advance any further without an agreement on the replacement for hal/debugPrint() (if any).
> We can either deprecate everything else first and exclude debugPrint() from this PR or just wait until this PR is ripen.

My opinion, to avoid any bikeshed: Remove deprecation of `debugPrint`.
Clearly this is something where people have different opinions, so it's best to avoid it for now.

Because it's so heated, also try to start a new discussion how it *could* be deprecated (= create an issue about a new visual printing API for console like applications; which seems to be the main point that is not covered by #172).

> JFR also raised the question referring to my "deprecated as of 05/2021":

I'd remove it for consistency with how we deprecated things in the past.

In fact, in the past we didn't even add messages: https://github.com/XboxDev/nxdk/commit/8e5a12fe49ef6c1f6ede6f26f8797b35c5940fd3, so I suggest to also remove message strings and the `replacement` parameter which will also resolve https://github.com/XboxDev/nxdk/pull/458#discussion_r614982007 without spending more time on it.
Replacements could be mentioned in a code comment instead, but people should be able to find this discussion via GitHub or google.

Most replacements are also common-sense - the reason we deprecate is to give people time to avoid bad code, not necessarily to teach them how to resolve it.

We should not over-engineer this. A stable solution (like solution we had in the past) is preferred over using shiny new toolchain features (like deprecation `message` + `replacement`, which might trip IDEs when used incorrectly / when it expects GNU code instead of clang code). [KISS](https://en.wikipedia.org/wiki/KISS_principle).

thrimbor:
This needs a rebase.

I also suggest to clean up the commits by squashing 3748c76c81ca0c60840503d7bd71c48b3f288f1e and d5ad2062784ad6640fa620f9280e2e53ba60b78b into 98981f281fb998b415b922b67fe8d7bbb8c1db5e, and removing the change in `lib/hal/font_sasteroids2.h` from the latter since it's being deleted in the next commit anyway.

Teufelchen1:
I would like to but I still wait for your feedback [here](https://github.com/XboxDev/nxdk/pull/458#discussion_r618573356).

lib/hal/audio.h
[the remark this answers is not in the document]
thrimbor: Oops, managed to miss this comment.
I think it's best to remove the deprecation attribute for now, and just add a comment in the audio header stating that this is only supposed to be used as a backend and that something like SDL2 should be preferred for applications.

lib/hal/audio.h
```diff
@@ -31,7 +31,9 @@ typedef struct
 	int                      numChannels;
 } AC97_DEVICE  __attribute__ ((aligned (8)));
 
-// note that I currently ignore sampleSizeInBits and numChannels.  They
+// The XAudio API is only supposed to be used as a backend. Using SDL2 for 
+// audio playback should be prefered for applications.
```
thrimbor, line 35: nit: "preferred"

lib/hal/audio.h
```diff
@@ -31,7 +31,9 @@ typedef struct
...
+// note that currently sampleSizeInBits and numChannels are ignored. They
```
thrimbor, line 36: nit: Capitalize "note"

thrimbor (approved):

lib/hal/audio.h
JayFoxRox: nit: "They" at end of line is equally unfortunate (I did not check how the comment continues, but this looks odd).

Also not sure if we do this in nxdk, but I personally also usually use markdown for code stuff in comments, so I'd use:

```
[..] currently `sampleSizeInBits` and [..]
```

lib/hal/audio.h
Teufelchen1: Applied.

lib/hal/audio.h
Teufelchen1: Applied both nits and the markdown-code-comment idea.

PR #461, opened 2021-04-19
Make hal includes relative to lib/ folder

Ryzee119:
Apart from the inconsistencies around the code base this has caught me out a few times.

If a user applications has a header file that shares the same name as the hal libs and is in a directory included by `-I` it will include the nxdk's header instead of the application. This is quite common as the hal libs use really common basic filenames.

i.e, if the user application has:
```
inc/debug.h
inc/input.h
```
instances of `#include <debug.h>` or `#include <input.h>` in the user application will include nxdk's headers instead of the users potentially causing some not immediately obvious errors. Using `"` or `< >` makes no difference in this instance.

This was mentioned here:
https://github.com/XboxDev/nxdk/pull/457#discussion_r615170326

This PR makes it so all nxdk hal includes must have <hal/xxxx.h> prefix. This should make the above occurances very rare.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,6 @@ TOOLS        = cxbe vp20compiler fp20compiler extract-xiso
 NXDK_CFLAGS  = -target i386-pc-win32 -march=pentium3 \
                -ffreestanding -nostdlib -fno-builtin \
                -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt/libc_extensions \
-               -I$(NXDK_DIR)/lib/hal \
                -isystem $(NXDK_DIR)/lib/pdclib/include \
                -I$(NXDK_DIR)/lib/pdclib/platform/xbox/include \
                -I$(NXDK_DIR)/lib/winapi \
--- a/lib/net/nforceif/include/arch/cc.h
+++ b/lib/net/nforceif/include/arch/cc.h
@@ -36,7 +36,7 @@
 
 /* Include some files for defining library routines */
 #include <string.h>
-#include <debug.h>
+#include <hal/debug.h>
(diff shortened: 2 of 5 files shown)


JayFoxRox:
This might be incomplete. Did you verify submodules?

I have some notes about submodules in my old version of something like this: https://github.com/JayFoxRox/nxdk/pull/61, but those have been fixed:

- `sdl/SDL2/src/video/xbox/SDL_xbframebuffer.c:#include <debug.h>` was removed in https://github.com/XboxDev/nxdk-sdl/commit/ecf5136df14a71e517952860903e0eb195380a09ae09148173d103a3e4e6d3d29ded65139dc47c6f9ac5715e340fa7d3061dd6db
- `pdclib/platform/xbox/functions/assert/assert.c:#include <debug.h>` was fixed in https://github.com/XboxDev/nxdk-pdclib/commit/598c7be2d5445ac2c16d00242c6f4dbaf9fe0975

However, there might be new problems which have been created since then, I'm also not sure if I checked for `audio.h`, `video.h`, `input.h` etc.

JayFoxRox:
Please fix order of commits: fix bugs first, then remove the include. Current order makes it impossible to run `git bisect`.
Alternatively squash those commits into a single one (removing the commit description for consistency with our other commits).

Ryzee119:
> Please fix order of commits: fix bugs first, then remove the include. Current order makes it impossible to run `git bisect`.
> Alternatively fixup/squash those commits into a single one (removing the long commit description for consistency with our other commits).
Commits have been reordered/shortened

JayFoxRox:
Thanks!

Merging this right now. If this breaks anyones project, it's easy enough to just change the include paths.

PR #462, opened 2021-04-19
winapi: Don't rename .edata section to .edataxb

thrimbor:
This lets the `.edata` section, which stores the export directory, keep its original name, instead of getting renamed to `.edataxb`.

Also see #419.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ NXDK_ASFLAGS = -target i386-pc-win32 -march=pentium3 \
                -nostdlib -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt
 NXDK_CXXFLAGS = -I$(NXDK_DIR)/lib/libcxx/include $(NXDK_CFLAGS) -fno-exceptions
 NXDK_LDFLAGS = -subsystem:windows -fixed -base:0x00010000 \
-               -stack:65536 -merge:.edata=.edataxb
+               -stack:65536 -merge:.edata=.edata
 
 ifeq ($(DEBUG),y)
 NXDK_ASFLAGS += -g -gdwarf-4
--- a/lib/winapi/libloaderapi.c
+++ b/lib/winapi/libloaderapi.c
@@ -31,13 +31,13 @@ BOOL FreeLibrary (HMODULE hLibModule)
     return TRUE;
 }
 
-static PIMAGE_EXPORT_DIRECTORY find_edataxb (void)
+static PIMAGE_EXPORT_DIRECTORY find_edata (void)
 {
     DWORD num_sections = CURRENT_XBE_HEADER->NumberOfSections;
     PXBE_SECTION_HEADER section_header_addr = CURRENT_XBE_HEADER->PointerToSectionTable;
 
     for (DWORD i = 0; i < num_sections; i++) {
-        if (memcmp(section_header_addr[i].SectionName, ".edataxb", 8) == 0) {
+        if (strcmp(section_header_addr[i].SectionName, ".edata") == 0) {
             return (PIMAGE_EXPORT_DIRECTORY)section_header_addr[i].VirtualAddress;
         }
     }
@@ -50,7 +50,7 @@ FARPROC GetProcAddress (HMODULE hModule, LPCSTR lpProcName)
     if (hModule == NULL) {
         // When no dll handle is given, the symbol gets looked up in the main module
 
-        PIMAGE_EXPORT_DIRECTORY exportdir = find_edataxb();
+        PIMAGE_EXPORT_DIRECTORY exportdir = find_edata();
         if (!exportdir) {
             SetLastError(ERROR_PROC_NOT_FOUND);
             return NULL;


lib/winapi/libloaderapi.c
```diff
@@ -37,7 +37,7 @@ static PIMAGE_EXPORT_DIRECTORY find_edataxb (void)
     PXBE_SECTION_HEADER section_header_addr = CURRENT_XBE_HEADER->PointerToSectionTable;
 
     for (DWORD i = 0; i < num_sections; i++) {
-        if (memcmp(section_header_addr[i].SectionName, ".edataxb", 8) == 0) {
+        if (strncmp(section_header_addr[i].SectionName, ".edata", 8) == 0) {
```
Teufelchen1, line 40: Hm shouldn't that be a `strncmp(section_header_addr[i].SectionName, ".edata", 6)`? (Changing the size argument from 8 to 6)

JayFoxRox: Was the intention to also support the old `.edataxb` style by only checking the prefix? Why? This wasn't mentioned in the PR description; behaviour is also a bit sneaky in the code (hard to spot that this might happen).

JayFoxRox:
To be more precise than the PR description: I started this in https://github.com/XboxDev/nxdk/pull/419#discussion_r615252278

This PR existed already by the time I learned:
> `-merge:.edata=.edata` works, but only with lld. It crashes the MSVC linker. 

So I'm a bit more skeptical now. I'm slightly concerned about llvm eventually saying: "if MSVC doesn't do it, then we shouldn't do it either"; maybe it's also just a bug that lld is accepting it, or it even breaks stuff internally (and just doesn't crash).

However, I think we should still consider doing it; if stuff breaks we can probably paddle back and use cxbe to rename sections for us. `.edata` is a much more revealing name: when you search documentation for it, you'll quickly find it.

lib/winapi/libloaderapi.c
thrimbor: > Hm shouldn't that be a `strncmp(section_header_addr[i].SectionName, ".edata", 6)`? (Changing the size argument from 8 to 6)

No, that would return a false positive for everything starting with `.edata`.

> Was the intention with strncmp to also support the old .edataxb style by only checking the prefix?

The opposite, I wanted to prevent that. However, while the above `strncmp` should work as expected, it's actually nonsense to do it that way, and the initial `memcmp` was incorrect as well - I somehow had it in my mind that the section names are null-terminated 8-byte fields, but they're not, the library names are. Section names are properly null-terminated strings, so a regular `strcmp` is the correct thing to do here.
I've updated the code to reflect that now.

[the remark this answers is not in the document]
thrimbor: > The function name is misleading now.

Fixed.

> In the future, we'll likely want a function to find a generic XBE section.

True. I haven't looked into it much yet, but if there is a winapi solution for this, I'll likely prefer that (and modify this code to use it when it lands), otherwise, I'd suggest such functionality goes into the nxdk lib, and then it probably makes sense to keep this piece of code separate. For now this code only exists to not clutter `GetProcAddress` with the section search.

Makefile
```diff
@@ -64,7 +64,7 @@ NXDK_ASFLAGS = -target i386-pc-win32 -march=pentium3 \
                -nostdlib -I$(NXDK_DIR)/lib -I$(NXDK_DIR)/lib/xboxrt
 NXDK_CXXFLAGS = -I$(NXDK_DIR)/lib/libcxx/include $(NXDK_CFLAGS) -fno-exceptions
 NXDK_LDFLAGS = -subsystem:windows -fixed -base:0x00010000 \
-               -stack:65536 -safeseh:no -merge:.edata=.edataxb
+               -stack:65536 -safeseh:no -merge:.edata=.edata
```
JayFoxRox, line 67: This needs a rebase because I merged #445 

nit: Should we add a comment that this part might crashes LINK.exe (which version?) but works fine in lld (which version?)? + maybe mention why we keep this section separate (and that it's a niche use-case)? Basically just future proofing for when someone tries to run MSVC on this, or when LLVM changes behaviour.

thrimbor: The crash was in VS2017, but I just noticed that VS2019 emits an error about a circular reference instead. It seems that lld just evaluates those parameters from left to right, while LINK.exe builds a graph internally.

I experimented with a few other ways (unfortunately, there's no direct way to disable it), one way would be to modify the section flags to be incompatible with `.rdata`: You can remove the readability attribute with `/SECTION:.edata,!R`, which has this effect, but unfortunately, lld does not support this syntax yet. I tried to mark the section executable with `/SECTION:.edata,E`, this also works on LINK.exe, but lld seems to just ignore incompatible section flags for the default merge rules.

TL;DR: I'm not sure this is a good change anymore, but there also doesn't seem to be a good solution...

thrimbor: Okay, I've come up with another solution: We can increase the Xbe header size and add our own fields. I've tested it here: https://github.com/thrimbor/nxdk/commit/314b327cd3941c9b3e2a3827ffc754c99efcf156

This eliminates the need for a separate `.edata` section, but I'm not sure how other tools react when the xbe header is a different size. They might assume the xbe is corrupt.
I kinda like this solution, it removes the need for yet another linker parameter, but if we decide to go this route, we should probably add a bunch of other interesting fields, too (like an import directory).

thrimbor: I'll probably open a new PR with a cleaner and extended version of my idea in the next few days.
Closing this PR for now.

JayFoxRox: > We can increase the Xbe header size and add our own fields

This is problematic, because real XDKs also produce different XBE header sizes. MS themselves kept extending it over the years. So I'm not sure if this is a good solution either.
I believe some debug DXTs by MS will also crash if the header size isn't proper (verifierx maybe? It has broken XBE parsing in general).

thrimbor: This is the first time I'm hearing this - the last documented field in the header is `Logo Bitmap Size`, which I assume was there from the start, and I'm not aware of any fields that follow it. I did intend to run a small checking program over the XBEs of the games I own to make sure there are no variations in the header size.
As for potentially crashing MS tools - I currently don't have a way to check this, that's why I provided the branch above, to allow others to test this.
If extending the image header is indeed problematic, we may be able to put the data in other spots in the headers - the end of the certificate header maybe, or we could put it in the image header like I did above, but don't increase the header size field and instead purely rely on some magic number for detection etc.

JayFoxRox: > This is the first time I'm hearing this

Take it with a grain of salt, I might be misremembering.
Although I feel pretty confident about what I said (~ some changes to XBE image structures as time went on).

> I did intend to run a small checking program over the XBEs of the games I own to make sure there are no variations in the header size.

It probably also makes sense to write a scraper for https://github.com/Cxbx-Reloaded/game-compatibility/issues

> If extending the image header is indeed problematic, we may be able to put the data in other spots in the headers - the end of the certificate header maybe

Now that you said that, I'm not sure *which* header they extended (or if it was just one, or multiple headers that got extended).
If I remember correctly, the stuff they added was mostly for XAPI / copy-protection.

thrimbor: I wrote my small tool yesterday and let it run over all the XBEs I currently have on my system (which consist of some games and copies of the dashboards from my Xboxes). I found that the image header varies between a size of 376 and 388 bytes. Here's the list of sizes in relation to the XDK version used in the XBE:
4928: 376
5028: 384
5233: 384
5455: 388
5558: 388
5659: 388
5788: 388
5849: 388
5960: 388
5971: 388

So we should extend the header past 388 bytes for our purposes, but we need to make sure we don't put anything in 376-388 that's problematic - I haven't looked into the meaning of those bytes that much, but the few XBEs I checked out in Ghidra don't access those bytes themselves.

> Now that you said that, I'm not sure which header they extended (or if it was just one, or multiple headers that got extended).
> If I remember correctly, the stuff they added was mostly for XAPI / copy-protection.

Yeah Luke mentioned that they extended the certificate header for that copy-protection / DRM stuff.

JayFoxRox: > So we should extend the header past 388 bytes for our purposes, but we need to make sure we don't put anything in 376-388 that's problematic

I believe the DXTs on devkits will check the size of the headers to classify the XBE; so if we have a different size they might assert or crash.
So I'm not sure if it's a good idea to do it this way.

---

I also looked at the DVD dongle XBE again (which is basically a DLL for the dash that also has access to dash functions) and it seems that the dash calls the dongle XBE with a list of the dash functions (= address of export table is not explicitly stored in XBE metadata and only known once the code runs).

That might be a good idea in general, because I'd really like to be able to register virtual DLLs, too (primarily to support optional windows functions, by providing trampolines/stubs for functions that are queried through `GetProcAddress` on kernel32.dll and user32.dll, without having to ship/drop those as actual DLLs).
A virtual default.exe with proper headers could act as stub for default.xbe exports.

We could also use a footer after kernel thunk to store the export table, or use the entry point (by registering a bit of memory for the export directory, then pointing the entry at that, and just have an 8 byte region to prefix the export table `.org 0; jmp real_entry_point); .org 8; <EXPORT_TABLE>;`); we'd then know that the export table is at `entry + 8`

Regardless of how we do it: it will be dirty in some way.
Given how much of a niche feature this is, I mostly care it's not also making our binaries fragile (a DLL should not fail in the future, when our XBE standard changes).

---

Why did we want to export from the main EXE again? Did we have a specific port in mind?

Typically you'd just pass callback function pointers from XBE to the DLL, so the DLL can communicate with the XBE.
From within the XBE you also know your exports at compile-time, so there's little benefit in parsing your own export table?

Most people will probably want to import functions from DLLs to their EXE/XBE.
Importing functions from the main EXE/XBE for use in a DLL (discussed here) is hopefully extremely rare.
Even then, there could just be a helper function to mark any DLL/EXE as main module; so a user could manually create a virtual DLL/EXE (or load a real one) and then tell the dynamic linker that the DLL/EXE with proper headers is the main module (instead of non-export capable default.xbe or the export capable xboxkrnl.exe).

So the bigger problem is probably `dllimport` which requires a stub; the existing XBE field for non-kernel imports is already used by debug kernels, so re-using it is probably a bad idea.

PR #475, opened 2021-04-22
Fix strict prototype warnings

### LINKED ISSUE #468: Warnings about missing `(void)` on function prototypes
We have some broken includes where must add `void`; see https://stackoverflow.com/a/47693

Namely:
```
nxdk/lib/winapi/windef.h:18:34: warning: this function declaration is not a prototype [-Wstrict-prototypes]
typedef int (FAR WINAPI *FARPROC)();
```
```
nxdk/lib/winapi/processthreadsapi.h:24:16: warning: this function declaration is not a prototype [-Wstrict-prototypes]
DWORD TlsAlloc ();
```

thrimbor:
Fixes #468.

Fixes a bunch of C function declarations/definitions that were using the old syntax `()`, but should've used `(void)` instead. The only part of code in the nxdk repo that now triggers a `-Wstrict-prototypes` warning is the (old) USB stack.

Also fixes a call where `XAudioPause` was incorrectly called with a parameter.

### DIFF
--- a/lib/hal/audio.c
+++ b/lib/hal/audio.c
@@ -115,7 +115,7 @@ static BOOLEAN __stdcall ISR(PKINTERRUPT Interrupt, PVOID ServiceContext)
 }
 
 
-void XDumpAudioStatus()
+void XDumpAudioStatus(void)
 {
 	volatile AC97_DEVICE *pac97device = &ac97Device;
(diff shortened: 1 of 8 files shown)


JayFoxRox:
That was *very* quick. Thanks.

PR #478, opened 2021-04-26
xboxkrnl: Add volatile keyword to atomic operation functions

### LINKED ISSUE #476: `volatile` keyword missing on atomic operation arguments in xboxkrnl.h
Around here (might include some which don't need the keyword, and might exclude some which do require it):

https://github.com/XboxDev/nxdk/blob/62a99875bdc9498c00fc8e276f112dba528add7f/lib/xboxkrnl/xboxkrnl.h#L4426-L4468

MSDN documents this, and it will actually trigger compile warnings / errors with real-world applications which attempt to use these functions.

JayFoxRox:
Closes #476 

I took some of MSDN as reference, but also https://doxygen.reactos.org/d0/d35/ntoskrnl_2ex_2interlocked_8c_source.html for some guidance.

I didn't check for completeness or correctness - but this is enough to get some code to compile that I had.

Please review before merge + feel free to fixup in my branch (I'm not sure when I'll find time to integrate any review feedback myself).

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -4221,7 +4221,7 @@ XBAPI OBJECT_TYPE ExMutantObjectType;
 
 XBAPI LARGE_INTEGER NTAPI ExInterlockedAddLargeInteger
 (
-    IN PLARGE_INTEGER Addend,
+    IN OUT PLARGE_INTEGER Addend,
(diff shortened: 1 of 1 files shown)


thrimbor:
I checked it against the msdn, and checked that there are no other Interlocked*-functions in xboxkrnl.h that should have a volatile parameter.
Looks good, merging.

PR #481, opened 2021-05-04
pktdrv: Silence network init

dracc:
Having link speed and EEPROM MAC address debugPrint'ed on startup is not always desirable.

Setting `NXDK_NET_SILENT` to `y` in your Makefile removes this clutter.

### DIFF
--- a/lib/net/pktdrv/pktdrv.c
+++ b/lib/net/pktdrv/pktdrv.c
@@ -34,8 +34,6 @@
 
 #include "pktdrv.h"
 
-#define DISPLAYMSG
-
 //Defines number of Rx & Tx descriptors, and number of buffers -ring- for received pkts
 #define NBBUFF	32
 


MasonT8198:
457BAEBF-E463-4F2E-818C-037C42F8F984 example

Ryzee119:
Gave this a quick test, and works as expected with `NXDK_NET_SILENT = y` in makefile.

thrimbor:
Hm, I think I'd prefer to just have the `#define` removed, without a new build flag. If anyone really wants to enable it, they can just add the definition to their CFLAGS themselves, and I'd like to keep the amount of build options low (especially since we want to move away from this kind of build system).

Teufelchen1:
I'm more with thrimbor on this. Beside: What ever behaviour is chosen, please don't forget to add this info to the Wiki. i.e. [here](https://github.com/XboxDev/nxdk/wiki/Libraries-and-advanced-build-configuration).

dracc:
PR updated to reflect solution presented by Thrimbor.

Makefile
```diff
@@ -82,7 +82,7 @@ include $(NXDK_DIR)/lib/Makefile
 OBJS = $(addsuffix .obj, $(basename $(SRCS)))
 
 ifneq ($(NXDK_CXX),)
-include $(NXDK_DIR)/lib/libcxx/Makefile.nxdk
+include $(NXDK_DIR)/lib/libcxx/Makefile.nxdk
```
Teufelchen1, line 85: would you mind removing that whitespace?

dracc: I sure would not, it was never supposed to be there in the first place.

thrimbor:
Can you update the commit title to reflect that change?

PR #488, opened 2021-05-31
Docker improvements

thrimbor:
This includes a few changes to improve our Docker images:
* Update the base from Alpine 3.12 to 3.13
* Use a multi-stage Dockerfile to reduce the size of the final image by not including unnecessary Alpine packages
* Prebuild the libraries and tools in the images
* Don't include the samples in the image
* Provide and test three images: nxdk, which is built with -02, nxdk-debug, which is built with debug info, and nxdk-lto, which is built with -O2 and LTO enabled
* Use a build matrix for the Docker images to keep the total runtime of the Actions low (was about 10m without the matrix)
* Add some metadata, like the build time, licenses, repository URL, git revision etc.

a7eee3906108b543c2378bd6e904c4630470a6d2 is a bit of a band-aid for our build system. I'm not too happy about it, but I couldn't integrate the -flto parameter like I did with -02 because Docker has some weird issues with whitespaces in build arguments.

At first I planned to integrate the Dockerfile from https://github.com/XboxDev/nxdk-buildbase into this one, but unfortunately the hash for an image layer takes the timestamp from when the file was last changed into account, so installing the Alpine packages in this Dockerfile would have caused unnecessarily large downloads for users updating the nxdk image. So instead, I updated https://github.com/XboxDev/nxdk-buildbase to provide two images, one with all the packages required to build nxdk itself and the tools, and one smaller one that only includes the packages required to build code with the nxdk image.

86803a10636f50da5ac984772302dd877f6b8fc3 is necessary to be able to build the samples with the Docker image. #446 had the same change for the other samples.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@ tools/cg/mac
 tools/cg/win
 lib/sdl/SDL2/test
 lib/libcxx/test
+samples/
--- a/.github/workflows/build_docker_image.yml
+++ b/.github/workflows/build_docker_image.yml
@@ -25,6 +25,9 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: recursive
+      - name: Generate timestamp
+        id: timestamp
+        run: echo ::set-output name=timestamp::$(date --rfc-3339=seconds --utc)
       - name: Generate tags
         id: tags
         run: |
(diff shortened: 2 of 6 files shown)


MasonT8198:
Did a read through on this PR it’s looks good to me! Hopefully this gets looked over. Good Work!

Ryzee119:
Ive tested a few of the samples build from docker. Im happy with the changes and samples build/work ok.

Im not too familiar with docker, but if there's no objections i'll merge in a couple days.

PR #489, opened 2021-06-05
Format string checking

thrimbor:
This enables format string checking for `debugPrint()`. This means that errors like this can be caught by the compiler and a warning will be issued:
```
debugPrint("i: %d\n"); // ,i is missing
```

I (and others) have been bitten by this kind of error a few times already while debugging.

Format string warnings caused by the code in the nxdk repo are fixed by the other commits, but there are remaining ones in the submodules.


Ryzee119:
Looks good. Ill merge

PR #490, opened 2021-06-07
Added Sample:Example of Menu with Controller Support

HodgkinsDev:
This is a new sample for the NXDK which is an example of how you would create a menu with controller support. It's easily editable, and commented so others can use this as a base or reference for there own projects!


thrimbor:
Hi, thanks for the contribution, but unfortunately I don't consider this code to be a good fit for the nxdk samples. The code for a sample should be "in a nutshell", anything that goes beyond demonstrating the correct use of the basics is out of scope.
There are also several issues with the code, among others concerning the incorrect use of `auto`, the commit title, general code structure, and the formatting.

PR #491, opened 2021-06-07
Use more kernel exports where appropriate

GXTX:
Test it, don't merge without testing, without warranty, without testing beyond a single sample. Could've been added in 2 more places but it didn't feel right.

### DIFF
--- a/lib/hal/debug.c
+++ b/lib/hal/debug.c
@@ -216,8 +216,8 @@ void debugAdvanceScreen( void )
 	
 	unsigned char* thisScreen = SCREEN_FB + (SCREEN_WIDTH * MARGIN)  * pixelSize;
 	unsigned char* prevScreen = thisScreen+lineSize;
-		
-	memmove(thisScreen, prevScreen, screenSize);
+
+	RtlMoveMemory(thisScreen, prevScreen, screenSize);
 
 	nextRow -= (FONT_HEIGHT+1);
 	nextCol  = MARGIN; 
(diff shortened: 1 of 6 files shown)


JayFoxRox:
No; prefer `memset`.

Reasons:
- `RtlZeroMemory` is an utterly useless MS function which should only be used for style-consistency in MS C++ code.
- `RtlZeroMemory` adds dependency on kernel lib, which shouldn't be necessary.
- `memset` is part of the C standard and can be handled even by the compiler (where it also outperforms `RtlZeroMemory`.. a lot).
- Commit titles are double-prefixed and don't follow our usual style ("prefix: Verb stuff").

GXTX:
Then the question is why are other kernel exports used? MmAllocCongiuous is used often in the codebase, why not malloc? Etc etc. Feel free to close if not merged as is.

LoveMHz:
> Then the question is why are other kernel exports used? MmAllocCongiuous is used often in the codebase, why not malloc? Etc etc. Feel free to close if not merged as is.
> 
> Imo we should be using the export for memmove as well.

Because other exports may have other implications based on kernel/hardware. These concerns should really be raised in another issue though. As each of those should be addressed separately. 

But back to this issue, I have to agree with @JayFoxRox on this. ``RtlZeroMemory`` is garbage compared to memset.

GXTX:
Added a couple of more exports, again they could be used more in some other areas of the tool chain however the areas they're used are what I would consider 3rd party.

thrimbor:
Replacing `memset` with `RtlZeroMemory` is not advisable for the reasons mentioned above, the same goes for `RtlMoveMemory`.
Replacing `malloc` with `MmAllocateContiguousMemory` is a bug, those functions are not interchangeable and do not implement the same functionality. The change has a high chance of crashing any application running that code, which is especially critical due to the functionality the code implements.

PR #492, opened 2021-06-07
Use less kernel exports where appropriate

GXTX:
As discussed in https://github.com/XboxDev/nxdk/issues/491#issuecomment-856235343 the tool chain currently uses many kernel exports which are unable to be properly optimized by the compiler. We should strive to move away from such exports, example Nt class exports.

Another example although unused in the tool chain itself is `XboxHardwareInfo` which is a struct containing information from known memory locations (but the compiler doesn't know this). We should ideally offer a way for users to get this information without relying on the kernel (or having to know the exact memory location of the values).

### DIFF
--- a/lib/nxdk/format.c
+++ b/lib/nxdk/format.c
@@ -84,7 +84,7 @@ bool nxFormatVolume (const char *volumePath, uint32_t bytesPerCluster)
         return false;
     }
 
-    RtlFillMemory(buffer, alignment, 0xFF);
+    memset(buffer, 0xFF, alignment);
     FATX_SUPERBLOCK *superblock = (FATX_SUPERBLOCK *)buffer;
     superblock->Signature = FATX_SIGNATURE;
     superblock->SectorsPerCluster = bytesPerCluster / diskGeometry.BytesPerSector;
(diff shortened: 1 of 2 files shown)


JayFoxRox:
The commit title is not very good; "less exports" could also mean that this removes exports from the kernel export table (which it doesn't do).
Please make the commit title more descriptive (but keep it short) and ensure it starts in uppercase after the prefix.

lib/nxdk/format.c
```diff
@@ -84,7 +84,7 @@ bool nxFormatVolume (const char *volumePath, uint32_t bytesPerCluster)
         return false;
     }
 
-    RtlFillMemory(buffer, alignment, 0xFF);
+    memset(buffer, 0xFF, alignment);
```
JayFoxRox, line 87: nit: One of the few cases where `RtlFillMemory` and friends might be useful, because this is essentially a kernel driver.
When working on a kernel driver, one might not have access to a full libc and the kernel is definitely available, too (so there's no risk of adding an unnecessary dependency).

However, at the moment this also uses `malloc` and friends, so it doesn't really matter.
For now, I prefer `memset`, but I wouldn't be surprised if we change it back eventually (when turning this into a low-level driver).

MasonT8198:
I did review and check if this PR worked with some existing NXDK made applications, e.g. NEvolutionX, Extract-Xiso, and XGU. It did without any issues, this may not be representative for all NXDK made applications however. Just reporting my test results

Ryzee119:
Fairly straight forward change. Ill merge this.
Thanks for testing MasonT8198.

GXTX:
Thanks, hopefully we can get some benchmarks between `RtlZeroMemory` and `memset`.

PR #494, opened 2021-06-19
winapi: Fix XBE file header

### LINKED ISSUE #493: XBE header magic too small
https://github.com/XboxDev/nxdk/blob/a7a5ecff1d5b6b6af187c4de013ddf26509851b5/lib/winapi/winnt.h#L286

The value of this one is `0x48454258` on retail titles, a full DWORD long, not just WORD.

dracc:
Closes #493

### DIFF
--- a/lib/winapi/winnt.h
+++ b/lib/winapi/winnt.h
@@ -283,7 +283,7 @@ typedef struct _XBE_SECTION_HEADER
 
 typedef struct _XBE_FILE_HEADER
 {
-    WORD Magic;
+    DWORD Magic;
     UCHAR Signature[256];
     DWORD ImageBase;
     DWORD SizeOfHeaders;


lib/winapi/winnt.h
```diff
@@ -283,8 +283,8 @@ typedef struct _XBE_SECTION_HEADER
 
 typedef struct _XBE_FILE_HEADER
 {
-    WORD Magic;
-    UCHAR Signature[256];
+    DWORD Magic;
+    UCHAR Signature[255];
```
thrimbor, line 287: This is not correct, the signature is 256 bytes, not 255. You can compare with cxbe and xboxdevwiki.

thrimbor (commented):
Good catch on the typo. Can't believe I missed that when testing (got screwed by padding I guess).

lib/winapi/winnt.h
dracc: My testing suggested this was correct. Thrimbor tested better, showing the opposite and that my testing simply gave false positives.

thrimbor:
Looks good, thanks for fixing!

PR #497, opened 2021-08-24
winapi: Run fls_init in startup thread, before regular constructors

### LINKED ISSUE #496: XBE crashes on start, when declaration added in class.
This issue was discussed in the discord channel, and in fact there is a workaround adding "-fno-threadsafe-statics" to the compiler flags (thanks to Thrimbor for that hint!!): it seems that adding certain declarations in a class, makes the program crash on start (even if that code is not executed). I built a minimal sample that has this behaviour and uploaded it here:

https://gist.github.com/hcfcoder/7d7cde1f55000d160084869f1c2fd141

In that sample, this declaration in the "Logger" class makes the program crash on boot:

std::stringstream m_str;

It crashes if we put it in the public part of the class, and also if we put it in the private part of the class too. On the other hand, putting that same declaration in the main function of the program, does not crash.

thrimbor:
This, together with https://github.com/XboxDev/nxdk-pdclib/pull/48, fixes the issue where constructors were run before thread-local storage was available by running them from the main thread instead of the startup thread. However, `fls_init` is a special constructor that needs to run before any thread creation, so this introduces and uses a new mechanism which allows for constructors to run in the startup thread instead by putting function pointers in a custom subsection of `.CRT` (the same thing is done automatically by the compiler for regular constructors).

This code can be used to verify that the change works. Before the changes, this would crash during the construction of the global `std::stringstream s`, but with the changes, it should run correctly.

```c++
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>
#include <sstream>

std::stringstream s;

int main(void) {

  debugPrint("hello\n");

  while (true) {
    Sleep(2000);
  }

  return 0;
}
```

Fixes #496.

This will be combined with a submodule update and turned into a normal PR when https://github.com/XboxDev/nxdk-pdclib/pull/48 is merged.

### DIFF
--- a/lib/nxdk/automount_d.c
+++ b/lib/nxdk/automount_d.c
@@ -11,10 +11,10 @@
 #include <windows.h>
 #include <xboxkrnl/xboxkrnl.h>
 
-__attribute__((constructor)) void automount_d_drive (void)
+__cdecl int automount_d_drive (void)
 {
     if (nxIsDriveMounted('D')) {
-        return;
+        return 0;
     }
 
     // D: doesn't exist yet, so we create it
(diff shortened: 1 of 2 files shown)


thrimbor:
This has sat for a while, and as it is a crash fix, I'm self-merging this now.

PR #499, opened 2021-09-11
winapi: Implement CopyFileA

abaire:
Implements CopyFileA. 

Tests can be found at abaire/nxdk_tests


lib/winapi/winioctl.h
```diff
@@ -0,0 +1,18 @@
...
+typedef struct _FILE_OBJECTID_BUFFER {
+  BYTE  ObjectId[16];
+  union {
+    struct {
+      BYTE BirthVolumeId[16];
+      BYTE BirthObjectId[16];
+      BYTE DomainId[16];
+    } DUMMYSTRUCTNAME;
+    BYTE ExtendedInfo[48];
+  } DUMMYUNIONNAME;
```
dracc, line 15: Nit: Consider using a "real" name.

lib/winapi/filemanip.c
```diff
@@ -241,6 +242,141 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+            return FALSE;
+        }
+
+        if (!bytesRead)
+        {
+            break;
+        }
+
+        while (bytesRead > 0)
+        {
```
dracc, line 344: Please follow the code style of the file, i.e. `{` on the same line as `)` for `if` and `while`.

lib/winapi/winioctl.h
abaire: I used the names from the (public domain) header in mingw32 and am not sure that I have an idea for more descriptive names. Happy to change if you have a suggestion though.

For reference: https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-w64-mingw32-4.8/+/refs/heads/studio-1.0-release/x86_64-w64-mingw32/include/winioctl.h#1641

lib/winapi/filemanip.c
abaire: Fixed, sorry. 

I even had to fight my autoformat to do it incorrectly, not sure what I was looking at for context that made me do it that way :)

lib/winapi/winioctl.h
dracc: Fair enough, I didn't realize you had a reason for the dummy names. :)

lib/winapi/winioctl.h
```diff
@@ -0,0 +1,18 @@
...
+} FILE_OBJECTID_BUFFER, *PFILE_OBJECTID_BUFFER;
```
JayFoxRox, line 16: Is this even used? Why is this part of this PR?

lib/winapi/filemanip.c
```diff
@@ -241,6 +242,130 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
     }
 }
 
+BOOL CopyFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName, BOOL bFailIfExists)
+{
+    static const DWORD readBufferSize = 65535;
```
JayFoxRox, line 247: For such constants I prefer constants like 0xFFFF because it will be much more obvious where it originates (not much of an issue for 16 bit, but still; also avoids confusion with 0x10000).

lib/winapi/filemanip.c
```diff
@@ -241,6 +242,130 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+            FILE_READ_ATTRIBUTES | FILE_SYNCHRONOUS_IO_NONALERT | FILE_OPEN_FOR_BACKUP_INTENT | FILE_NO_EA_KNOWLEDGE);
+    if (!NT_SUCCESS(status)) {
+        SetLastError(RtlNtStatusToDosError(status));
+        return FALSE;
+    }
+
+    // Fail overwriting the file if the target == the source. This is
+    // intentionally done after the check for existence of the source
+    // to set the most appropriate LastError value.
+    if (!strcmp(lpExistingFileName, lpNewFileName)) {
```
JayFoxRox, line 277: I'd expect more windows-ish code in the winapi code?

JayFoxRox: There are probably many ways to cheat this, by using something like "./foo" and "./bar/../foo" ?

lib/winapi/filemanip.c
```diff
@@ -241,6 +242,130 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+        DWORD lastError = RtlNtStatusToDosError(status);
+        SetLastError(lastError);
+        return FALSE;
+    }
+
+    RtlInitAnsiString(&targetPath, lpNewFileName);
+    InitializeObjectAttributes(&objectAttributes, &targetPath, OBJ_CASE_INSENSITIVE, ObDosDevicesDirectory(), NULL);
+
+    status = NtCreateFile(
+            &targetHandle,  // FileHandle
```
JayFoxRox, line 299: Double-spaces before comments is odd style; not sure if it's used elsewhere in nxdk

lib/winapi/filemanip.c
```diff
@@ -241,6 +242,130 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+            bFailIfExists ? FILE_CREATE : FILE_SUPERSEDE,  // CreateDisposition
+            FILE_SYNCHRONOUS_IO_NONALERT | FILE_OPEN_FOR_BACKUP_INTENT);  // CreateOptions
+    if (!NT_SUCCESS(status)) {
+        NtClose(sourceHandle);
+        DWORD lastError = RtlNtStatusToDosError(status);
+        SetLastError(lastError);
+        return FALSE;
+    }
+
+    readBuffer = MmAllocateContiguousMemory(readBufferSize);
```
JayFoxRox, line 315: Not sure if a contiguous memory allocation is a good idea here. We should reserve those for GPU resources which can't deal with paging?

Not sure what other parts of winapi do.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -2861,6 +2861,7 @@ XBAPI NTSTATUS NTAPI NtOpenDirectoryObject
 
 #define FILE_DEVICE_FILE_SYSTEM 0x00000009
 #define FSCTL_DISMOUNT_VOLUME CTL_CODE(FILE_DEVICE_FILE_SYSTEM, 8, METHOD_BUFFERED, FILE_ANY_ACCESS)
+#define FSCTL_GET_OBJECT_ID CTL_CODE(FILE_DEVICE_FILE_SYSTEM, 39, METHOD_BUFFERED, FILE_ANY_ACCESS)
```
JayFoxRox, line 2864: This appears to be unused? Why is this part of this PR?

lib/winapi/filemanip.c
```diff
@@ -241,6 +242,130 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+            DWORD bytesWritten = 0;
+            if (!WriteFile(targetHandle, bufferPos, bytesRead, &bytesWritten, NULL)) {
+                MmFreeContiguousMemory(readBuffer);
+                NtClose(sourceHandle);
+                NtClose(targetHandle);
+                return FALSE;
+            }
+            bytesRead -= bytesWritten;
+            bufferPos += bytesWritten;
+        }
```
JayFoxRox, line 346: Doesn't the kernel offer some call or IOCTL to copy files?
This user-space copy loop feels a bit involved and there's probably a lot that can go wrong (stuff like not applying the proper flags to the file when creating it, not taking advantage of file-system optimizations etc.)

lib/winapi/filemanip.c
```diff
@@ -241,6 +242,130 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+        return FALSE;
+    }
+
+    while (TRUE) {
+        const BYTE *bufferPos = readBuffer;
+        if (!ReadFile(sourceHandle, readBuffer, readBufferSize, &bytesRead, NULL)) {
+            MmFreeContiguousMemory(readBuffer);
+            NtClose(sourceHandle);
+            NtClose(targetHandle);
+            return FALSE;
```
JayFoxRox, line 329: Wouldn't these error-cases leave a broken file? Is that also how it works on Windows?

lib/winapi/filemanip.c
JayFoxRox: For minor fixes like this, please try to squash up your changes into the clean commits, so if the review passes, it can be merged immediately without another round-trip.

lib/winapi/winioctl.h
abaire: You're right, it's not anymore. It was part of my attempt to use `NtFsControlFile` to determine if the two paths referenced the same underlying file and I failed to clean it out when I abandoned that approach.

Temporarily adding back in that code in the hopes that there's a silly mistake that you can point out. The issue I'm seeing is that the NFCF call returns STATUS_INVALID_DEVICE_REQUEST

lib/winapi/filemanip.c
abaire: Good point, changed.

lib/winapi/filemanip.c
abaire: Definitely easy ways to bypass this trivial check, ideally I'd like to check if the underlying object IDs used when opening both paths were the same, but it was failing in my xemu-based tests but this was failing.

lib/winapi/filemanip.c
abaire: Habit from the code style at a previous workplace, removed as the file doesn't comment parameters generally.

lib/winapi/filemanip.c
abaire: I didn't see any other allocations in the winapi dir but happy to use whatever is most appropriate (I'm not very familiar w/ the NT kernel).

lib/winapi/filemanip.c
abaire: Totally agree; I didn't see a function that looked appropriate in `xboxkrnl.h` so I assumed there was no lower level method that could be utilized. Similarly I did not have luck searching the win32 docs.

Is there someplace else I could look?

lib/winapi/filemanip.c
abaire: They would. My reasoning was that an attempt to cleanup may spawn further failures and in the case that a file is copied into itself by bypassing the trivial check it might be dangerous to attempt to delete it.

lib/winapi/filemanip.c
abaire: Will do.

lib/winapi/winioctl.h
abaire: Fully removed, it seems likely that FATX doesn't support objectids at all.

lib/winapi/filemanip.c
abaire: On further investigation it seems that fatx doesn't support ".." directory entries (attempting to use one when creating a file returns STATUS_OBJECT_NAME_INVALID)

lib/winapi/filemanip.c
abaire: Switched to `NtAllocateVirtualMemory`

lib/xboxkrnl/xboxkrnl.h
abaire: Removed along with the other objectid checking code.

lib/winapi/filemanip.c
abaire: Poked around a bit more, the only lower level copying I found was added as part of windows server after win2k/xp days.

GXTX: Possibly something more like `NtWriteFile()` than using the user abstraction `WriteFile()` is what he's hinting at?

abaire: Ah, I assume he was looking for something more like https://docs.microsoft.com/en-us/windows/win32/api/winioctl/ni-winioctl-fsctl_duplicate_extents_to_file where the entire copy loop is handled at a lower level (but that's not available in the XBOX kernel). 

I think a switch to `NtWriteFile` would end up duplicating much of the content of `WriteFile` and still operating in user-space. I don't know enough to say if the efficiency gain warrants the maintenance cost, but happy to make that change if it's preferred.

@JayFoxRox can you clarify?

lib/winapi/filemanip.c
```diff
@@ -241,6 +241,138 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+    }
+
+    status = NtCreateFile(
+            &targetHandle,
+            FILE_GENERIC_WRITE,
+            &objectAttributes,
+            &ioStatusBlock,
+            &networkOpenInformation.AllocationSize,
+            networkOpenInformation.FileAttributes,
+            FILE_SHARE_READ | FILE_SHARE_WRITE,
```
thrimbor, line 307: This is a bug: The file should be opened with `0` as the sharing flags, not `FILE_SHARE_READ | FILE_SHARE_WRITE`, or another thread could open the same file and mess with it while we're copying. Using `0` will also make the call fail if we're trying to open the source file again.

lib/winapi/filemanip.c
```diff
@@ -241,6 +241,138 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+    IO_STATUS_BLOCK ioStatusBlock;
+    FILE_NETWORK_OPEN_INFORMATION networkOpenInformation;
+    LPVOID readBuffer = NULL;
+    SIZE_T readBufferRegionSize = 0x1000;
+    DWORD bytesRead;
+
+    RtlInitAnsiString(&sourcePath, lpExistingFileName);
+    InitializeObjectAttributes(&objectAttributes, &sourcePath, OBJ_CASE_INSENSITIVE, ObDosDevicesDirectory(), NULL);
+
+    status = NtOpenFile(
```
thrimbor, line 261: I suggest to replace this with a call to `CreateFile` to reduce duplicated code.

thrimbor (changes requested):

lib/winapi/filemanip.c
thrimbor: Personally, I'd prefer it like `64*1024`.
As for the buffer size, it currently is only 4KiB (64KiB previously). That's a bit small imho - XDK code seem to copy in 16KiB chunks (at least that's the size Halo 2 uses to copy files), which I think is a reasonable compromise between performance and saving memory.

lib/winapi/filemanip.c
thrimbor: Good catch! I'd prefer to have the change in a follow-up PR, you can create an issue if you want for tracking it.

lib/winapi/fileapi.h
```diff
@@ -37,6 +37,7 @@ BOOL DeleteFileA (LPCTSTR lpFileName);
 BOOL RemoveDirectoryA (LPCSTR lpPathName);
 BOOL CreateDirectoryA (LPCSTR lpPathName, LPSECURITY_ATTRIBUTES lpSecurityAttributes);
 BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName);
+BOOL CopyFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName, BOOL bFailIfExists);
```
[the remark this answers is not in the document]
thrimbor: Correct, I added `DeleteFileA` to the issue to keep track of it.

lib/winapi/filemanip.c
```diff
@@ -241,6 +241,138 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+    status = NtQueryInformationFile(
+        sourceHandle,
+        &ioStatusBlock,
+        &networkOpenInformation,
+        sizeof(networkOpenInformation),
+        FileNetworkOpenInformation);
+    if (!NT_SUCCESS(status)) {
+        DWORD lastError = RtlNtStatusToDosError(status);
+        SetLastError(lastError);
+        return FALSE;
```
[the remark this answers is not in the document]
thrimbor: Sorry, I should've been more clear about what I meant. I meant to suggest you use `CloseHandle` instead of `NtClose` where we're interested in the error value, since `CloseHandle` already converts the `NTSTATUS` and calls `SetLastError`, but where we don't care about that (or want to preserve the previous error), a call to `NtClose` is better.

lib/winapi/filemanip.c
```diff
@@ -241,6 +246,145 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+    FILE_BASIC_INFORMATION fileBasicInformation;
+    FILE_NETWORK_OPEN_INFORMATION networkOpenInformation;
+    LPVOID readBuffer = NULL;
+    SIZE_T readBufferRegionSize = 64 * 1024;
+    DWORD bytesRead;
+
+    sourceHandle = CreateFile(
+        lpExistingFileName,
+        FILE_GENERIC_READ,
+        0,
```
thrimbor, line 267: You can use `FILE_SHARE_READ` here instead, that will allow other threads to read from the same file, while still preventing it from getting modified.

lib/winapi/filemanip.c
```diff
@@ -241,6 +241,138 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+                NtClose(targetHandle);
+                return FALSE;
+            }
+            bytesRead -= bytesWritten;
+            bufferPos += bytesWritten;
+        }
+    }
+
+    NtFreeVirtualMemory(&readBuffer, &readBufferRegionSize, MEM_RELEASE);
+
```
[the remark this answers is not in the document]
thrimbor: > I was under the impression that passing the file attributes to `NtCreateFile` above would already copy them over when creating/superseding the file.

That should be enough to copy the file attribute flags, correct.

> I also assumed that the access info should actually be based on the time of copy rather than brought over from the original.

The wording in the MSDN is a bit ambiguous, so I checked on Windows. It appears that only `LastWriteTime` is copied from the original file, the others are left as-is.

lib/winapi/filemanip.c
```diff
@@ -241,6 +246,145 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+        NULL,
+        OPEN_EXISTING,
+        FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN,
+        NULL);
+    if (sourceHandle == INVALID_HANDLE_VALUE) {
+        return FALSE;
+    }
+
+    RtlInitAnsiString(&sourcePath, lpExistingFileName);
+    InitializeObjectAttributes(&objectAttributes, &sourcePath, OBJ_CASE_INSENSITIVE, ObDosDevicesDirectory(), NULL);
```
thrimbor, line 277: These can be removed now, together with `sourcePath`.

lib/winapi/filemanip.c
abaire: Sounds good, filed #504

lib/winapi/filemanip.c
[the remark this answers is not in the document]
abaire: Gotcha. 
I'd be a little worried that mixing the two may look like a bug/mistake if the reader does not know the nuanced difference between the behavior around setting `LastError` (like me, with very rusty memory about win32). I'd personally be inclined to use the more explicit error handling across the board (i.e., stick with `NtClose`) despite the slight duplication just for the sake of consistency, but defer to you.

lib/winapi/filemanip.c
abaire: Sorry, should've caught that before pushing. Fixed.

lib/winapi/filemanip.c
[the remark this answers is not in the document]
abaire: Cool, looks like passing 0 should cause it to retain the copied file's value so I'm explicitly copying over `LastWriteTime` and (re)setting `FileAttributes`.

lib/winapi/filemanip.c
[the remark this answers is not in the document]
abaire: (I'll wait to make sure we're aligned before changing one way or the other and will wait to squash changes again until it's done so comments don't get orphaned while still under review)

lib/winapi/filemanip.c
abaire: Done.

lib/winapi/filemanip.c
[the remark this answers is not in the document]
abaire: @thrimbor - just wanted to make sure you saw my comment above. Happy to go in whatever direction you prefer!

lib/winapi/filemanip.c
thrimbor: It's fine like this.

thrimbor (changes requested):
This looks very close to being ready to merge now, so I should probably mention that, while this isn't explicitly mentioned in this file, the winapi part of nxdk is licensed under the MIT license. Just want to make sure you're aware and fine with that.
You don't have to add any license or copyright info to this file yourself, I'll add that when I slap SPDX headers on everything.

lib/winapi/filemanip.c
```diff
@@ -100,14 +101,22 @@ BOOL SetFileAttributesA (LPCSTR lpFileName, DWORD dwFileAttributes)
     return TRUE;
 }
 
+static NTSTATUS DeleteHandle(HANDLE handle)
```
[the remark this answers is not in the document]
abaire: Fixed. 

I ended up adding clang-format to NevolutionX to help avoid having these types of oversights eat up reviewer time, I wonder if there's any appetite to do something similar here? Given that the nxdk is more active it might be a bit more painful since the initial fixup pass will (probably) make a lot of formatting changes that might trigger merge conflicts on outstanding PRs, but once past that it would enforce consistency.

lib/winapi/filemanip.c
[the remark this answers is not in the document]
abaire: Sounds good, I think it may be too nuanced to split usage and worry someone might "fix" it in the future, potentially introducing changes to the `LastError` behavior.

lib/winapi/filemanip.c
```diff
@@ -241,6 +246,139 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+    status = NtCreateFile(
+            &targetHandle,
+            FILE_GENERIC_WRITE,
+            &objectAttributes,
+            &ioStatusBlock,
+            &networkOpenInformation.AllocationSize,
+            networkOpenInformation.FileAttributes,
+            0,
+            bFailIfExists ? FILE_CREATE : FILE_SUPERSEDE,
+            FILE_SYNCHRONOUS_IO_NONALERT | FILE_OPEN_FOR_BACKUP_INTENT);
```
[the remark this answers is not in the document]
abaire: Done, much appreciated.

I notice there are a few other usages in this file ('DeleteFileA', `RemoveDirectoryA`, and `MoveFileA`), should they get similar changes (clearly not the same for `RemoveDirectoryA` at least, but none of those usecases seem to imply backups to me).

abaire (commented):
Thanks for the heads up on the license, MIT is totally fine with me.

lib/winapi/filemanip.c
```diff
@@ -241,6 +246,146 @@ BOOL MoveFileA (LPCTSTR lpExistingFileName, LPCTSTR lpNewFileName)
...
+            bytesRead -= bytesWritten;
+            bufferPos += bytesWritten;
+        }
+    }
+
+    status = NtFreeVirtualMemory(&readBuffer, &readBufferRegionSize, MEM_RELEASE);
+    if (!NT_SUCCESS(status)) {
+      DWORD lastError = RtlNtStatusToDosError(status);
+      SetLastError(lastError);
+    }
```
abaire, line 353: Should the method fail if it fails to free the temp memory? Technically I think the copy could still succeed, and I don't know that there's anything the caller could do about the leak other than abort entirely.

lib/winapi/filemanip.c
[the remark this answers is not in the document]
abaire: Changed.

Issue #500, opened 2021-09-15
XLaunchXBE cannot launch paths that start with "d:"


abaire:
Calling XLaunchXBE with a path that starts with `d:` or `D:` fails to launch the XBE.
Passing an explicit expanded path (i.e., prefixing with `\Device\CdRom0`) works, as does passing a path with no drive prefix at all).

I'm guessing this is due to the table at https://github.com/XboxDev/nxdk/blob/master/lib/hal/fileio.c#L15 that expands the `d:` prefix to `\??\D:\`.

Example project to repro the issue at abaire/nxdk_launch_test

JayFoxRox:
> I'm guessing this is due to the table at https://github.com/XboxDev/nxdk/blob/master/lib/hal/fileio.c#L15 that expands the `d:` prefix to `\??\D:\`.

Yes; these are old OpenXDK functions (which suck). They have a lot of edge cases and hardcoded behaviour.

Those old functions should be rewritten for libnxdk in https://github.com/XboxDev/nxdk/tree/master/lib/nxdk (with proper mapping and mounting / unmounting handled by the kernel, as intended by MS).

PR #501, opened 2021-09-15
Expands XLaunchXBE to allow caller to set LaunchData.

abaire:
* Adds an extended version of `XLaunchXBE` that allows the `LaunchData` member to be set on the persisted launch data struct.
* Adds a helper method to retrieve the launch type and data (with handling of the launch-from-dashboard case).

Basic test at abaire/nxdk_launch_test

Note that there seems to be a bug in the handling of "D:" prefixed paths (see #500).

### DIFF
--- a/lib/hal/xbox.c
+++ b/lib/hal/xbox.c
@@ -9,51 +9,78 @@
 
 #define KernelMode 0
 
+#define LaunchDataPageSize 0x1000
+
 void XReboot()
 {
-	HalReturnToFirmware(HalRebootRoutine);
+    HalReturnToFirmware(HalRebootRoutine);
 }
 
-/**
- * Launches an XBE.  Examples of xbePath might be:
- *   c:\\blah.xbe
- *   c:/foo/bar.xbe
- * If the XBE is able to be launched, this method will
- * not return.  If there is a problem, then it return.
- */
-void XLaunchXBE(char *xbePath)
+int XGetLaunchInfo(unsigned long *launchDataType, const unsigned char **launchData)
 {
-#if 0
-	struct stat statbuf;
-	if (stat(xbePath, &statbuf) < 0)
-		return;
-#endif
+    *launchDataType = LDT_NONE;
+    *launchData = NULL;
 
-    if (LaunchDataPage == NULL)
-        LaunchDataPage = MmAllocateContiguousMemory(0x1000);
+    if (LaunchDataPage == NULL) {
+        LaunchDataPage = MmAllocateContiguousMemory(LaunchDataPageSize);
+        if (LaunchDataPage == NULL) {
+            return -1;
+        }
+        memset(LaunchDataPage, 0, LaunchDataPageSize);
+        LaunchDataPage->Header.dwLaunchDataType = LDT_FROM_DASHBOARD;
+    }
 
-    if (LaunchDataPage == NULL)
-		return;
+    *launchDataType = LaunchDataPage->Header.dwLaunchDataType;
+    *launchData = LaunchDataPage->LaunchData;
+    return 0;
+}
 
-    MmPersistContiguousMemory(LaunchDataPage, 0x1000, TRUE);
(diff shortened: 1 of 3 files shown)


thrimbor:
Do you have a specific use case for this code?
Not that I think what you're adding isn't useful, but that code is inherited from OpenXDK and has several fundamental flaws and limitations (some of which https://github.com/XboxDev/nxdk/issues/500 hints at). We're slowly deprecating and replacing the old OpenXDK parts, and it might make sense to delay adding new features until we have come up with a better API.

abaire:
Initially I had two use cases:
One was dusting off my old MAMEoX project to see if I can get it buildable w/ the nxdk (now that I'm more familiar with the nxdk I think this is still a ways away)

The other would be using a similar approach to MAMEoX in having a menu launcher and a series of worker XBEs in some of the tests I've been writing to aid xemu contributions (e.g., https://github.com/abaire/nxdk_pgraph_tests). There are a few tests that are reliant on having a clean boot state to be valid; though this may just be because I haven't figured out the right set of explicit commands to reset to initial state (e.g., the behavior of the fixed function pipeline when lighting is enabled in a BEGIN_END but no normals are explicitly provided w/ the vertex data).

That said, at the moment I'm already maintaining some patches on top of the nxdk to expose the internals of pbkit and I could always write info to a cache file so I'm not blocked by not having this functionality.

lib/hal/xbox.c
```diff
@@ -9,51 +9,78 @@
...
-	//   \Device\Harddisk0\Partition2\blah;doom.xbe
-	char *lastSlash = strrchr(LaunchDataPage->Header.szLaunchPath, '\\');
-	if (lastSlash != NULL)
-	{
-		*lastSlash = ';';
-		HalReturnToFirmware(HalQuickRebootRoutine);
-	}
+    if (launchData) {
+        memcpy(launchDataPage->LaunchData, launchData, sizeof(launchDataPage->LaunchData));
+    }
```
JayFoxRox, line 83: If `launchData` is `NULL` the `LaunchDataPage` will still be allocated and point at zero-bytes; is this intended?

lib/hal/xbox.c
```diff
@@ -9,51 +9,78 @@
...
-		return;
-#endif
+    XLaunchXBEEx(xbePath, NULL);
+}
 
-    if (LaunchDataPage == NULL)
+void XLaunchXBEEx(const char *xbePath, const void *launchData)
+{
+    if (LaunchDataPage == NULL) {
         LaunchDataPage = MmAllocateContiguousMemory(0x1000);
```
JayFoxRox, line 46: This should probably be using `LaunchDataPageSize` now?

lib/hal/xbox.c
```diff
@@ -9,51 +9,78 @@
...
-	memset((void*)LaunchDataPage, 0, 0x1000);
+        // one last thing... xbePath now looks like:
+        //   \Device\Harddisk0\Partition2\blah\doom.xbe
+        // but it has to look like:
+        //   \Device\Harddisk0\Partition2\blah;doom.xbe
+        char *lastSlash = strrchr(launchDataPage->Header.szLaunchPath, '\\');
+        if (!lastSlash) {
+            // if we couldn't find a trailing slash, the conversion to
+            // the xbox path mustn't have worked, so we will return
+            return;
```
JayFoxRox, line 75: nit: What should happen to the `LaunchDataPage` here? Should the allocation be undone?

thrimbor:
I gave this another look (sorry for taking so long), and I think it's fine to merge this as is. There are a couple of design issues, but those stem from the original code, not your changes.
If there's no objection, I'll merge it tomorrow.

JayFoxRox:
Commit prefix is incorrect; should be `hal` as its modifying old OpenXDK HAL code, not the newer nxdk lib.

thrimbor:
I took the liberty of fixing that myself. Merging now.

PR #505, opened 2021-09-17
winapi: Implement SetFileTime()/GetFileTime()

### LINKED ISSUE #470: `GetFileTime` is missing
There are a handful of projects that seem to depend on this. Should be easy to implement, so leaving it here as a potential task.

thrimbor:
Implements `SetFileTime()` and `GetFileTime()` (I just read up on file attributes again and decided to go for these). Not tested beyond compilation yet.

Fixes #470


thrimbor:
It's fixed and tested now.

Issue #507, opened 2021-09-17
Graphical corruption when using `HalQuickRebootRoutine` to return to dashboard


abaire:
When using `HalQuickRebootRoutine` with a `NULL` `szLaunchPath` and a `dwLaunchDataType` of `LDT_NONE` (`0xFFFFFFFF`) there is a brief period of graphical corruption before the dashboard is displayed.

`XVideoInit` looks like it's doing reasonable things with respect to setting `AvSetSavedDataAddress` to the framebuffer so I'm not sure what is causing this. 

- I do not see this corruption when doing a trivial launch of another nxdk-built XBE on the same data storage.
- I confirmed that I can grab the framebuffer via `XVideoGetFB`, memset it, flush it, and see the change reflected briefly before it's replaced by corruption.
- Reproduced both in xemu and on hardware, relatively trivial test program at https://github.com/abaire/nxdk_launch_test

abaire:
Playing around with this a bit more, it seems that freeing the framebuffer set by the dashboard and retrieved via `AvGetSavedDataAddress` at https://github.com/XboxDev/nxdk/blob/f2c634f8774b2de2b958326f5d9e78a27ce827bf/lib/hal/video.c#L377 is what triggers the problem (if I comment out the lines unpersisting and freeing the previous FB the issue goes away).

I'm not sure if this is an indication that `AvSetSavedDataAddress` is somehow not working or if the kernel/dashboard/... just blindly uses the memory at `0x80061000` as the framebuffer when initializing the dashboard.

PR #509, opened 2021-09-23
hal: Rewrite framebuffer reusal logic

### LINKED ISSUE #506: Possible leak of `framebufferMemory` in `XVideoInit`
`XVideoInit` does a check to see if the previously saved framebuffer is already the size of the requested framebuffer at
https://github.com/XboxDev/nxdk/blob/f2c634f8774b2de2b958326f5d9e78a27ce827bf/lib/hal/video.c#L373. If it is the same size, it reuses the previously saved framebuffer, but does not deallocate `framebufferMemory` or pass the newly reassigned `_fb` pointer to `AvSetDisplayMode`.

I'm not sure if it's a leak or if it's important that there's a mismatch between the framebuffer used in `AvSetDisplayMode` and the data persisted via `AvSetSavedDataAddress` and `XVideoGetFB`.

thrimbor:
There were a few issues with the previous code:
* The memory allocated for a new framebuffer was leaked if a previous framebuffer was reused
* There was separate code to check whether the previous framebuffer allocated by the currently running program could be reused - this is unnecessary if we already use `AvGetSavedDataAddress`/`AvSetSavedDataAddress`, and the global variable `framebufferMemory` can be removed
* `XVideoSetFB` was modifying and persisting the passed framebuffer address, which it shouldn't, because it doesn't own the memory and can't guarantee that the allocation will still be valid after the program quits

This reorganizes and simplifies the code to solve the issues above.

Fixes #506

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -35,7 +35,6 @@
 #define VIDEO_R5G6B5				0x00000011
 #define VIDEO_A8R8G8B8				0x00000012
 
-static unsigned char*	framebufferMemory = NULL;
 static unsigned char*	_fb = NULL;
 static DWORD			dwEncoderSettings 	= 0;
 static VIDEO_MODE		vmCurrent = { 0, 0, 0, 0 };
@@ -297,10 +296,6 @@ void XVideoSetFB(unsigned char *fb)
 	assert(((unsigned int)fb & ~0x7FFFFFFF) == 0);
 	_fb = fb;
 
-	SIZE_T fbSize = MmQueryAllocationSize(_fb);
-	MmPersistContiguousMemory(_fb, fbSize, TRUE);
-	AvSetSavedDataAddress(_fb);
-
 	VIDEOREG(PCRTC_START) = (unsigned int)_fb & 0x7FFFFFFF;
 }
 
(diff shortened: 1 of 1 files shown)


lib/hal/video.c
```diff
@@ -341,18 +336,28 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
...
 	XVideoSetVideoEnable(FALSE);
 
+	PVOID framebufferMemory = AvGetSavedDataAddress();
+
 	if (framebufferMemory != NULL) {
-		MmPersistContiguousMemory(framebufferMemory, screenSize, FALSE);
-		MmFreeContiguousMemory(framebufferMemory);
+		SIZE_T frameBufferSize = MmQueryAllocationSize(framebufferMemory);
+		if (frameBufferSize != screenSize) {
+			MmPersistContiguousMemory(framebufferMemory, frameBufferSize, FALSE);
```
abaire, line 344: In case you didn't already see it, #507 seems to be triggered by freeing the saved framebuffer when an XBE is launched by the dashboard. 
I'm wondering if the dashboard expects to have the framebuffer at a specific address and some work needs to be done to relocate mem such that the new framebuffer allocated later in this method can be put at the same addr?

thrimbor: I know the dashboard retrieves the address from `AvGetSavedDataAddress`, so that shouldn't be an issue.
I do have a suspicion what the problem could be, though, and will try to confirm it later today.

thrimbor:
Superseded by #511.

PR #511, opened 2021-09-25
Revert #424, free persistent framebuffer

### LINKED ISSUE #506: Possible leak of `framebufferMemory` in `XVideoInit`

### LINKED ISSUE #507: Graphical corruption when using `HalQuickRebootRoutine` to return to dashboard

thrimbor:
This reverts #424, but adds code that deletes a persistent framebuffer if one exists, so that #357 stays resolved.

I was looking into why #507 might happen for #509, and found that the issue is most likely caused by a fundamental issue with how we treat the allocated memory. Our code treats the memory like it just contains the plain framebuffer, while the XDK code in the dash expects to find a `D3DSurface` (name derived from Cxbx-Reloaded) struct describing the framebuffer, with the data following after that. Rebooting from our code to the dash would then pass a corrupted struct to the dash, which I assume is the reason for the screen corruption. This would also affect other any other xbe launched from nxdk code, hence the removal.

Allowing proper handling of framebuffer persistence is a feature I'd like to have, but this API is just not up to the task, a future redesign/rewrite is necessary.

Fixes #506 
Fixes #507 
Supersedes #509

@abaire: I'd appreciate if you could confirm that this fixes #507 for you.

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -336,6 +336,14 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 
 	XVideoSetVideoEnable(FALSE);
 
+	PVOID previousFB = AvGetSavedDataAddress();
+	if (previousFB != NULL) {
+		SIZE_T previousFBSize = MmQueryAllocationSize(previousFB);
+		MmPersistContiguousMemory(previousFB, previousFBSize, FALSE);
+		MmFreeContiguousMemory(previousFB);
+		AvSetSavedDataAddress(NULL);
+	}
+
 	if (framebufferMemory != NULL) {
 		MmFreeContiguousMemory(framebufferMemory);
 	}
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -296,11 +296,6 @@ void XVideoSetFB(unsigned char *fb)
 {
 	assert(((unsigned int)fb & ~0x7FFFFFFF) == 0);
 	_fb = fb;
-
-	SIZE_T fbSize = MmQueryAllocationSize(_fb);
-	MmPersistContiguousMemory(_fb, fbSize, TRUE);
-	AvSetSavedDataAddress(_fb);
-
 	VIDEOREG(PCRTC_START) = (unsigned int)_fb & 0x7FFFFFFF;
 }
 
@@ -342,7 +337,6 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 	XVideoSetVideoEnable(FALSE);
 
 	if (framebufferMemory != NULL) {
-		MmPersistContiguousMemory(framebufferMemory, screenSize, FALSE);
 		MmFreeContiguousMemory(framebufferMemory);
 	}
 	framebufferMemory = MmAllocateContiguousMemoryEx(screenSize,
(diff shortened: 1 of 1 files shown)


abaire:
Very interesting, thanks! 

This definitely improves the behavior, but it's still acting a bit odd.

I tested two cases, both behave the same:
- Main dashboard in 480p which launches NevolutionX in 720p. From NeX I `reboot` (which just triggers an `exit(0);`, I think we'd ideally explicitly launch the dashboard via #501).
- Main dash in 480p and NeX in 480p

Previously on hardware I would get a very predictable pattern of corruption that was entirely unrelated to anything NeX had displayed prior to reboot (Xemu would also display a predictable, but different pattern). That feels like it validates your finding about the dashboard expecting a struct and not just the framebuffer. It was roughly an alternating pattern of several lines of primarily (but not entirely) white then several lines of black all across the screen. Now I get a corrupted version of the NeX background image along the top part of the screen with the bottom part remaining clear. This happens in both 720->480 and 480->480 scenarios.

Out of curiosity, I tried grabbing the framebuffer and memsetting a chunk of it to 0 prior to the `exit` call:
```
  unsigned char* fb = XVideoGetFB();
  memset(fb, 0, 480 * 480 * 4);
  XVideoFlushFB();
  exit(0);
```
This gives me a solid black section at the top (as expected), but also shows that there is a period of time where the un-memset portion of the NeX background remains uncorrupted before becoming corrupt. I'm not sure if this is when execution is handed back over from the NeX process, but it takes long enough that it's clearly noticeable (but still well less than a second). It also does not seem to be related to a mode change for the video encoder, it stays in 720p mode until the primary dashboard starts up again.


For NeX I think we can just properly blank the whole framebuffer for now so this patch is clearly a major improvement; that said it'd be great to be able to persist an actual loading screen so it's clear the dashboard didn't crash to a black screen so I'm definitely looking forward to the future redesign to make that possible :)

Here's the new corrupted state: (Apologies for the crappy picture of my monitor display, I don't have a capture device)
424_new_behavior

abaire:
As a followup: I also patched #501 and switched the `exit` out for a call to launch the dashboard - no change in behavior. I can also repro less pronounced corruption in xemu (which is not exactly the same behavior as on HW).

abaire:
And another followup: I made the change in NeX to blank and flush the framebuffer before exiting and still see some corruption (still dramatically better than without this patch). Looks like it's approximately in the same row where the pitch mismatch ends in the non-blank case (the row just before the IP display in the screenshot I posted previously).

thrimbor:
I've experimented with NeX a bit in xemu and on real hardware, and I think the behavior you're seeing might be "normal".

When rebooting, all the memory allocations done by our code become invalid (except the persistent ones ofc), so the framebuffer might get overwritten by the kernel or the new xbe. The GPU however does not know of such things, and will happily continue reading from the physical memory location where the framebuffer resided, and turn that into an image on screen, so until the video output gets reconfigured, you might see garbage on screen.

I see two options to prevent this:
1. A persistent framebuffer (a proper implementation will require a redesigned API though, so this is a mid- to long-term thing)
2. Turning the video output off. A `XVideoSetVideoEnable(FALSE);` before the `exit(0);` will make the screen turn black until video output gets turned on again (by the next xbe). xemu doesn't seem to implement this behavior yet, so you'll still see a little bit of corruption there.

abaire:
Makes sense, I'll modify my PR to disable video output - thanks for the advice!

Ryzee119:
Tested this on a few samples and a few of my ports etc. No issues.

I rebased locally and tested against current master too. All good

PR #518, opened 2021-10-16
build system restructuring

thrimbor:
This starts with some of the restructuring necessary to make nxdk look more like a traditional toolchain.

It adds wrappers for a couple of necessary tools, and adds a script that performs the necessary environment variable modifications as well as an LLVM version check. The script is inspired by Python's `virtualenv` and [this article](https://web.archive.org/web/20201111231240/https://datagrok.org/python/activate/). It can either be run without any parameters, in which case it will create a properly set up subshell, or it can be integrated into a shell with `eval $(./bin/activate -s)`.

A wrapper for `pkg-config` is included, as well as the necessary pkgconfig files for the non-nxdk libraries we currently include in the repo. This allows programs to be built without our `Makefile`, for example, the ttf sample can be built like this `nxdk-cc main.c $(nxdk-pkg-config SDL2_image --libs --cflags) ${NXDK_DIR}/lib/libpdclib.lib ${NXDK_DIR}/lib/libwinapi.lib ${NXDK_DIR}/lib/xboxkrnl/libxboxkrnl.lib ${NXDK_DIR}/lib/nxdk_usb.lib ${NXDK_DIR}/lib/libxboxrt.lib ${NXDK_DIR}/lib/libnxdk_hal.lib`

It also includes CMake support. There's a chance it won't handle corner-cases correctly yet, but it's enough to build `zstd` with my package manager prototype.

**Required changes for nxdk users:**
* If you're using the Docker image, no changes should be required. The entrypoint script takes care of everything.
* If you're using nxdk natively, either run your build in a subshell created by running the `activate` script, or run `eval $(./usr/src/nxdk/bin/activate -s)` in your buildscript to set things up.

### DIFF
--- a/lib/pkgconfig/SDL2_image.pc
+++ b/lib/pkgconfig/SDL2_image.pc
@@ -1,0 +1,6 @@
+Name: SDL2_image
+Description: image loading library for Simple DirectMedia Layer
+Version: 2.0.5
+Requires: sdl2 >= 2.0.9 libjpeg libpng
+Libs: ${NXDK_DIR}/lib/libSDL2_image.lib
+Cflags: -I${NXDK_DIR}/lib/sdl/SDL2_image
--- a/lib/pkgconfig/SDL2_ttf.pc
+++ b/lib/pkgconfig/SDL2_ttf.pc
@@ -1,0 +1,6 @@
+Name: SDL2_ttf
+Description: ttf library for Simple DirectMedia Layer with FreeType 2 support
+Version: 2.0.14
+Requires: sdl2 >= 2.0.9
+Libs: ${NXDK_DIR}/lib/libSDL_ttf.lib ${NXDK_DIR}/lib/libfreetype.lib
+Cflags: -I${NXDK_DIR}/lib/sdl -I${NXDK_DIR}/lib/sdl/SDL_ttf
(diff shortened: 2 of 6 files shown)


thrimbor:
cc @JayFoxRox because he worked on similar wrappers

dracc:
I've now tested building NevolutionX with these changes.
I get some linking issues as I haven't added SDL2 and the USB library to my `Makefile`.

Is CMake the recommended build system once this PR has been merged? The `make` path feels like it becomes quite a bit rougher.

thrimbor:
> Is CMake the recommended build system once this PR has been merged? The `make` path feels like it becomes quite a bit rougher.

CMake is somewhat of an option now, but you can still use `make` just like before, apart from having to call the `activate` script.

> Am I being too sloppy or is there an error here somewhere?

I'm not sure - it doesn't really make sense to use `pkg-config` for building nxdk itself, but adding that to `CXXFLAGS` shouldn't cause an error, especially since the file is not C++.

dracc:
Silly me. It appears I had a faulty SDL2 branch checked out, as updating my submodules fixed my issues.
NevolutionX builds just fine with this and the resulting .iso runs just like before.

LGTM! :+1:

PR #520, opened 2021-11-07
vcrt: Fix sign related compilation warnings

dracc:
Small change silencing a few compilation warnings related to signed/unsigned comparisons.


GXTX:
We seem to use cpp style casts elsewhere in this file, is there any reason to not use them here?

dracc:
> We seem to use cpp style casts elsewhere in this file, is there any reason to not use them here?

Lack of attention. I'll correct that. Thanks!

thrimbor (changes requested):

thrimbor (approved):
Changes look good. The only modification I'd like to see is the commit title prefix - we have used `vcruntime` instead of `vcrt` in the past.

PR #523, opened 2021-12-23
fixes incorrect NV04_SOLID_LINE_END value

abairmj:
NV04_SOLID_LINE_END was incorrectly set to same value as NV04_SOLID_LINE_START.


thrimbor:
Where does the new value come from? Is there a test so we can verify it's valid?
Can you fix the commit title to follow the convention?

LukeUsher:
Are you certain this is correct? NV04_SOLID_LINE_END is 0x400 in Nouveau also. 

See: https://github.com/Starlink/mesa/blob/c9c0fa3814f8d915da0e214e00c007a05d3c5c13/src/mesa/drivers/dri/nouveau/nouveau_reg.h#L210

abairmj:
I'm guessing 2d lines aren't an often tested path.  

The test for it is below
https://github.com/abaire/nxdk_pgraph_tests/blob/5ffe30554301a677be2340cc2c75ca83ad19149f/tests/two_d_line_tests.cpp#L79

Also it didn't make sense to push the same command for start and end, I would think it would just overwrite the value for the first command I pushed.

The observed behavior from the test seems to confirm this.  Without this fix no line is ever drawn (same as start == end in the test code)

Output from the 2d line tests on hw available here.
 https://github.com/abaire/nxdk_pgraph_tests_golden_results/blob/a726cbb46b8e0902f869a32f909a2e0cf3e0eb0b/2DLine-15-C000003E0-400_200-400_400.png

thrimbor:
Confirmed working on my Xbox. Thanks for the fix!

PR #524, opened 2021-12-25
build: Fix Docker entrypoint not getting cli arguments

thrimbor:
While going through a GH Actions log, I realized that the Docker container was no longer building the samples, despite the command exiting without an error. Checking the Docker docs, I realized that the two different syntaxes for `ENTRYPOINT` have different behavior, and the one we were using was suppressing cli arguments, so our entrypoint script was getting no command to run, and just happily quit.

Due to this screw-up on my part, all Docker images built since https://github.com/XboxDev/nxdk/pull/518 was merged were broken.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ ENV NXDK_DIR=/usr/src/nxdk
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


Ryzee119:
Good pickup. Thanks

PR #525, opened 2021-12-30
hal: Disable flicker and soften filter on video init

GXTX:
Is there any reason why we set encoder filters for the user on init?

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -367,8 +367,8 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 	   set it here, too. */
 	_fb = framebufferMemory;
 
-	XVideoSetFlickerFilter(5);
-	XVideoSetSoftenFilter(TRUE);
+	XVideoSetFlickerFilter(0);
+	XVideoSetSoftenFilter(FALSE);
 
 	for(i = 0; i < 256; i++) {
 		defaultGammaRampEntries[i].red = i;


thrimbor:
> Is there any reason why we set encoder filters for the user on init?

To have a default, and not rely on whatever was set previously.

Is there a reason for changing those values?

GXTX:
> Is there a reason for changing those values?

Is there a reason to set them on? It seems like something the developer/user would want to set if required. I certainly wouldn't expect an `init` function to set them for me.

thrimbor:
So that's a no?

GXTX:
@thrimbor off is a sane default. That's the reason to change them.

thrimbor:
> @thrimbor off is a sane default. That's the reason to change them.

As is "on", so what's your point?

This code and its defaults are older than nxdk itself, and thus might be expected / relied upon. Its behavior should not be changed in such a way unless there's an objective reason for doing so.

GXTX:
I don't know what more justification you want to hear me say. Off is sane, on with random settings such as 5 whereas reported max is 7 is not.

thrimbor:
> I don't know what more justification you want to hear me say. Off is sane, on with random settings such as 5 whereas reported max is 7 is not.

As I already said, a single objective reason for changing the defaults would suffice. But with you being unable or unwilling to provide one, I don't see this

GXTX:
Unfortunate you feel that way.

PR #526, opened 2022-01-02
vcruntime: Fix SEH _finally blocks running in wrong stack frame

thrimbor:
This fixes an error in our SEH implementation I noticed while working on C++ exceptions (it caused a crash there).
The error is that, instead of running the `__finally` funclet directly with the correct `ebp` value (funclets use `ebp`-relative addressing of local variables due to being called into from code that executes higher up on the stack), the code sets `ebp`and then calls `_local_unwind2`, which modifies `ebp` itself, leading to an incorrect value in the register when executing the funclet.
This causes all accesses to function-local variables by the funclet to read from or write to the stack frame of `_local_unwind2` instead of the function the funclet resides in.

To fix this, I introduced a wrapper function `call_ebp_func`, which handles calling a funclet with a specific `ebp`-value set including restoring the original value, removed the `ebp`-modification before the call to `_local_unwind2`, modified `_local_unwind2` to use the new wrapper function instead of directly calling the funclet like a function, and reduced the amount of inline assembly where it wasn't really necessary anymore.

<details>
<summary>Code to test the change is here. Only after the fix will it show the correct value of `deadbeef`.</summary>

```
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>

void b ()
{
    debugPrint("%s\n", NULL);
}

void a ()
{
    int x = 0xDEADBEEF;
    __try {
        b();
    } __finally {
        debugPrint("%x\n", x);
    }
}

int main(void)
{

    __try {
        a();
    } __except (EXCEPTION_EXECUTE_HANDLER) {
    }

    while(1) {
        Sleep(2000);
    }

    return 0;
}
```

</details>

### DIFF
--- a/lib/xboxrt/vcruntime/excpt.cpp
+++ b/lib/xboxrt/vcruntime/excpt.cpp
@@ -32,6 +32,23 @@ extern "C" int _nested_unwind_handler (_EXCEPTION_RECORD *pExceptionRecord, EXCE
     return DISPOSITION_COLLIDED_UNWIND;
 }
 
+static inline void *call_ebp_func (void *func, void *_ebp)
+{
+    void *result;
+
+    asm volatile (
+        "pushl %%ebp;"
+        "movl %%ebx, %%ebp;"
+        "call *%%eax;"
+        "popl %%ebp;"
+        : "=a"(result)
+        : "a"(func), "b"(_ebp)
+        : "ecx", "edx", "memory"
+    );
+
+    return result;
+}
+
 extern "C" void _local_unwind2 (EXCEPTION_REGISTRATION_SEH3 *pRegistrationFrame, int stop)
 {
     // Manually install exception handler frame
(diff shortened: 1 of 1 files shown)


mborgerson:
FYI I've not reviewed this carefully, and have no objections to it. You can self merge if you think this is ready.

lib/xboxrt/vcruntime/excpt.cpp
```diff
@@ -32,6 +32,23 @@ extern "C" int _nested_unwind_handler (_EXCEPTION_RECORD *pExceptionRecord, EXCE
     return DISPOSITION_COLLIDED_UNWIND;
 }
 
+static inline void *call_ebp_func (void *func, void *_ebp)
+{
+    void *result;
+
+    asm volatile (
+        "pushl %%ebp;"
```
JayFoxRox, line 40: Not sure, but there might be SSE stack alignment violations here (I assume this also affects Pentium 3?).
I believe the stack must be incremented and decremented in steps of 0x10 bytes if the callee (potentially) uses local SSE variables (I think gcc started doing this).
Otherwise, if the handler has local SSE variables, those might be suddenly misaligned and then the CPU would crash on access (when using aligned access, which is faster).

We might also violate similar rules elsewhere in nxdk, and I'm not sure about the specifics. I just wanted to mention it in case there are weird (seemingly random) crashes later on. Someone should probably take a look at the compiler docs.

I'm also aware of similar rules with malloc, because some C compilers will assume specific alignments on returned buffers - while working on some game wrappers for Windows games I ran into this, sometimes as late as 20 minutes into gameplay, because some tiny-mem-allocator would double-free due to some address-logic.

PR #528, opened 2022-01-04
build: Undefine __STDC_NO_THREADS__ for compatibility with clang 14

thrimbor:
The current git version of clang, which will become clang 14 once released, defines `__STDC_NO_THREADS__`, because MSVC does not provide C11 threads: https://reviews.llvm.org/D112081
nxdk, however, does, and the define breaks building our C11 threads implementation. This adds a parameter to our compiler wrappers to remove that define again.

Tested with this morning's git version of LLVM (https://github.com/llvm/llvm-project/commit/75db002725156fba9e9c38b7cefe57b7ed713734).


dracc:
Looks good to me. :+1:

PR #531, opened 2022-01-08
hal: Remove deprecated Port I/O functions

thrimbor:
We deprecated these in May 2021 so I think it's ok to get rid of them now.

Alternatives are the kernel's `READ_PORT_BUFFER_*` functions and inline assembly.


mborgerson:
Preferably we avoid inline assembly in nxdk and nxdk apps. Since kernel provides these interfaces this is fine

PR #533, opened 2022-01-27
lwip: Update submodule to STABLE-2_1_3_RELEASE

Ryzee119:
Tested on the httpd samples. No issues identified


JayFoxRox:
That's probably not enough for lwip.

There are out-of-tree copies of config and source files from lwip in https://github.com/XboxDev/nxdk/tree/master/lib/net/nforceif
So if the lwip submodule is updated, ideally, the drivers should also be rebased, so we get the full set of options / don't keep dead options / implement all necessary functionality for the new version.

thrimbor:
2.1.3 is mostly a bugfix release, it doesn't look like any relevant options changes were made (apart from adding `LWIP_PBUF_CUSTOM_DATA`, which is for storing custom data in pbufs, which we don't use)

thrimbor:
I've skimmed the diff from 2.1.2 to 2.1.3, and nothing else jumped out at me. 2.1.3 does fix a double-free issue though which @Ryzee119 and I have been encountering while working on `nvnetdrv` (it's most likely this one https://github.com/lwip-tcpip/lwip/commit/b5e8ab6c150e14f2418ba99d8ab91df2f3869d66), so unless there is some issue reported in the meantime, I will most likely merge it this evening.

Ryzee119:
When looking into the network code I noticed upstream lwip master branch has a win32 port https://github.com/lwip-tcpip/lwip/tree/master/contrib/ports/win32

Ive got a prototype working on my own repo https://github.com/Ryzee119/nxdk/commits/lwipwin32. It almost works out of the box with nxdk's winapi, I just needed to stub some crypto and console input calls.

This may we worth investigating in the future to minimise maintenaince on nxdk's own sys_arch.c

PR #539, opened 2022-02-04
Addition of issue forms to NXDK

MasonT8198:
This will require enabling on the organization end, however this adds a bug report issue form to NXDK, and we can add forms for other forms of issues as well, this first form is based on the implementation of issue forms over at the xemu project, specifically this commit here (https://github.com/mborgerson/xemu/commit/cc6d19e0990ab94aa55632302afb76b22f3ca09d).

Documentation here: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

I've also added a config.yml file that disallows blank issues and also adds a link to the XboxDev discord server

I'm very free to further workshop these forms and perhaps add forms for different topics.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -2,4 +2,4 @@ blank_issues_enabled: false
 contact_links:
   - name: XboxDev
     url: https://discord.gg/WxJPPyz
-    about: Visit the discord server for questions about developing xbox software using nxdk.
+    about: Visit the discord server for questions about developing Xbox software using nxdk.
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: XboxDev
+    url: https://discord.gg/WxJPPyz
+    about: Visit the discord server for questions about developing xbox software using nxdk.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: XboxDev
+    url: https://discord.gg/WxJPPyz
+    about: Visit the discord server for questions about developing xbox software using nxdk.
```
JayFoxRox, line 5: nit: "Xbox", not "xbox"

JayFoxRox:
I don't think this should be merged.

- GitHub might not be a permanent host for XboxDev and this ties it more closely to GitHub.
- The template makes very little sense for nxdk, because it's not aimed at end-users.
- The quality of issues would not be improved by this template; most issues are fine, or they are incorrect in ways which this template wouldn't address.

Ryzee119:
> GitHub might not be a permanent host for XboxDev and this ties it more closely to GitHub.
This might be a minor inconvenience in the distance future, but should not hold back this change.

> The template makes very little sense for nxdk, because it's not aimed at end-users.
Please provide suggestions of what you would like to see otherwise this will just stall.

JayFoxRox:
> Please provide suggestions of what you would like to see otherwise this will just stall.

I don't think nxdk needs these templates at all, so this PR should be closed without replacement (in my opinion).
Issue quality is mostly fine and it's usually a problem with people not reading the pinned issues / not seeing issues in submodules.

Issue #537, opened 2022-02-03
READ_PORT_BUFFER_* WRITE_PORT_BUFFER_* and friends have wrong type hints


GXTX:
The typehints in `xboxkrnl.h` for these functions are wrong leading to warnings while building projects with them.

See my info script for correct usage.

thrimbor:
The types in the function signature match Microsoft's declaration in `wdm.h` and `miniport.h`, as well as ReactOS's declarations. The kernel code it applies to is functionally equivalent to ReactOS's implementation and byte-for-byte the same as Microsoft's implementation in `hal.dll` on Windows 7 x86.

Please clarify what you think is wrong.

GXTX:
Hi, please see my complaints in your software development channel on you discord.

thrimbor:
The chat history doesn't provide what I asked for. Given what I wrote above, please clarify what you think is wrong.

GXTX:
Hi @thrimbor, the chat convo I had with Matt about this issue should accurately describe the problem. I implore you to attempt to use these functions as their type hints suggest, your Xbox will not like it.

Due to reasons I unfortunately don't have time these days to help figure out why this is broken, but looking at my info script (perhaps I never shared this with you? other maintainers should have access and can give you a link), or again testing these functions yourself, you will see what is and isn't working.

Thanks for the future fix!

thrimbor:
Please refrain from opening issues if you are unwilling to even describe your actual problem.

Since this is not the first case of you demonstrating uncooperative behavior on XboxDev repositories, I feel obliged to inform you that continuing this pattern will possibly lead to getting blocked from interacting with these repositories in the future.

Should you have a legitimate problem, please feel free to create a more constructive issue, ideally after consulting a guide describing how to create a good GitHub issue.

GXTX:
> demonstrating uncooperative behavior

I seem to just have this issue with you. I'm not sure why. The conversation with Matt on Discord accurately displays what the issue is and what the correct types should be. I'm not in that server so I cannot give you the logs of it, however you are.

This is a legitimate issue, and it's unfortunate you've decided to close it instead of investigate as the maintainer. Hopefully this issue being left unresolved does not create confusion for future users of nxdk.

GXTX:
@thrimbor this issue was fixed in CXBX-R. Please consider looking at their changes and reevaluating this issue. Thanks.

dracc:
cxbx-reloaded/cxbx-reloaded#2318 for context.

[...]

PR #540, opened 2022-02-05
xboxkrnl: Align READ_PORT_BUFFER_ULONG sig with MS

abaire:
Addresses one of the points from #537.

MS documentation indicates that the `Count` parameter should be a `ULONG` rather than a `PULONG`: https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-read_port_buffer_ulong 

Thanks @GXTX for pointing out the inconsistency.


thrimbor:
Thanks!

PR #541, opened 2022-02-05
xboxkrnl: Add documentation for *_PORT_BUFFER_* methods

abaire:
Fixes part of #537

The type signature of these methods matches Microsoft but has caused confusion and is not entirely aligned with expectations around how IO space is generally referenced. Since we want to keep the type in sync with the [MS docs](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-read_port_buffer_uchar) and implementation, the decision was made to attempt to resolve this confusion via documentation.

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -1729,6 +1729,12 @@ XBAPI XBOX_KEY_DATA XboxEEPROMKey;
 
 XBAPI XBOX_KEY_DATA XboxAlternateSignatureKeys[];
 
+/**
+ * Writes a number of USHORT values from the specified port address into a buffer.
+ * @param Port The memory mapped I/O address of the port. E.g., `(PUSHORT)0x0cf8` for port 0x0CF8.
+ * @param Buffer Pointer to a buffer containing the values to be written.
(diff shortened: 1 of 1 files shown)


abaire:
@GXTX I'd appreciate it if you could take a look at this one and let me know if there's any wordsmithing that might improve readability.

thrimbor:
Documentation looks good to me (I'll leave it unmerged for now to give @GXTX time to respond), but it seems an accidental submodule downgrade slipped in :)

abaire:
> Documentation looks good to me (I'll leave it unmerged for now to give @GXTX time to respond), but it seems an accidental submodule downgrade slipped in :)

Yikes! Sorry about that, reverted.

GXTX:
I'd rather see the method change but maintainers see otherwise; I have no input on this.

thrimbor:
Thanks for the contribution!

JayFoxRox:
> I'd rather see the method change but maintainers see otherwise; I have no input on this.

We can't change it because this is an import signature and the compiler ABI might require it to be like this; if you want a different signature, then create a wrapper for it - but it probably shouldn't have to be in nxdk.
If the Windows API has it differently, then feel free to create a wrapper in the nxdk winapi abstraction.

PR #550, opened 2022-02-27
xboxkrnl: Add some necessary constants

thrimbor:
These two commits are split out from `nvnetdrv`, the new NIC driver Ryzee and I worked on. They are useful for reading EEPROM settings (such as the Ethernet address) and for waking up a thread using an event while also boosting thread priority (used by `nvnetdrv`'s IRQ processing).

I'm submitting these commits separately because they're also used by the network configuration API I'm working on that I plan to submit before `nvnetdrv` to make the transition easier.

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -3149,6 +3149,27 @@ XBAPI NTSTATUS NTAPI NtCreateFile
 #define FILE_EXISTS 0x00000004
 #define FILE_DOES_NOT_EXIST 0x00000005
 
+/* Priority increment values for events */
+#define ERESOURCE_INCREMENT 4
+#define EVENT_INCREMENT 1
+#define EVENT_PAIR_INCREMENT 1
+#define LPC_RELEASE_WAIT_INCREMENT 1
+#define IO_NO_INCREMENT 0
+#define IO_CD_ROM_INCREMENT 1
+#define IO_DISK_INCREMENT 1
+#define IO_PARALLEL_INCREMENT 1
+#define IO_VIDEO_INCREMENT 1
+#define IO_MAILSLOT_INCREMENT 2
+#define IO_NAMED_PIPE_INCREMENT 2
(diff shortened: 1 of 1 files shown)


lib/xboxkrnl/xboxkrnl.h
```diff
@@ -3149,6 +3149,27 @@ XBAPI NTSTATUS NTAPI NtCreateFile
 #define FILE_EXISTS 0x00000004
 #define FILE_DOES_NOT_EXIST 0x00000005
 
+/* Priority increment values for events */
+#define ERESOURCE_INCREMENT 4
+#define EVENT_INCREMENT 1
+#define EVENT_PAIR_INCREMENT 1
+#define LPC_RLEASE_WAIT_INCREMENT 1
```
Ryzee119, line 3156: RLEASE -> R**E**LEASE?

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -3149,6 +3149,27 @@ XBAPI NTSTATUS NTAPI NtCreateFile
...
+#define IO_NO_INCREMENT 0
+#define IO_CD_ROM_INCREMENT 1
+#define IO_DISK_INCREMENT 1
+#define IO_PARALLEL_INCREMENT 1
+#define IO_VIDEO_INCREMENT 1
+#define IO_MAILSLOT_INCREMENT 4
```
Ryzee119, line 3162: Most references online have IO_MAILSLOT_INCREMENT at 2

Ryzee119 (commented):
Looks good,
Not too sure where the ExSaveNonVolatileSetting offsets came from but they match CXBX's 👍

lib/xboxkrnl/xboxkrnl.h
thrimbor: Fixed.

lib/xboxkrnl/xboxkrnl.h
thrimbor: Fixed.

PR #553, opened 2022-03-01
winapi: Fix LTO optimizing away the fls_init constructor

### LINKED ISSUE #552: Building `NXDK_ONLY=y` with LTO creates xbe that crash
Building nxdk libs with LTO seem to generate xbes that crash/hang. Tested on hardware.

```
wsl@DESKTOP:~/Test/nxdk$ eval $(bin/activate -s) NXDK_ONLY=y LTO=y make -j8
[...]
wsl@DESKTOP:~/Test/nxdk$ cd samples/hello
wsl@DESKTOP:~/Test/nxdk/samples/hello$ eval $(../../bin/activate -s) LTO=y make
```

Building only the sample as LTO functions correctly.

I had hoped to use this to reduce xbe size (although it's not a real issue). The smallest size is 104KB with `-O3` on just the application, ~65KB with LTO on both libs and application, and ~49KB with LTO & `-O3` on both.

thrimbor:
This fixes a crash that happens when building `winapi` with LTO. LLVM detects there are no references to `__fls_init_p` (it is never called directly, only via iterating over the `.CRT$XXT` subsection which contains these special constructors)  and removes it, but this function pointer is necessary to initialize the variables used by `fls_register_thread`, which is run whenever a new thread is created. The uninitialized variables cause the latter function to cause a crash in kernel code.

The fix is to force the linker to include the symbol whenever the object file containing `fls_register_thread` is being included.

Fixes #552.

My initial checks with LLVM 13.0.1 suggested that this is only an issue with specific versions of LLVM, however, I've been able to reproduce it with this version now as well, suggesting that my initial test was flawed (potentially caused by leftover files).

### DIFF
--- a/lib/winapi/fiber.c
+++ b/lib/winapi/fiber.c
@@ -24,6 +24,7 @@ static __cdecl VOID fls_init (VOID)
     InitializeCriticalSection(&fls_lock);
     InitializeListHead(&fls_nodes_list);
 }
+#pragma comment(linker, "/include:___fls_init_p")
 __attribute__((section(".CRT$XXT"))) void (__cdecl *const __fls_init_p)(void) = fls_init;
 
 VOID fls_register_thread (VOID)


GXTX:
Thanks.

PR #555, opened 2022-03-04
hal: Relicense audio code to MIT

thrimbor:
I have managed to find the original authors and got their permission to relicense this under the MIT license (I can share proof of the conversations in a non-public channel on request).

I've combined the license change with adding proper attribution by using SPDX-tags.


Ryzee119:
Thanks thrimbor for taking the time to reach out to the authors and thanks to the authors for allowing this generous relicense.

Ive sighted the authors acceptance and will merge.

PR #556, opened 2022-03-04
winapi: Implement SetEndOfFile()

### LINKED ISSUE #471: `SetEndOfFile` is missing
Needed by a couple of projects, and looks easy to implement, so leaving it here, as a potential task.

kosmas12:
Closes #471

Please, excuse the failed Ubuntu CI on my repo - It's because my fork is really old.

### DIFF
--- a/lib/winapi/fileio.c
+++ b/lib/winapi/fileio.c
@@ -287,6 +287,32 @@ BOOL SetFilePointerEx (HANDLE hFile, LARGE_INTEGER liDistanceToMove, PLARGE_INTE
     }
 }
 
+BOOL SetEndOfFile (HANDLE hFile)
+{
+    // Get current position in file
+    LONG highPosDWORD = 0;
+    PLONG lpHighPosDWORD = &highPosDWORD;
+    DWORD lowPosDWORD = SetFilePointer(hFile, 0, lpHighPosDWORD, FILE_CURRENT);
+
+    // Combine low and high position DWORDs for full position
+    LONGLONG fullCurrentPos;
+    fullCurrentPos = highPosDWORD;
+    fullCurrentPos = fullCurrentPos << 32 | lowPosDWORD;
+
+
+    IO_STATUS_BLOCK ioStatusBlock;
+    FILE_END_OF_FILE_INFORMATION fileEOFInfo;
+    fileEOFInfo.EndOfFile.QuadPart = fullCurrentPos;
+
+    NTSTATUS status = NtSetInformationFile(hFile, &ioStatusBlock, &fileEOFInfo, sizeof(fileEOFInfo), FileEndOfFileInformation);
+    if (NT_SUCCESS(status)) {
+        return FALSE;
+    }
+    else {
+        return TRUE;
+    }
+}
+
 DWORD GetFileSize (HANDLE hFile, LPDWORD lpFileSizeHigh)
 {
     LARGE_INTEGER fileSize;
--- a/lib/winapi/fileapi.h
+++ b/lib/winapi/fileapi.h
@@ -28,6 +28,7 @@ BOOL SetFilePointerEx (HANDLE hFile, LARGE_INTEGER liDistanceToMove, PLARGE_INTE
 
 DWORD GetFileSize (HANDLE hFile, LPDWORD lpFileSizeHigh);
 BOOL GetFileSizeEx (HANDLE hFile, PLARGE_INTEGER lpFileSize);
+BOOL SetEndOfFile (HANDLE hFile);
 
 HANDLE FindFirstFileA (LPCSTR lpFileName, LPWIN32_FIND_DATAA lpFindFileData);
 BOOL FindNextFileA (HANDLE hFindFile, LPWIN32_FIND_DATAA lpFindFileData);


lib/winapi/fileio.c
```diff
@@ -287,6 +287,32 @@ BOOL SetFilePointerEx (HANDLE hFile, LARGE_INTEGER liDistanceToMove, PLARGE_INTE
...
+    fullCurrentPos = fullCurrentPos << 32 | lowPosDWORD;
+
+
+    IO_STATUS_BLOCK ioStatusBlock;
+    FILE_END_OF_FILE_INFORMATION fileEOFInfo;
+    fileEOFInfo.EndOfFile.QuadPart = fullCurrentPos;
+
+    NTSTATUS status = NtSetInformationFile(hFile, &ioStatusBlock, &fileEOFInfo, sizeof(fileEOFInfo), FileEndOfFileInformation);
+    if (NT_SUCCESS(status)) {
+        return FALSE;
```
thrimbor, line 309: lastError needs to get set here, you can check other functions to see how.

lib/winapi/fileio.c
```diff
@@ -287,6 +287,32 @@ BOOL SetFilePointerEx (HANDLE hFile, LARGE_INTEGER liDistanceToMove, PLARGE_INTE
     }
 }
 
+BOOL SetEndOfFile (HANDLE hFile)
+{
+    // Get current position in file
+    LONG highPosDWORD = 0;
+    PLONG lpHighPosDWORD = &highPosDWORD;
+    DWORD lowPosDWORD = SetFilePointer(hFile, 0, lpHighPosDWORD, FILE_CURRENT);
```
thrimbor, line 295: `SetFilePointer` is not an ideal choice as its handling of 64 bit offsets is awkward. `SetFilePointerEx` is better, but I'd prefer it this code used `NtQueryInformationFile` directly.

lib/winapi/fileio.c
```diff
@@ -287,6 +287,32 @@ BOOL SetFilePointerEx (HANDLE hFile, LARGE_INTEGER liDistanceToMove, PLARGE_INTE
...
+    FILE_END_OF_FILE_INFORMATION fileEOFInfo;
+    fileEOFInfo.EndOfFile.QuadPart = fullCurrentPos;
+
+    NTSTATUS status = NtSetInformationFile(hFile, &ioStatusBlock, &fileEOFInfo, sizeof(fileEOFInfo), FileEndOfFileInformation);
+    if (NT_SUCCESS(status)) {
+        return FALSE;
+    }
+    else {
+        return TRUE;
+    }
```
thrimbor, line 313: That's awkward formatting. Personally I prefer to keep the `return TRUE;` case outside of the if.

lib/winapi/fileio.c
```diff
@@ -287,6 +287,32 @@ BOOL SetFilePointerEx (HANDLE hFile, LARGE_INTEGER liDistanceToMove, PLARGE_INTE
...
+}
```
thrimbor, line 314: `FileAllocationInformation` needs treatment equivalent to `FileEndOfFileInformation` here.

thrimbor (changes requested):

lib/winapi/fileio.c
kosmas12: I was expecting this, but problem is, I don't know what to set the error to. Is there some kind of standard way we should follow to mimic Windows? Or can we just have our own error messages? Still not sure about the proper way to contribute to WinAPI code.

lib/winapi/fileio.c
kosmas12: I'll check that possibility out.

lib/winapi/fileio.c
```diff
@@ -287,6 +287,32 @@ BOOL SetFilePointerEx (HANDLE hFile, LARGE_INTEGER liDistanceToMove, PLARGE_INTE
...
+{
+    // Get current position in file
+    LONG highPosDWORD = 0;
+    PLONG lpHighPosDWORD = &highPosDWORD;
+    DWORD lowPosDWORD = SetFilePointer(hFile, 0, lpHighPosDWORD, FILE_CURRENT);
+
+    // Combine low and high position DWORDs for full position
+    LONGLONG fullCurrentPos;
+    fullCurrentPos = highPosDWORD;
+    fullCurrentPos = fullCurrentPos << 32 | lowPosDWORD;
```
JayFoxRox, line 300: Not sure if `<< 32` will work properly. You might have to use `<< 32LL`. Please verify

Also check operator precedence please.

lib/winapi/fileio.c
JayFoxRox: *(for future reviews: If you are aware of missing code / existing issues, please leave a comment in the code or, better yet, add a review comment yourself)*

lib/winapi/fileio.c
JayFoxRox: It's not just formatting, it's also required for some IDEs (or even compilers?)

thrimbor:
Closing this as https://github.com/XboxDev/nxdk/pull/583 has been merged

PR #557, opened 2022-03-06
Introduce a better log system

apfel:
This replaces `hal/debug.h` **completely** with a custom logging system built upon a registering system.

The code I pushed here is like 2 years old at this point (I had some stuff to deal with 🙃) and probably could go well with some cleanup, I mainly decided to add a draft PR so that people can take a look at this

All samples have been updated, but PDCLib requires a patch applied to assert.c to allow the nxdk to be compiled correctly (how should I handle this? it's a one line fix replacing `debugPrint` with `nxLogPrintf` lol)

tl;dr tell me your thoughts : )


apfel:
I implemented some level of TV logging, honestly can't decide whether to leave the font sizes to be hardcoded as 8x16 or not (do we still use 8x8 fonts today?)

apfel:
Alright, I'm reasonably confident that this PR is now ready for review.

Some elements might benefit from refining, the TV scrolling implementation doesn't care about the margins (although whether it should is doubtworthy tbh) and is honestly quite slow.

Ryzee119:
Thanks for putting this together. nxdk is in desperate need of a better logging system.

This will break probably every nxdk application is existence as they almost certainly use debugPrint somewhere :)
Perhaps we could add debugPrint wrapper in hal/debug.h to wrap nxLogPrint  with a depreciated warning so existing usages of debugPrint wont break. (Unless I missed it)

Before merge we should add SPDX license headers onto your work to match the latest nxdk standard.

I havent tested yet, but hopefully I can give it a shot asap.

apfel:
Breaking `debugPrint` was on purpose, to be honest - I wanted to make software break away from it and move onto `nxLogPrintf` instead.

I never really considered just deprecating `hal/debug.h` tbh, it'd probably be best to do that instead of immediately breaking the API there.

I'll make sure to add the SPDX headers.

Console scrolling on the TV is pretty slow tbh, it's just the CPU moving 4 pixels backwards at a time and clearing the last line at the moment.
There's probably a better way to do this, but I failed to come up with a solution myself as of right now.

mborgerson:
> Breaking `debugPrint` was on purpose, to be honest - I wanted to make software break away from it and move onto `nxLogPrintf` instead.
> 
> I never really considered just deprecating `hal/debug.h` tbh, it'd probably be best to do that instead of immediately breaking the API there.

I think adding a deprecation note would be easy and nice to do
 
> Console scrolling on the TV is pretty slow tbh, it's just the CPU moving 4 pixels backwards at a time and clearing the last line at the moment. There's probably a better way to do this, but I failed to come up with a solution myself as of right now.

The solution is to not scroll on a per-pixel basis. Instead, track which lines should be displayed (with some pointer/offset) then simply redraw the text on screen

apfel:
I've added SPDX headers to the files I created; I hope they're well formed.

I've not found much time recently to work on this, but I will replace the very badly performing scrolling code very soon ~~(at least I hope I do)~~.

lib/nxdk/log_udp.c
```diff
@@ -0,0 +1,142 @@
...
+        return nxLogResult_MaximumNumberReached;
+
+    default:
+        break;
+    }
+
+#ifdef NXDK_UDP_USE_INTERNAL_NETIF_IMPL
+    sys_sem_t init_complete;
+    sys_sem_new(&init_complete, 0);
+	tcpip_init(tcpip_init_done, &init_complete);
```
thrimbor, line 72: I don't think the logging code should do (de-) initialization of the network stack.

lib/usb/libusbohci_xbox/usbh_xbox.c
```diff
@@ -1,93 +1,94 @@
-#include <stdint.h>
-#include <stdarg.h>
-#include <windows.h>
-#include <xboxkrnl/xboxkrnl.h>
-#include <hal/debug.h>
-#include "usbh_config_xbox.h"
```
thrimbor, line 6: Something's not right with this diff. Maybe an accidental CRLF<->LF change?

samples/hello++/main.cpp
```diff
@@ -1,24 +1,26 @@
...
 #include <string>
 #include <vector>
+
+#include <hal/video.h>
+#include <nxdk/log_console.h>
 #include <windows.h>
 
 int main(void) {
-  XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
+    XVideoSetMode(640, 480, 32, REFRESH_DEFAULT);
```
thrimbor, line 9: Indentation should be kept / the change separated.

lib/net/pktdrv/pktdrv.c
```diff
@@ -1,944 +1,944 @@
-//XBOX low level network interface driver (packet driver)
-//=======================================================
-//Minimum capabilites : send and receive ethernet packets
```
thrimbor, line 3: Something's not right with this diff. Maybe an accidental CRLF<->LF change?

samples/httpd/main.c
```diff
@@ -85,20 +85,17 @@ int main(void)
...
-	debugPrint("\n");
-	debugPrint("IP address.. %s\n", ip4addr_ntoa(netif_ip4_addr(g_pnetif)));
-	debugPrint("Mask........ %s\n", ip4addr_ntoa(netif_ip4_netmask(g_pnetif)));
-	debugPrint("Gateway..... %s\n", ip4addr_ntoa(netif_ip4_gw(g_pnetif)));
-	debugPrint("\n");
+	nxLogPrintf("\nIP address.. %s\nMask........ %s\nGateway..... %s\n\n", ip4addr_ntoa(netif_ip4_addr(g_pnetif)), ip4addr_ntoa(netif_ip4_netmask(g_pnetif)), ip4addr_ntoa(netif_ip4_gw(g_pnetif)));
 
 	http_server_netconn_init();
-	while (1) NtYieldExecution();
+	while (1) SwitchToThread();
```
thrimbor, line 97: I'm fine with this change, but it should not be part of this PR as it is unrelated to the functionality it adds.

samples/usb/main.c
```diff
@@ -1,39 +1,39 @@
-#include <hal/debug.h>
-#include <hal/video.h>
-#include <windows.h>
```
thrimbor, line 3: Something's not right with this diff. Maybe an accidental CRLF<->LF change?

thrimbor (changes requested):
I like the general direction this is going in, but the code needs some cleaning up: unrelated / whitespace changes reverted, formatting changes reverted, formatting of the new code changed to fit the pre-existing (always use `{` and `}` with `if`, `switch` label indentation, spaces instead of tabs, etc)

The commits may also need some restructuring to ensure that each commit hash represents a functional repository state. At the moment, the first commit removes the old debug print code, which will make manual intervention necessary when this commit is encountered with automated bisecting. Fixes should also be integrated into the commits that added the code that was fixed.

In terms of functionality, what I would like to see is multiple log channels. `nxLogPrint` could get a channel id as the first parameter, with 0 being the equivalent of stdout, 1 being the equivalent of stderr, and the rest being free for custom use, and it would be possible to attach different handlers to all the channels. This would also allow us to switch to using the more-standard `printf` and `fprintf(stderr, ...)` ways of printing stuff - that would require changes to PDCLib obviously, so it might make sense for this PR to only add the new functionality and mark the old one as deprecated, then a separate PR for PDCLib will be necessary, followed by another one that changes all the places that output log messages.

lib/nxdk/log_udp.c
apfel: This was mainly for projects that don't use networking on their own but don't want to use the console log for whatever reason, so that they don't have to initialize it themselves.

samples/usb/main.c
apfel: Looks fine to me, all that's replaced is `hal/debug.h` with `nxdk/log_console.h`

thrimbor: No, the diff replaces every single line in the whole file.

lib/nxdk/log_console.c
```diff
@@ -0,0 +1,145 @@
...
+                switch (video_mode.bpp)
+                {
+                case 32:
+                    *((uint32_t*)(framebuffer + fb_offset)) = (log_console_font[message[i] * ((8 + 7) / 8) * 16 + font_offset]) & mask ? 0x00FFFFFF : 0x00000000;
+                    fb_offset += sizeof(uint32_t);
+                    break;
+
+                case 16:
+                case 15:
+                    *((uint16_t*)(framebuffer + fb_offset)) = (log_console_font[message[i] * ((8 + 7) / 8) * 16 + font_offset]) & mask ? (video_mode.bpp == 16 ? 0xFFFF : 0x7fff) : 0x0000;
```
dracc, line 61: Very long line, please split.

lib/nxdk/log_udp.c
JayFoxRox: Maybe a system like the nxdk automount stuff for networking instead?
I agree with thrimbor that most libs should not mess up the network state.

JayFoxRox:
Here are some random thoughts:

> In terms of functionality, what I would like to see is multiple log channels. 

API wise I don't care how log stream separation works.
An easy way to hook up stdout / stderr to these logging libraries could be independent (and I consider it very important that we get some way to register stdout / stderr / file-io handlers without necessarily going through kernel drivers).
However, my point is: API does not have to be matching that of `fprintf` / `write`.

So, there are some design alternatives to this, too:
- There could be separate calls to bump the logger into a certain state (state-machine).
- The logger could also be instanced (object oriented).
- Escape sequences in log messages to re-route signal (poor, do not do this-approach)

Potentially being able to respecting ANSI escape codes and something like `terminfo` in future would be a bonus though, so we can more easily hook up existing logging code. Not sure if this is a goal for this library though.

> [...] font sizes to be hardcoded as 8x16 or not (do we still use 8x8 fonts today?)

I'd prefer font-sizes to be not hardcoded at all.
We use whatever font sizes fit the task at hand. It depends on wether I use the logging as a crude UI for end-users on a TV, or wether I run this on an emulator or real machine for development on a sharp display, with a ton of debug output.
With the old `debugPrint`, my SWE1R port logs more than 30k lines of output and I try to keep as much as possible of it on the screen at once.

---

That said, this lib doesn't have to be a magical unicorn that does everything either.
Minimalism might not be bad; separate logging libraries could be made if we have an easy way to reroute file streams elsewhere / API to identify terminals (so applications know how many symbols fit on the screen + simulate cursor position).

Note that I mostly focus on the TV side display, but I do understand that file-logging and network-logging are also wanted.
In terms of requirements, I think we should learn from existing implementations / backends for stdout / stderr (like Unix terminals, which can stream to file / network / printer, while also being able to display nicely).

JayFoxRox:
I think this PR should be mentioned somewhere on the wiki and potentially be moved into a lib, so the code and effort isn't lost.

PR #564, opened 2022-03-17
pbkit: Rename lib to libpbkit

antangelo:
Renaming the pbkit lib output to libpbkit. Since it isn't part of nxdk itself, it should be distinct.


JayFoxRox:
nit: Commit title should start with "pbkit: Rename" (convention is uppercase after colon)

PR #565, opened 2022-03-19
cxbe: Simplify license statements and attribution with SPDX

thrimbor:
This changes cxbe's license and copyright statements to use SPDX like we do for other code already, which reduces verbosity and is machine readable (there are tools to check license conformity etc.).

### DIFF
--- a/tools/cxbe/Common.cpp
+++ b/tools/cxbe/Common.cpp
@@ -1,6 +1,7 @@
-// Licensed under GPLv2 or (at your option) any later version.
-// Copyright (C) 2019 Jannik Vogel
-// Copyright (C) 2002-2003 Aaron Robinson <caustik@caustik.com>
+// SPDX-License-Identifier: GPL-2.0-or-later
+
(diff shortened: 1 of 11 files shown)


tools/cxbe/Common.cpp
```diff
@@ -1,6 +1,7 @@
-// Licensed under GPLv2 or (at your option) any later version.
-// Copyright (C) 2019 Jannik Vogel
-// Copyright (C) 2002-2003 Aaron Robinson <caustik@caustik.com>
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// SPDX-FileCopyrightText: 2002-2003 Aaron Robinson <caustik@caustik.com>
+// SPDX-FileCopyrightText: 2019 Jannik Vogel
```
JayFoxRox, line 4: Not sure how easy it is, but feel free to put my changes under MIT license (or even CC0)

thrimbor: I'll keep that in mind, but since this code isn't included in xbes I don't think I'll go for relicensing cxbe.

PR #567, opened 2022-03-20
xboxkrnl: Improve licensing info

thrimbor:
This replaces the non-standard and missing license and copyright statements in `xboxkrnl` files with SPDX.

### DIFF
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -1,9 +1,13 @@
+// SPDX-License-Identifier: CC0-1.0
+
+// SPDX-FileCopyrightText: 2017-2022 Stefan Schmidt
+// SPDX-FileCopyrightText: 2018-2021 Jannik Vogel
+// SPDX-FileCopyrightText: 2018 Sean Koppenhafer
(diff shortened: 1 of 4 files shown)


lib/xboxkrnl/xboxkrnl.h
```diff
@@ -1,9 +1,14 @@
+// SPDX-License-Identifier: CC0-1.0
+
+// SPDX-FileCopyrightText: 2017-2022 Stefan Schmidt
+// SPDX-FileCopyrightText: 2018-2021 Jannik Vogel
+// SPDX-FileCopyrightText: 2018 Sean Koppenhafer
+// SPDX-FileCopyrightText: 2022 Erik Abair
+
 /**
  * @file xboxkrnl.h
  * @author Stefan Schmidt
```
JayFoxRox, line 10: Shall we remove this, as to not be confused with licensing information?
We could also rebrand this as "XboxDev Community" or something if we want to keep the tag.

I'm fine with keeping as-is though

thrimbor: Hm, removing is probably the best option. Done.

PR #568, opened 2022-03-20
cg: Add NVIDIA license info

thrimbor:
This adds the copyright and license info for the Cg binaries we distribute.

### DIFF
--- a/LICENSES/LicenseRef-NVIDIA-CG.txt
+++ b/LICENSES/LicenseRef-NVIDIA-CG.txt
@@ -1,0 +1,138 @@
+License For Customer Use of NVIDIA Software
+
+IMPORTANT NOTICE -- READ CAREFULLY: This License For Customer Use of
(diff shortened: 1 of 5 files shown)


LICENSES/LicenseRef-NVIDIA-CG.txt
```diff
@@ -0,0 +1,138 @@
+License For Customer Use of NVIDIA Software
```
JayFoxRox, line 1: *About filename:* Is there some standard which prefixes the filename with `LicenseRef-`? I'd have expected it in the SPDX tag, but not in the filename itself.

thrimbor: https://reuse.software/faq/#custom-license mentions this. I find it a bit awkward as well, I did it incorrectly at first and had `reuse lint` complain about it.

PR #571, opened 2022-03-20
Add license info to smaller files

thrimbor:
This adds SPDX-style licensing info to some of our smaller files that don't warrant a separate PR, such as the build scripts.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,11 @@
+# SPDX-License-Identifier: CC0-1.0
+
+# SPDX-FileCopyrightText: 2016-2019 Matt Borgerson
+# SPDX-FileCopyrightText: 2018 Lucas Jansson
+# SPDX-FileCopyrightText: 2019-2020 Stefan Schmidt
+# SPDX-FileCopyrightText: 2021 Margen67
+# SPDX-FileCopyrightText: 2021 Ryan Wendland
+
 [submodule "tools/extract-xiso"]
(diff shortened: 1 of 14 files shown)


.gitmodules
```diff
@@ -1,3 +1,11 @@
+# SPDX-License-Identifier: CC0-1.0
+
+# SPDX-FileCopyrightText: 2015 espes
+# SPDX-FileCopyrightText: 2016 Matt Borgerson
+# SPDX-FileCopyrightText: 2017 Luke Usher
+# SPDX-FileCopyrightText: 2018-2019 Stefan Schmidt
+# SPDX-FileCopyrightText: 2019 Jannik Vogel
```
JayFoxRox, line 7: Does this file really need copyright info? Is there even anything non-factual that *could* potentially be copyrighted?

thrimbor: Probably not, but that's part of the reason why I made it explicitly CC0, so that no-one will have to worry about that (as well as making `reuse lint` and similar tools happy).

Ryzee119: Agree license info isnt necessary here but happy to leave it for lint tools.
Where is this author list from? Commit history has different contributors than listed.

thrimbor: Oops, that was the commit history of `.gitignore` (mixed it up in the commit title, too). Fixed it now, and added another commit that adds license and copyright info for `.gitignore`.

PR #573, opened 2022-04-10
winapi: Add Mutex functions

Ryzee119:
Add CreateMutex and ReleaseMutex synchapi.h functions

I did a basic test with this code:

```
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>
#include <assert.h>

static HANDLE mutex;
static HANDLE thread;
static int mutex_variable;
DWORD WINAPI increment_variable(LPVOID arg)
{
    while (1)
    {
        WaitForSingleObject(mutex, INFINITE);
        //We have a lock on mutex_variable. Do a couple independent non atomic operations
        //with a big delay between them to test the mutex lock.
        mutex_variable++;
        //Put thread to sleep for a while. If we access mutex_variable from another thread during this sleep,
        //it will be an odd number! Shouldnt happen if mutex is working!
        Sleep(1000);
        mutex_variable++;
        ReleaseMutex(mutex);
    }
    return TRUE;
}

int main(void)
{

    mutex_variable = 0;
    mutex = CreateMutex(NULL, FALSE, NULL);
    thread = CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)increment_variable, NULL, 0, NULL);
    while (1)
    {
        WaitForSingleObject(mutex, INFINITE);
        //Printed variable should always be an even number
        debugPrint("mutex_variable: %d\n", mutex_variable);
        assert((mutex_variable % 2) == 0);
        ReleaseMutex(mutex);
    }
    return 0;
}
```

### DIFF
--- a/lib/winapi/sync.c
+++ b/lib/winapi/sync.c
@@ -2,6 +2,7 @@
 
 // SPDX-FileCopyrightText: 2019-2020 Stefan Schmidt
 // SPDX-FileCopyrightText: 2020 Jannik Vogel
+// SPDX-FileCopyrightText: 2022 Ryan Wendland
 
 #include <synchapi.h>
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


thrimbor:
Looks good! I've allowed myself to add you to the copyright info, and I made a small change that renames `CreateMutex` to `CreateMutexA`, and provides the `CreateMutex` name via a macro (it's what we do for the other functions that have unicode versions on Windows).

I'll let CI run and then merge.

PR #576, opened 2022-04-22
Make pkg-config work within CMake projects

sharkwouter:
This is needed if you want to use the FindPkgConfig package and find packages with the ``pkg_search_module`` command.


dracc:
Looks good to me.

PR #577, opened 2022-04-22
Fail CMake if NXDK_DIR is not set

sharkwouter:
If it's missing, it's going to fail anyway. This makes it more visible to the user what is going on.


share/toolchain-nxdk.cmake
```diff
@@ -1,4 +1,8 @@
-set(NXDK_DIR $ENV{NXDK_DIR})
+if(DEFINED ENV{NXDK_DIR})
+    set(NXDK_DIR $ENV{NXDK_DIR})
+else()
+    message(FATAL_ERROR "The environment variable NXDK_DIR needs to be defined.")
+endif()
```
JayFoxRox, line 5: [Please check history](https://github.com/XboxDev/nxdk/commits/master) for commit-style.

sharkwouter: I see, I redid the commit for both PRs.

PR #580, opened 2022-05-15
lib: Add FTP lib as a submodule

dracc:


JayFoxRox:
I'm against this for reasons described in response to [thrimbors thoughts on that matter](https://github.com/XboxDev/nxdk/pull/418#issuecomment-752165932): see https://github.com/XboxDev/nxdk/pull/418#issuecomment-822691345 (which also affects new libs, not only ports)

I think we should decentralize the lib code and have a prober build-system.
Temporarily, I think it's best if user submodule your lib it in their project for now?

PR #582, opened 2022-06-02
video: Add additional video defines

Ryzee119:
I had a need to read the widescreen video flag set by MSDash and realised we did not have this defined.

I took the liberty of adding some others and removed some of the magic numbers in video.c


thrimbor:
Looks good!

PR #583, opened 2022-06-19
winapi: Implement SetEndOfFile()

### LINKED ISSUE #471: `SetEndOfFile` is missing

thrimbor:
In agreement with @kosmas12 this supersedes #556. The code itself is rather old (May 2021, for some reason I didn't upstream it and forgot about it), I just rebased and tested it today.

Closes #471 

Simple test code (this will truncate the file `test.bin` in the same directory as the XBE, I tested with a 1MiB file full of zeroes):
```C
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>

int main(void)
{

    HANDLE h = CreateFile("D:\\test.bin", GENERIC_ALL, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
    if (h == INVALID_HANDLE_VALUE) {
        debugPrint("Error: %x\n", GetLastError());
        Sleep(10000);
        return 0;
    }

    SetFilePointer(h, 512*1024, NULL, FILE_BEGIN);
    SetEndOfFile(h);
    CloseHandle(h);

    debugPrint("Done!\n");
    Sleep(10000);

    return 0;
}
```


JayFoxRox:
The sample above only talks about truncation. Did you also test extending a file?
Did you test a mix of operations, say, truncating and then immediately extending again: will the section be zero filled or does the old content remain somehow? I'd assume zero'ing, but I'm not sure how Windows reacts, or what the Xbox kernel does.

thrimbor:
Extending filled the file with zeroes, truncating and immediately extending without closing the file seems to preserve the content, closing the file after truncating and reopening it before extending seems to do the same, but I wouldn't rely on that.

Ryzee119:
I tested this locally on my xbox and works as expected.

Win32 docs indicate that extending the file results in undefined data in the extended area:
`If the file is extended, the contents of the file between the old end of the file and the new end of the file are not defined.`

I can merge this

PR #584, opened 2022-06-21
jsoncpp: Add jsoncpp 1.9.5

sharkwouter:
Jsoncpp is a library for parsing JSON in C++. It's very portable and does not require any Xbox specific patches to work. I personally use it a lot.

This PR includes a sample which prints a JSON value from a file.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,3 +42,6 @@
 [submodule "lib/usb/libusbohci"]
 	path = lib/usb/libusbohci
 	url = https://github.com/XboxDev/libusbohci.git
+[submodule "lib/jsoncpp/jsoncpp"]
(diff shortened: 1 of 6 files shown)


Ryzee119:
This is very cool. Thanks for your effort.

I understand the goal of nxdk is to improve its package manager to make including such libraries much eaiser. (reduce maintainer burden etc). This library would be an ideal candidate for that in the future.

Would you be happy to include this work in the wiki instead? https://github.com/XboxDev/nxdk/wiki/Projects-using-nxdk

RadWolfie:
Please close this pull request since it's not going to be upstream plus there are various json libraries to choose from.

PR #585, opened 2022-06-22
pbkit: Support non-standard NVCLK

### LINKED ISSUE #536: pbkit quits if the nvclk freq is not default
It shouldn't do this.

GXTX:
Fixes #536 

With the release of [software based overclocks](https://github.com/GXTX/XboxOverclock), and hardware modding becoming more popular I think pbkit should support any given frequency just as XDK does. A stock Xbox will run fine at ~250MHz and as low as ~25MHz.

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -2808,14 +2808,36 @@ int pb_init(void)
 
     if (mdiv)
     {
-        //Xtal in Xbox is at 16.666 Mhz but we want 31.25Mhz for GPU...
-        if (((DW_XTAL_16MHZ*ndiv)/(odiv<<pdiv))/mdiv!=233333324)
-        {
-            //This PLL configuration doesn't create a 233.33 Mhz freq from Xtal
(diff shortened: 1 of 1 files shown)


Ryzee119:
Looks pretty good, tested on my 1.6 with stock clocks and didnt break anything.
Got this output:
```
input freq : 233338Hz
num        : 58334
den        : 7812
timer freq : 31248Hz
```
Output matches the original hardcoded value of ~31.25Mhz

GXTX:
Thanks for the review. ^

Ernegien:
Closing due to lack of MIT consent.

GXTX:
The code is taken directly from Linux, I cannot re-license Linux.

JayFoxRox:
Could be rewritten to simplify fractions using a GCD, or you can find a source for fraction simplification using this algorithm to ensure it's factual / non-copyrightable (while also providing a second source).

I think it *is* already factual, although I don't fully understand the choice of the constant 32 and 5.

PR #586, opened 2022-07-01
bin/nxdk-pkg-config: Fix bashism

glebm:
It runs as `sh` and so should be POSIX shell compatible.


thrimbor:
Change looks good, I've allowed myself to modify the commit title to match our convention. Thanks!

PR #590, opened 2022-07-01
CMake toolchain: Fix pkg-config support

glebm:
`PKG_CONFIG_EXECUTABLE` must be a cache variable, otherwise only the first call to `find_program(PkgConfig)` succeeds.

Follow-up to #576


thrimbor:
Change looks good, I've allowed myself to modify the commit title to match our convention. Thanks!

PR #601, opened 2022-07-02
lib/fmt: New library

### LINKED ISSUE #594: Port libfmt
Several of the winapi issues I filed today are coming from [libfmt](https://github.com/fmtlib/fmt), a rather ubiqitous C++ library these days.

These issues happen because NXDK defines `_WIN32`, which makes the library assume WIN32 API support (and rightly so).

Should NXDK stop defining `_WIN32`? If not, should a patched libfmt be available as one of the SDK libraries?

glebm:
libfmt/fmt requires some patching to make it work on NXDK

I've done the patching in my fork https://github.com/glebm/fmt/tree/nxdk

It'd be better to have the fork be in the XboxDev org but I don't have permissions to set it up (please feel free to do so)

Fixes #594

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,3 +42,7 @@
 [submodule "lib/usb/libusbohci"]
 	path = lib/usb/libusbohci
 	url = https://github.com/XboxDev/libusbohci.git
+[submodule "lib/fmt/fmt"]
+	path = lib/fmt/fmt
+	url = https://github.com/glebm/fmt.git
+	branch = nxdk
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,7 @@ QUIET=>/dev/null
 endif
 
 DEPS := $(filter %.c.d, $(SRCS:.c=.c.d))
+DEPS += $(filter %.cc.d, $(SRCS:.cc=.cc.d))
 DEPS += $(filter %.cpp.d, $(SRCS:.cpp=.cpp.d))
 
 $(OUTPUT_DIR)/default.xbe: main.exe $(OUTPUT_DIR) $(CXBE)
@@ -127,6 +128,10 @@ endif
 %.obj: %.cpp
 	@echo "[ CXX      ] $@"
 	$(VE) $(CXX) $(NXDK_CXXFLAGS) $(CXXFLAGS) -MD -MP -MT '$@' -MF '$(patsubst %.obj,%.cpp.d,$@)' -c -o '$@' '$<'
+
+%.obj: %.cc
+	@echo "[ CXX      ] $@"
+	$(VE) $(CXX) $(NXDK_CXXFLAGS) $(CXXFLAGS) -MD -MP -MT '$@' -MF '$(patsubst %.obj,%.cc.d,$@)' -c -o '$@' '$<'
 
 %.obj: %.c
 	@echo "[ CC       ] $@"
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -8,3 +8,4 @@ include $(NXDK_DIR)/lib/xboxrt/Makefile
 include $(NXDK_DIR)/lib/zlib/Makefile
 include $(NXDK_DIR)/lib/libpng/Makefile
 include $(NXDK_DIR)/lib/libjpeg/Makefile
+include $(NXDK_DIR)/lib/fmt/Makefile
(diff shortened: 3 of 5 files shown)


AJenbo:
afaik this is also part of C++20 so devs would start to expect it afaik?

dracc:
> This is also part of C++20 so devs would start to expect it afaik?

libcxx already supports C++20, doesn't it? In that case, this library would be extraneous. :)

glebm:
`fmt` implements a lot more than what's in C++20.

Both DevilutionX and SDL_audiolib use it and we do not plan to migrate to `std::format` anytime soon as we need to support a language level somewhere around C++14 (gcc 6.3).

JayFoxRox:
Thanks for the work; however, I think this port should not be a part of the nxdk repository.

See:
- https://github.com/XboxDev/nxdk/pull/584#issuecomment-1173026976
- https://github.com/XboxDev/nxdk/pull/418#issuecomment-822691345
  - Also enforced in https://github.com/XboxDev/nxdk/pull/580#issuecomment-1127801877

This PR, for example, has to add a generic rule for *.cc, which probably should not be part of the nxdk build-system.
We have been abusing the old nxdk Makefile for way too long.

---

We should go forward with adding better package support or submodule / install dependencies from projects until nxdk has a package manager.
DevolutionX (for which this libfmt is meant) already has a `brewfile` and install instructions to install deps for most platforms - why not include libfmt install instructions for nxdk?

Upstreaming of packages is still a good goal of course, so the community benefits from any ports.
Hence, it makes sense to [list new dependencies on the wiki](https://github.com/XboxDev/nxdk/wiki/Projects-using-nxdk#libraries), until we have a better solution (~package-repository + package-manager)

Adding more packages to nxdk only complicates maintenance of nxdk and makes it harder to create a clean package manager (especially with old projects depending on more and more default packages or even hardcoded paths within nxdk).

---

Packages which should continue to live in nxdk (for now) are drivers (such as pbkit) and low level standard APIs which mimic the OS / userspace APIs (win32 / POSIX / ...).
Inclusion of SDL2 is an edge-case already and we mostly include it to standardize APIs until we settle on underlying drivers (as we had no stable API for audio and input, it was easier to recommend SDL2 as stable API).

Any API which is purely optional or just implementing logic should be avoided. I consider libfmt such a library.
Other libraries which should be removed from the core set of nxdk libraries are zlib, libjpeg and libpng.

glebm:
Makes sense.

NXDK does not implement enough Windows APIs to compile libfmt, so the NXDK version of libfmt requires heavy patching.

We can switch over to the patched version of libfmt in DevilutionX.

Can you fork libfmt to XboxDev/fmt and give me access?
I'll push our patch there to the `nxdk` branch so that everyone can benefit.

JayFoxRox:
> Can you fork https://github.com/fmtlib/fmt to XboxDev/fmt and give me access?

I personally think that it's also fine to keep the libfmt port on a personal account for now.
I'm concerned with diffusion of responsibility while also actively promoting libfmt as an official XboxDev port which *should* imply a maintained port.
However, I don't think we have enough people in XboxDev to keep a libfmt port up-to-date (we already struggle to keep SDL2 and libcxx updated - they are outdated by years).
Clearly we need a better solution to keep track of all these libraries (and appoint responsible people or encourage updates to libraries).

However, I'm mostly inactive, so I'll let other XboxDev maintainers decide.

AJenbo:
We build the NXDK version of DevilutionX multiple times each day, we would be happy to maintain the libfmt port in NXDK and update it as we bump our requirements.

glebm:
libfmt no longer requires patching as of https://github.com/fmtlib/fmt/pull/3636

Only a couple of options need to be set.

E.g. when including libfmt as a CMake subproject:

```cmake
set(FMT_OS OFF) # required for nxdk
include(FetchContent)
FetchContent_Declare(libfmt
  URL https://github.com/fmtlib/fmt/archive/a8a73da7e44e26d7c18d752976522eff7a21e0bf.tar.gz
  URL_HASH MD5=4ddd617a0921e919f1c2e6ab7c36c349
)
FetchContent_MakeAvailable(libfmt)
target_compile_definitions(fmt PUBLIC FMT_WINDOWS_NO_WCHAR) # required for nxdk
```

PR #605, opened 2022-07-14
nxdk: Fix DNS server overriding

thrimbor:
This fixes the DNS overriding behavior in the nxdk network initialization API. Despite being called `numdns`, `dns_setserver`'s first parameter is *not* a count, it's an index. This resulted in DNS overriding having no actual effect.
This change fixes that, and also only overrides the setting when the specified IP is valid.

### DIFF
--- a/lib/nxdk/net.c
+++ b/lib/nxdk/net.c
@@ -42,7 +42,7 @@ int nxNetInit(const nx_net_parameters_t *parameters)
     ip4_addr_t ipaddr, netmask, gateway;
     bool ipv4_dhcp;
     ip_addr_t dns[2];
-    bool dns_override = false;
+    memset(dns, 0, sizeof(dns));
 
     if (!parameters || parameters->ipv4_mode == NX_NET_AUTO) {
         nxdk_network_config_sector_t configSector;
@@ -79,7 +79,6 @@ int nxNetInit(const nx_net_parameters_t *parameters)
                                  (configSector.manual.secondaryDns >> 24 & 0xff));
             dns[0].type = IPADDR_TYPE_V4;
             dns[1].type = IPADDR_TYPE_V4;
-            dns_override = true;
         }
     } else if (parameters->ipv4_mode == NX_NET_DHCP) {
         ipv4_dhcp = true;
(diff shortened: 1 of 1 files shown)


lib/nxdk/net.c
```diff
@@ -152,12 +151,14 @@ int nxNetInit(const nx_net_parameters_t *parameters)
...
             dns[1].type = IPADDR_TYPE_V4;
-            dns_override = true;
         }
     }
 
-    if (dns_override) {
-        dns_setserver(2, dns);
+    if (dns[0].u_addr.ip4.addr != 0) {
+        dns_setserver(0, &dns[0]);
+    }
```
JayFoxRox, line 159: What if the user had previously set a DNS server, and now wants to deactivate it by setting it to all-zero? Could that be a realistic use-case?

Also affects `(parameters->ipv4_dns1 != 0 || parameters->ipv4_dns2 != 0)` check above.

thrimbor: If the user wants to change the DNS settings it's best to do that via the config sector functions. The idea behind allowing overriding here is to allow applications to force lwIP to use a specific DNS server, no matter what the Xbox was configure for.

We could extend this API in the future to allow changing settings after initialization, then a program could dynamically remove or add DNS servers (now that I think about it, this might be what you meant? Initializing with custom DNS server IPs, and then removing them later, falling back to the settings from the config sector?). We'll probably want to keep a copy of the settings or re-read the settings from the config sector in those functions.

GXTX: Is this in the best interest of the user? I'd be quite annoyed if I found an application not using the servers I set. Such an application could have malicious intent and allow for rebinding.

thrimbor: There is nothing that nxdk can or could do to keep developers from implementing malicious behavior, and an application not using your preferred DNS server will be the least of your worries in case you run malware.
That said, allowing software to use the DNS server of their choice is common, and there are legitimate usage scenarios (such as testing for DNS hijacking). Not to mention that this feature is always there, this PR is just about fixing a bug in our wrapping API.

PR #608, opened 2022-07-18
nxdk: Add `extern "C"` in configsector.h

thrimbor:
I seem to have forgotten to add the usual `extern "C"` to this header back when I wrote it, this fixes that so it can be used and the declared function properly linked again from C++.


thrimbor:
I'm self-merging this, as it's a low-risk change, tested, and blocks something I'm working on atm.

PR #609, opened 2022-07-19
nxdk: Implement support for saving the network config sector

thrimbor:
This implements saving the network config sector. I tested it with the following code, and the resulting config sector was accepted and correctly decoded by the 5960 dash:

```c
nxdk_network_config_sector_t config;
nxLoadNetworkConfig(&config);
config.manual.secondaryDns = 0xF0F0F0F0;
nxSaveNetworkConfig(&config);
```

### DIFF
--- a/lib/nxdk/configsector.c
+++ b/lib/nxdk/configsector.c
@@ -31,6 +31,10 @@ typedef struct
     DWORD footer;
 }  __attribute__ ((packed)) NXDK_CONFIG_SECTOR;
 
+_Static_assert(sizeof(NXDK_CONFIG_SECTOR) == NXDK_HDD_SECTOR_SIZE, "NXDK_CONFIG_SECTOR size mismatch");
+_Static_assert(sizeof(nxdk_network_config_sector_t) == NXDK_CONFIG_DATA_SIZE, "nxdk_network_config_sector_t size mismatch");
+
+#define NXDK_NETWORK_CONFIG_HEADERV2 'XBV2'
 #define NXDK_NETWORK_CONFIG_FOOTER 'XBCP'
 
 HANDLE nxOpenConfigPartition ()
@@ -79,8 +83,6 @@ DWORD nxNetworkConfigChecksum(NXDK_CONFIG_SECTOR *sectorData, UINT size)
 bool nxLoadNetworkConfigSector(HANDLE hPartition, UINT sectorIndex, BYTE *buffer, UINT size)
 {
     NXDK_CONFIG_SECTOR configSector;
-
-    _Static_assert(sizeof(configSector) == 512, "NXDK_CONFIG_SECTOR size mismatch");
 
     assert(hPartition != INVALID_HANDLE_VALUE);
     if (hPartition == INVALID_HANDLE_VALUE) {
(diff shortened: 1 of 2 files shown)


lib/nxdk/configsector.c
```diff
@@ -147,3 +148,66 @@ bool nxLoadNetworkConfig(nxdk_network_config_sector_t *config)
...
+    NXDK_CONFIG_SECTOR configSector;
+
+    _Static_assert(sizeof(configSector) == 512, "NXDK_CONFIG_SECTOR size mismatch");
+
+    assert(hPartition != INVALID_HANDLE_VALUE);
+    if (hPartition == INVALID_HANDLE_VALUE) {
+        return false;
+    }
+
+    assert(sectorIndex < NXDK_CONFIG_SECTOR_COUNT);
```
Ryzee119, line 163: assert here, then a check just below too? Should we remove the assert?

lib/nxdk/configsector.c
```diff
@@ -147,3 +148,66 @@ bool nxLoadNetworkConfig(nxdk_network_config_sector_t *config)
...
+    }
+
+    return true;
+}
+
+bool nxSaveNetworkConfig(const nxdk_network_config_sector_t *config)
+{
+    HANDLE hPartition = nxOpenConfigPartition();
+
+    assert(hPartition != INVALID_HANDLE_VALUE);
```
Ryzee119, line 199: assert here, then a check just below. Should we remove the assert?

lib/nxdk/configsector.c
thrimbor: That's on purpose - if this fails, you have a bug in your code, so when it happens in a non-debug build, it should gracefully fail, but in a debug build, I prefer stuff problems like these to be put right in my face so I can't ignore them.

lib/nxdk/configsector.c
```diff
@@ -147,3 +148,66 @@ bool nxLoadNetworkConfig(nxdk_network_config_sector_t *config)
 
     return true;
 }
+
+bool nxSaveNetworkConfigSector(HANDLE hPartition, UINT sectorIndex, const BYTE *buffer, UINT size)
+{
+    NXDK_CONFIG_SECTOR configSector;
+
+    _Static_assert(sizeof(configSector) == 512, "NXDK_CONFIG_SECTOR size mismatch");
```
JayFoxRox, line 156: To me, this feels like implicit documentation that config sectors (`NXDK_CONFIG_SECTOR`) are 512 bytes.

Therefore, I think I'd prefer if this was right below the struct definition of `NXDK_CONFIG_SECTOR` and not whereever `NXDK_CONFIG_SECTOR` is used, unless that's important for the task.

In this case, it might make sense to check against `NXDK_HDD_SECTOR_SIZE`? The function implies that this expects both to be the same value.

If you keep the assert here, it should be moved further down in my opinion (to whereever the assumption actually matters).

thrimbor: It's a safeguard against accidentally changing the size when the struct gets modified in the future when the meaning or type of more fields is figured out. I've de-duplicated and moved up the assert, and also added one for `nxdk_network_config_sector_t` to protect it, too.

Ryzee119:
Tested this on my 1.6 v5960 Dash 
Wrote these and all were successful and detected in MSDash:
```
config.manual.subnetMask = 0xABABABAB;
config.manual.defaultGateway = 0xCDCDCDCD;
config.manual.primaryDns = 0xEFEFEFEF;
config.manual.secondaryDns = 0xF0F0F0F0;
```

Issue #460, opened 2021-04-16
Improve network performance


JayFoxRox:
We need better network performance for various tasks:

- nv2a-trace (streaming/sending *a lot* of memory)
- xbe-loader (XISO streaming)
- Realtime games (enet for example, which is already ported in https://github.com/JayFoxRox/nxdk/pull/84)
- Audio communication over network when #438 hits and XBL Communicator support lands

This is a collection of unfinished work by @thrimbor in that area:

- https://github.com/thrimbor/nxdk/commits/net_improvements_clean
- https://github.com/thrimbor/nxdk/commits/net_improvements
- https://github.com/thrimbor/nxdk/commits/nvnetdrv
- https://github.com/thrimbor/nxdk/commits/pktdrv_perf

I'm not sure what the plans are.

I also checked, and we are already on latest stable lwip.

dracc:
Not to mention NevolutionX ftp/sntp/...

Less important when xbe-loader becomes performant but still nice-to-have IMO.

Ryzee119:
For visibility, Thrimbor and I have been working more on nvnetdrv. https://github.com/Ryzee119/nxdk/tree/nvnetdrv/lib/nvnetdrv
I have been using it for some time and is working very well for me with excellent performance >90mbit/s normally

Might just need a few more tweaks

dracc:
> For visibility, Thrimbor and I have been working more on nvnetdrv. https://github.com/Ryzee119/nxdk/tree/nvnetdrv/lib/nvnetdrv I have been using it for some time and is working very well for me with excellent performance >90mbit/s normally
> 
> Might just need a few more tweaks

Care to create a draft PR? Great work that's a bit hard to keep track of for outsiders. :)

Issue #129, opened 2019-06-09
Pktdrv calculates wrong packet length


JayFoxRox:
Formula for packet length should probably be this (based on disassembly):

```c
unsigned int length = flag & 0xFFFF;
if (flag & NV_RX_SUBSTRACT1) {
  assert(length > 0);
  length--;
}
```

*(Code in XQEMU and Cxbx-R downstream should be checked, too)*

JayFoxRox:
This was supposed to be closed by #611 

Closed.

PR #611, opened 2022-07-21
Replace pkdrv with nvnetdrv

Ryzee119:
This is a replacement network driver that intends to improve network performance of nxdk and replace pktdrv. Will close https://github.com/XboxDev/nxdk/issues/460 and https://github.com/XboxDev/nxdk/issues/129.
This work was started by @thrimbor and I kind of took it over.

~Relies on https://github.com/XboxDev/nxdk/pull/610.~

It has a few key improvements over pktdrv:
* Zero-copy TX and RX transmission
* Automatic scatter gather of TX packets that cross page boundaries so the user doesnt need to worry about it (Up to one page crossing, which should cover all ethernet frame lengths required for 100Mbit)
* Multithreaded buffer manager
* Supports different ring sizes for TX and RX so you can optimise your application for specifc directions to reduce RAM usage. Defaults to 64/64 which seems like a reasonable all around mode even under heavy loads.

As a comparison running lwip iperf server on the same Xbox and PC network config:

Current master (rx then tx test):
```
[  4] local 192.168.1.115 port 52244 connected with 192.168.1.113 port 5001
[ ID] Interval       Transfer     Bandwidth
[  4]  0.0-10.0 sec  39.4 MBytes  33.0 Mbits/sec
[  4] local 192.168.1.115 port 5001 connected with 192.168.1.113 port 51705
[  4]  0.0-10.3 sec  25.2 KBytes  20.1 Kbits/sec
```

nvnetdrv (rx then tx test)::
```
[  4] local 192.168.1.115 port 51764 connected with 192.168.1.113 port 5001
[ ID] Interval       Transfer     Bandwidth
[  4]  0.0-10.0 sec   112 MBytes  94.2 Mbits/sec
[  4] local 192.168.1.115 port 5001 connected with 192.168.1.113 port 51715
[  4]  0.0-10.0 sec   111 MBytes  92.9 Mbits/sec
```

It also performs quite well in full duplex mode (tx/rx in parallel):
```
[  4] local 192.168.1.115 port 57034 connected with 192.168.1.113 port 5001
[  5] local 192.168.1.115 port 5001 connected with 192.168.1.113 port 51704
[ ID] Interval       Transfer     Bandwidth
[  4]  0.0-10.0 sec   108 MBytes  90.3 Mbits/sec
[  5]  0.0-10.0 sec   106 MBytes  88.9 Mbits/sec
```

Current PRd as draft mainly due to it being a fairly large change and subject to feedback. I have tested this quite extensively and currently use it on my homebrew [dashboard](https://github.com/Ryzee119/LithiumX) in the form of an FTP server.

The `lwipopt.h` tweak is as per https://lwip.fandom.com/wiki/Tuning_TCP, which improved tx speed from ~60mbits to 90+mbits in my test with nvnetdrv. It made little impact to pktdrv perf.
```
Send-buffer (TCP_SND_BUF):
... For maximum throughput, set this to the same value as TCP_WND
```

Also fixes a bug in our network init function. We should be using `tcpip_input` instead of `ethernet_input` for OS mode (threaded) operation. Ref https://www.nongnu.org/lwip/2_0_x/group__netif.html#gade5498543e74067f28cc6bef0209e3be

### DIFF
--- a/lib/net/Makefile
+++ b/lib/net/Makefile
@@ -9,8 +9,8 @@ LWIPSRCS := $(COREFILES) \
             $(NETIFFILES)
 
 # Include driver sources
-DRIVERSRCS := $(NXDK_DIR)/lib/net/pktdrv/pktdrv.c \
-              $(NXDK_DIR)/lib/net/nforceif/src/driver.c \
+DRIVERSRCS := $(NXDK_DIR)/lib/net/nvnetdrv/nvnetdrv.c \
+              $(NXDK_DIR)/lib/net/nvnetdrv/nvnetdrv_lwip.c \
               $(NXDK_DIR)/lib/net/nforceif/src/sys_arch.c
 
 NET_SRCS := \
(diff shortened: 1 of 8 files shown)


Ryzee119:
Anything I can do to help this along?

Ernegien:
Things look solid with retail but testing with a debug-enabled BIOS the following mixed results were observed. Since debug interference already exists in nxdk without this PR, I'm inclined to support a merge, but this is an issue that should still be looked further into. Thoughts?

| Dash					| Pulls IP 	| XBDM Works |
|-----------------------|-----------|------------|
| LithiumX v0.6 		| No		| No		 |
| LithiumX v0.3 		| No		| No		 |
| NevolutionX v0.2.0 	| Yes		| No		 |
| UnleashX v0.39		| Yes		| Yes		 |

PR #614, opened 2022-09-07
hal: Relicense fileio code to MIT

thrimbor:
I got the original authors' permission to relicense this code under the MIT license. I can share proof of this on Discord on request, and have already shown it to @Ryzee119.

There was a commit by Carcharius, who I haven't been able to contact yet, but it only touched code that we already removed.


Ryzee119:
Can confirm I sighted a number of emails from previous authors who were happy to relicense.
Thanks for taking the time to do this

PR #625, opened 2023-02-04
net: Whitespace and macro cleanup in cc.h

thrimbor:
This removes a few bits of unnecessary preprocessor stuff and whitespaces, and removes the redefinition of `fflush` and `printf` that could cause some issues with header include order.


LoveMHz:
Thanks! That was a quick fix for the issue we were having. (re: discussion about this being an issue via the XboxDev Discord server).

PR #626, opened 2023-02-20
Add XBE->Exe Functionality

mariaWitch:
There are numerous analysis tools which are compatible with Windows PE files, but do not work with XBE files even though they are functionally compatible. Adding this functionality would allow for the conversion of Xbe files to a dummy Exe file allowing for the use of these tools. There is no need in needing these exe files to be executable or even functionally correct, as long as the correct header information is maintained between both files. 

Overall, the produced Exe file is not meant to be executed on actual systems, but only meant as a way to further additional analysis.

### DIFF
--- a/tools/cxbe/Exe.cpp
+++ b/tools/cxbe/Exe.cpp
@@ -6,9 +6,14 @@
 // SPDX-FileCopyrightText: 2021 Stefan Schmidt
 
 #include "Exe.h"
+#include "Xbe.h"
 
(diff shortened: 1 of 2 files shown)


mariaWitch:
I welcome any additions that Maintainers may have, as most of the information that I am using is coming from the Exe->Xbe conversion process that already exists in Cxbe, and EmuExe.cpp from Cxbx.

PR #628, opened 2023-03-02
pbkit: Enable flip stall; expose ramin, print_char, depth buffer internals

antangelo:
This PR upstreams the pbkit changes required by the [pgraph test suite](https://github.com/abaire/nxdk_pgraph_tests).

They are all combined into one PR, but can be split it if preferred.
CC: @abaire

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -1607,9 +1607,32 @@ static void pb_3D_init(void)
 #endif
 }
 
+DWORD pb_reserve_instance(DWORD size) {
+    DWORD ret = pb_FreeInst;
+    pb_FreeInst += (size>>4);
+    return ret;
+}
 
+void pb_create_gr_instance(int ChannelID,
+                           int Class,
+                           DWORD instance,
+                           DWORD flags,
+                           DWORD flags3D,
+                           struct s_CtxDma *pGrObject)
+{
+    DWORD offset = instance << 4;
+    VIDEOREG(NV_PRAMIN + offset + 0x00) = flags;
+    VIDEOREG(NV_PRAMIN + offset + 0x04) = flags3D;
+    VIDEOREG(NV_PRAMIN + offset + 0x08) = 0;
+    VIDEOREG(NV_PRAMIN + offset + 0x0C) = 0;
 
+    memset(pGrObject,0,sizeof(struct s_CtxDma));
 
+    pGrObject->ChannelID = ChannelID;
+    pGrObject->Class = Class;
+    pGrObject->isGr = 1;
+    pGrObject->Inst = instance;
+}
 
 void pb_create_gr_ctx(   int ChannelID,
                 int Class,
@@ -1643,14 +1666,13 @@ void pb_create_gr_ctx(   int ChannelID,
         }
     }
 
-    Inst=pb_FreeInst; pb_FreeInst+=(size>>4);
+    Inst = pb_reserve_instance(size);
 
     if (flags3D)
     {
         pb_3DGrCtxInst[pb_FifoChannelID]=Inst;
         pb_3D_init();
     }
-
 
     flags=Class&0x000000FF;
     flags3D=0x00000000;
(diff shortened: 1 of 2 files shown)


lib/pbkit/pbkit.c
```diff
@@ -2468,7 +2468,7 @@ int pb_finished(void)
     p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_WAIT_MAKESPACE,0); //wait/makespace (obtains null status)
     p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_PARAMETER_A,pb_back_index); //set param=back buffer index to show up
     p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_FIRE_INTERRUPT,PB_FINISHED); //subprogID PB_FINISHED: gets frame ready to show up soon
-//  p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_STALL_PIPELINE,0); //stall gpu pipeline (not sure it's needed in triple buffering technic)
+    p=pb_push1(p,NV20_TCL_PRIMITIVE_3D_STALL_PIPELINE,0); //stall gpu pipeline (not sure it's needed in triple buffering technic)
```
JayFoxRox, line 2476: Why is this needed?

lib/pbkit/pbkit.c
```diff
@@ -2630,6 +2631,11 @@ void pb_set_color_format(unsigned int fmt, bool swizzled) {
     assert(swizzled == false);
 }
 
+void pb_set_fb_size_multiplier(unsigned int multiplier) {
+    assert(multiplier > 0);
+    pb_FBSizeMultiplier = multiplier;
```
JayFoxRox, line 2639: When would you use this?

lib/pbkit/pbkit.h
```diff
@@ -159,8 +159,6 @@ void pb_create_gr_instance(int ChannelID,
                         DWORD flags3D,
                         struct s_CtxDma *pGrObject);
 
-// Exposed so pb_print can be overridden using a version of vsprintf that
-// supports floats.
```
JayFoxRox: Why was this comment removed?

lib/pbkit/pbkit.h
```diff
@@ -144,6 +144,10 @@ void pb_create_gr_ctx(   int ChannelID,
                       struct s_CtxDma *pGrObject  );
 void pb_bind_channel(struct s_CtxDma *pCtxDmaObject);
 
+uint8_t *pb_depth_stencil_buffer();
+DWORD pb_depth_stencil_pitch();
+DWORD pb_depth_stencil_size();
```
JayFoxRox, line 152: I always thought pbkit had this interface backwards. Currently pbkit is a mix of a low-level and high-level driver.

pbkit can submit pushbuffers, but it also constructs them, some using rather high level operations.
Instead, I believe that pbkit should allow submission of pushbuffers, but it should primarily be the users responsibility to construct them.
pbkit should only control the hardware so it can process commands from the pushbuffers.

More importantly, pbkit also manages buffers internally (sometimes like a singleton), but doesn't allow the user to manually create or control these buffers.
Instead, I believe that pbkit should provide the user with buffer creation / managing functions, but the user should always explicitly create these buffers and keep track of them.

This means I'd prefer to see something like `pb_create_depth_stencil_buffer(size, pitch)` and then the user should be responsible for creating these buffers.
So, in an ideal world, the application would never even have to query for the buffer address / pitch / size, because it created the buffer itself.

This is also required (or at least helpful) for more advanced applications which want to switch render-targets.

JayFoxRox:
Commit titles don't follow our usual style (`prefix: <Verb> [...]` without any punctuation at the end)

lib/pbkit/pbkit.c
abaire: If I remember correctly it was just to mimic end of frame behavior observed in DirectX based titles in order to make tooling (like the pgraph tracer) simpler.

lib/pbkit/pbkit.c
abaire: For fullscreen antialiasing: https://github.com/abaire/nxdk_pgraph_tests/blob/e4e1dbce47d2ee7741f4ef1b219952ad6a410f78/src/main.cpp#L154

lib/pbkit/pbkit.h
abaire: Totally agree, Thrimbor also agrees and has plans for a rewrite. My fork uses this as a stopgap; not sure it's worth the effort to take a half-step in the direction of providing user control if there's a major refactor pending. I also think some of this may have been added in commits I upstreamed more recently.

abaire (commented):
This would at least need a rebase since some of the changes were upstreamed and I ended up doing a minor refactor to break pbkit.c into multiple files for ease of maintenance. Personally I don't feel strongly that it's worth the effort given that pbkit is slated for replacement.

PR #631, opened 2023-03-13
CMake: Toolchain fixes

RadWolfie:
After some troubleshooting process to figure out what went wrong with docker image for cmake usage. I found out there are several both manually incorrect coded and missing some support.

- pkgconfig for libjpeg missed replacement of  round to parenthesis brackets
- Whole *.pc files were missing -l, lowercase L, before each library files in `Libs:` variable. I had confirmed the correction now appear in `<PREFIX>_LIBRARIES` and `<PREFIX>_LINK_LIBRARIES` from output message test in cmake.
- FindThreads.cmake module is a requirement due to stock module will try to find pthread instead and fail. Since `CMAKE_SYSTEM_NAME` variable is set to Generic instead of Windows which fall back to unix-like's search method.
- FindSDL2.cmake should be a requirement in anyway due to definitions does not get include and only be found in `<PREFIX>_CFLAGS_OTHER`. Although, it does show up in `<PREFIX>_CFLAGS` except got exclude at compile/link time for unknown reason.
- For unknown reason, find_package did not look in nxdk's directory therefore require manually add path into `CMAKE_MODULE_PATH`.
  **NOTE:** `CMAKE_MODULE_PATH` is document intended for project level than toolchain.
- Add xbe conversion and check if author want xiso generated after compilation completion.

The changes made were tested with experimental package published on GitHub to test with NevolutionX application using cmake. It is able to compile with nxdk and with linux libraries (require install sdl2 packages).

### DIFF
--- a/bin/nxdk-cc
+++ b/bin/nxdk-cc
@@ -17,6 +17,9 @@ clang \
     -I${NXDK_DIR}/lib/pdclib/platform/xbox/include \
     -I${NXDK_DIR}/lib/winapi \
     -I${NXDK_DIR}/lib/xboxrt/vcruntime \
+    -I${NXDK_DIR}/lib/net/lwip/src/include \
+    -I${NXDK_DIR}/lib/net/nforceif/include \
+    -I${NXDK_DIR}/lib/net/nvnetdrv \
     -DNXDK \
     -D__STDC__=1 \
     -U__STDC_NO_THREADS__ \
--- a/share/cmake/Modules/FindSDL2.cmake
+++ b/share/cmake/Modules/FindSDL2.cmake
@@ -1,0 +1,18 @@
+# FPHSA_NAME_MISMATCHED is require to suppress warning message 
+set(FPHSA_NAME_MISMATCHED TRUE)
+include(FindPkgConfig)
+pkg_check_modules(sdl2 REQUIRED sdl2)
+unset(FPHSA_NAME_MISMATCHED)
+
+add_library(SDL2::SDL2 INTERFACE IMPORTED)
+set(SDL2_INCLUDE_DIRS ${sdl2_INCLUDE_DIRS})
+set(SDL2_LIBRARIES ${sdl2_LIBRARIES})
+set(SDL2_LINK_LIBRARIES ${sdl2_LINK_LIBRARIES})
+set_target_properties(SDL2::SDL2 PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES "${SDL2_INCLUDE_DIRS}"
+  INTERFACE_LINK_LIBRARIES "${SDL2_LINK_LIBRARIES}"
+  # NOTE: pkg_check_modules' Cflags definition for "XBOX" did not get included and was passed to CFLAGS_OTHER for some reason... 
+  INTERFACE_COMPILE_OPTIONS "${sdl2_CFLAGS}"
+)
+set(SDL2_FOUND 1)
+add_library(SDL2 ALIAS SDL2::SDL2)
--- a/share/toolchain-nxdk.cmake
+++ b/share/toolchain-nxdk.cmake
@@ -19,7 +19,7 @@ set(WIN32 1)
 set(NXDK 1)
 
 set(CMAKE_C_COMPILER "${NXDK_DIR}/bin/${TOOLCHAIN_PREFIX}-cc")
-set(CMAKE_C_STANDARD_LIBRARIES "${NXDK_DIR}/lib/libwinapi.lib ${NXDK_DIR}/lib/xboxkrnl/libxboxkrnl.lib ${NXDK_DIR}/lib/libxboxrt.lib  ${NXDK_DIR}/lib/libpdclib.lib ${NXDK_DIR}/lib/libnxdk_hal.lib ${NXDK_DIR}/lib/libnxdk.lib ${NXDK_DIR}/lib/nxdk_usb.lib") #"${CMAKE_CXX_STANDARD_LIBRARIES_INIT}"
+set(CMAKE_C_STANDARD_LIBRARIES "${NXDK_DIR}/lib/libwinapi.lib ${NXDK_DIR}/lib/xboxkrnl/libxboxkrnl.lib ${NXDK_DIR}/lib/libxboxrt.lib  ${NXDK_DIR}/lib/libpdclib.lib ${NXDK_DIR}/lib/libnxdk_hal.lib ${NXDK_DIR}/lib/libnxdk.lib ${NXDK_DIR}/lib/nxdk_usb.lib ${NXDK_DIR}/lib/libnxdk_net.lib") #"${CMAKE_CXX_STANDARD_LIBRARIES_INIT}"
 set(CMAKE_C_LINK_EXECUTABLE "${NXDK_DIR}/bin/${TOOLCHAIN_PREFIX}-link <FLAGS> <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -out:<TARGET> <LINK_LIBRARIES>")
 
 set(CMAKE_CXX_COMPILER "${NXDK_DIR}/bin/${TOOLCHAIN_PREFIX}-cxx")
--- a/share/toolchain-nxdk.cmake
+++ b/share/toolchain-nxdk.cmake
@@ -1,3 +1,10 @@
+# NOTE: Require to prevent obtaining pre-existing information
+#       and calling macro repeatly unintentionally.
+if (__TOOLCHAIN_NXDK)
+  return()
+endif()
+set(__TOOLCHAIN_NXDK 1)
+
 if(DEFINED ENV{NXDK_DIR})
     set(NXDK_DIR $ENV{NXDK_DIR})
 else()
@@ -49,3 +56,30 @@ set(CMAKE_C_COMPILE_OPTIONS_IPO -flto)
 #       find out what's wrong with CMAKE_FIND_ROOT_PATH as it is not working for some reason?
 set(CMAKE_MODULE_PATH "${NXDK_DIR}/share/cmake/Modules")
 set(PKG_CONFIG_EXECUTABLE "${NXDK_DIR}/bin/nxdk-pkg-config" CACHE STRING "Path to pkg-config")
+
+# Additional toolchain steps required for create xbe and xiso files.
+macro(add_executable)
+  # Perform normal call since there's no modification needed except for additional steps require
+  _add_executable(${ARGV})
+
+  # If author request generate iso file, then prepare the command here.
+  if(GEN_XISO)
+    set(XISO_OUTPUT "${NXDK_DIR}/tools/extract-xiso/build/extract-xiso" -c "$<TARGET_FILE_DIR:${ARGV0}>/bin" "$<TARGET_FILE_DIR:${ARGV0}>/${GEN_XISO}")
+  endif()
+
+  # If custom title is not given, use given executable name.
+  if(NOT XBE_TITLE)
+    set(XBE_TITLE ${ARGV0})
+  endif()
+
+  # Attach additional toolchain steps to executable project.
+  add_custom_command(TARGET ${ARGV0} POST_BUILD
+    COMMAND mkdir -p $<TARGET_FILE_DIR:${ARGV0}>/bin
+    # Generate xbe binary conversion
+    COMMAND "${NXDK_DIR}/tools/cxbe/cxbe" "-OUT:$<TARGET_FILE_DIR:${ARGV0}>/bin/default.xbe" "-TITLE:${XBE_TITLE}" "$<TARGET_FILE_DIR:${ARGV0}>/${ARGV0}.exe"
+    # Optionally generate xiso file if requested
+    COMMAND "$<$<BOOL:${XISO_OUTPUT}>:$<JOIN:${XISO_OUTPUT},;>>"
+    COMMAND_EXPAND_LISTS
+    VERBATIM
+  )
+endmacro(add_executable)
--- a/share/toolchain-nxdk.cmake
+++ b/share/toolchain-nxdk.cmake
@@ -45,4 +45,7 @@ set(_CMAKE_C_IPO_SUPPORTED_BY_CMAKE YES)
 set(_CMAKE_C_IPO_MAY_BE_SUPPORTED_BY_COMPILER YES)
 set(CMAKE_C_COMPILE_OPTIONS_IPO -flto)
 
+# TODO: CMAKE_MODULE_PATH is intended for project level than toolchain,
+#       find out what's wrong with CMAKE_FIND_ROOT_PATH as it is not working for some reason?
+set(CMAKE_MODULE_PATH "${NXDK_DIR}/share/cmake/Modules")
 set(PKG_CONFIG_EXECUTABLE "${NXDK_DIR}/bin/nxdk-pkg-config" CACHE STRING "Path to pkg-config")
--- a/bin/nxdk-cxx
+++ b/bin/nxdk-cxx
@@ -18,6 +18,9 @@ clang \
     -I${NXDK_DIR}/lib/pdclib/platform/xbox/include \
     -I${NXDK_DIR}/lib/winapi \
     -I${NXDK_DIR}/lib/xboxrt/vcruntime \
+    -I${NXDK_DIR}/lib/net/lwip/src/include \
+    -I${NXDK_DIR}/lib/net/nforceif/include \
+    -I${NXDK_DIR}/lib/net/nvnetdrv \
     -DNXDK \
     -D__STDC__=1 \
     -U__STDC_NO_THREADS__ \
--- a/lib/pkgconfig/SDL2_image.pc
+++ b/lib/pkgconfig/SDL2_image.pc
@@ -2,5 +2,5 @@ Name: SDL2_image
 Description: image loading library for Simple DirectMedia Layer
 Version: 2.0.5
 Requires: sdl2 >= 2.0.9 libjpeg libpng
-Libs: ${NXDK_DIR}/lib/libSDL2_image.lib
+Libs: -l${NXDK_DIR}/lib/libSDL2_image.lib
 Cflags: -I${NXDK_DIR}/lib/sdl/SDL2_image
--- a/lib/pkgconfig/SDL2_ttf.pc
+++ b/lib/pkgconfig/SDL2_ttf.pc
@@ -2,5 +2,5 @@ Name: SDL2_ttf
 Description: ttf library for Simple DirectMedia Layer with FreeType 2 support
 Version: 2.0.14
 Requires: sdl2 >= 2.0.9
-Libs: ${NXDK_DIR}/lib/libSDL_ttf.lib ${NXDK_DIR}/lib/libfreetype.lib
+Libs: -l${NXDK_DIR}/lib/libSDL_ttf.lib -l${NXDK_DIR}/lib/libfreetype.lib
 Cflags: -I${NXDK_DIR}/lib/sdl -I${NXDK_DIR}/lib/sdl/SDL_ttf
--- a/lib/pkgconfig/libjpeg.pc
+++ b/lib/pkgconfig/libjpeg.pc
@@ -2,4 +2,4 @@ Name: libjpeg
 Description: A SIMD-accelerated JPEG codec that provides the libjpeg API
 Version: 2.0.4
 Libs: ${NXDK_DIR}/lib/libjpeg.lib
-Cflags: -I${NXDK_DIR}/lib/libjpeg/libjpeg-turbo -I$(NXDK_DIR)/lib/libjpeg
+Cflags: -I${NXDK_DIR}/lib/libjpeg/libjpeg-turbo -I${NXDK_DIR}/lib/libjpeg
--- a/lib/pkgconfig/libjpeg.pc
+++ b/lib/pkgconfig/libjpeg.pc
@@ -1,5 +1,5 @@
 Name: libjpeg
 Description: A SIMD-accelerated JPEG codec that provides the libjpeg API
 Version: 2.0.4
-Libs: ${NXDK_DIR}/lib/libjpeg.lib
+Libs: -l${NXDK_DIR}/lib/libjpeg.lib
 Cflags: -I${NXDK_DIR}/lib/libjpeg/libjpeg-turbo -I${NXDK_DIR}/lib/libjpeg
--- a/lib/pkgconfig/libpng.pc
+++ b/lib/pkgconfig/libpng.pc
@@ -2,5 +2,5 @@ Name: libpng
 Description: Loads and saves PNG files
 Version: 1.6.37
 Requires: zlib
-Libs: ${NXDK_DIR}/lib/libpng.lib
+Libs: -l${NXDK_DIR}/lib/libpng.lib
 Cflags: -I${NXDK_DIR}/lib/libpng -I${NXDK_DIR}/lib/libpng/libpng
--- a/lib/pkgconfig/sdl2.pc
+++ b/lib/pkgconfig/sdl2.pc
@@ -3,5 +3,5 @@ Description: Simple DirectMedia Layer is a cross-platform multimedia library des
 Version: 2.0.9
 Requires:
 Conflicts:
-Libs: ${NXDK_DIR}/lib/libSDL2.lib
+Libs: -l${NXDK_DIR}/lib/libSDL2.lib
 Cflags: -I${NXDK_DIR}/lib/sdl/SDL2/include -DXBOX
--- a/lib/pkgconfig/zlib.pc
+++ b/lib/pkgconfig/zlib.pc
@@ -3,5 +3,5 @@ Description: zlib compression library
 Version: 1.2.11
 
 Requires:
-Libs: ${NXDK_DIR}/lib/libzlib.lib
+Libs: -l${NXDK_DIR}/lib/libzlib.lib
 Cflags: -I${NXDK_DIR}/lib/zlib/zlib -DZ_SOLO
--- a/share/cmake/Modules/FindThreads.cmake
+++ b/share/cmake/Modules/FindThreads.cmake
@@ -1,0 +1,6 @@
+# NOTE: custom FindThreads module is a requirement due to CMake's FindThreads module is looking for pThreads.
+# nxdk's toolchain is using system named Generic instead of "Windows" which could had work but inaccurate name.
+# xboxkrnl and pdclib have C11 thread support out of the box
+set(CMAKE_HAVE_THREADS_LIBRARY 1)
+set(Threads_FOUND TRUE)
+add_library(Threads::Threads INTERFACE IMPORTED)


share/toolchain-nxdk.cmake
```diff
@@ -45,4 +52,34 @@ set(_CMAKE_C_IPO_SUPPORTED_BY_CMAKE YES)
...
+
+# Additional toolchain steps required for create xbe and xiso files.
+macro(add_executable)
+  # Perform normal call since there's no modification needed except for additional steps require
+  _add_executable(${ARGV})
+
+  # If author request generate iso file, then prepare the command here.
+  if(GEN_XISO)
+    set(XISO_OUTPUT "${NXDK_DIR}/tools/extract-xiso/build/extract-xiso" -c "$<TARGET_FILE_DIR:${ARGV0}>/bin" "$<TARGET_FILE_DIR:${ARGV0}>/${GEN_XISO}")
+  endif()
```
JayFoxRox, line 68: I suggest to leave XISO out of this. It also doesn't feel CMake-ish.

It also doesn't handle a common use-case where a user will want to selectively add more files to the XISO; so this will need a better design.

Also, if we support `GEN_XISO` now, then we can't remove it without breaking projects.
As nxdk matures, we should avoid breakage like that.

---

[In my own CMake work I just created another CMake function](https://github.com/JayFoxRox/nxdk/pull/74/files#diff-f4116787576a8fd4b6cba2c395d8f2fa515c539bec4c101de9cf47bc77d3dce4R33-R38) but I believe I still did XISO creation manually in a shell script for flexibility.

share/toolchain-nxdk.cmake
```diff
@@ -19,7 +26,7 @@ set(WIN32 1)
 set(NXDK 1)
 
 set(CMAKE_C_COMPILER "${NXDK_DIR}/bin/${TOOLCHAIN_PREFIX}-cc")
-set(CMAKE_C_STANDARD_LIBRARIES "${NXDK_DIR}/lib/libwinapi.lib ${NXDK_DIR}/lib/xboxkrnl/libxboxkrnl.lib ${NXDK_DIR}/lib/libxboxrt.lib  ${NXDK_DIR}/lib/libpdclib.lib ${NXDK_DIR}/lib/libnxdk_hal.lib ${NXDK_DIR}/lib/libnxdk.lib ${NXDK_DIR}/lib/nxdk_usb.lib") #"${CMAKE_CXX_STANDARD_LIBRARIES_INIT}"
+set(CMAKE_C_STANDARD_LIBRARIES "${NXDK_DIR}/lib/libwinapi.lib ${NXDK_DIR}/lib/xboxkrnl/libxboxkrnl.lib ${NXDK_DIR}/lib/libxboxrt.lib  ${NXDK_DIR}/lib/libpdclib.lib ${NXDK_DIR}/lib/libnxdk_hal.lib ${NXDK_DIR}/lib/libnxdk.lib ${NXDK_DIR}/lib/nxdk_usb.lib ${NXDK_DIR}/lib/libnxdk_net.lib") #"${CMAKE_CXX_STANDARD_LIBRARIES_INIT}"
```
JayFoxRox, line 29: I'm not sure why networking would be part of the C standard.
Applications which want to use this, should add it explicitly.

Also see https://github.com/JayFoxRox/nxdk/pull/84/files#r611210118

Once nxdk has winsock, that could also take care of it.

share/toolchain-nxdk.cmake
RadWolfie: I like the concept of making it as CMake function which I now notice from you. Initially I'm following how makefile handling things which currently use `GEN_XISO`. If preferred the change to CMake function, I'll be happily change it. Yet may need to preserve backward compatibility for the change from make to CMake?

share/toolchain-nxdk.cmake
```diff
@@ -45,4 +52,34 @@ set(_CMAKE_C_IPO_SUPPORTED_BY_CMAKE YES)
...
+  # Perform normal call since there's no modification needed except for additional steps require
+  _add_executable(${ARGV})
+
+  # If author request generate iso file, then prepare the command here.
+  if(GEN_XISO)
+    set(XISO_OUTPUT "${NXDK_DIR}/tools/extract-xiso/build/extract-xiso" -c "$<TARGET_FILE_DIR:${ARGV0}>/bin" "$<TARGET_FILE_DIR:${ARGV0}>/${GEN_XISO}")
+  endif()
+
+  # If custom title is not given, use given executable name.
+  if(NOT XBE_TITLE)
```
JayFoxRox, line 71: I believe this variable name needs a prefix to not clash with user variables in CMake?

There's probably some convention for this in CMake.

share/cmake/Modules/FindSDL2.cmake
```diff
@@ -0,0 +1,18 @@
...
+unset(FPHSA_NAME_MISMATCHED)
+
+add_library(SDL2::SDL2 INTERFACE IMPORTED)
+set(SDL2_INCLUDE_DIRS ${sdl2_INCLUDE_DIRS})
+set(SDL2_LIBRARIES ${sdl2_LIBRARIES})
+set(SDL2_LINK_LIBRARIES ${sdl2_LINK_LIBRARIES})
+set_target_properties(SDL2::SDL2 PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES "${SDL2_INCLUDE_DIRS}"
+  INTERFACE_LINK_LIBRARIES "${SDL2_LINK_LIBRARIES}"
+  # NOTE: pkg_check_modules' Cflags definition for "XBOX" did not get included and was passed to CFLAGS_OTHER for some reason...
```
JayFoxRox, line 14: nit: Most comments in this PR have typos (`pkg_check_modules'`) or poor wording. Someone should spell check these.

share/toolchain-nxdk.cmake
RadWolfie: NXDK_NET option is dropped, so it is always included. See #575 pull request changes.

share/toolchain-nxdk.cmake
JayFoxRox: > Yet may need to preserve backward compatibility for the change from make to CMake?

I don't think so.

Users who move from the makefile to CMake will have to change their code anyway, so they can add the function call themselves.

I believe the original nxdk makefile (for use outside of nxdk internally) will also be retired eventually, and users who still want to use a makefile can use their own OR find some build-system they like (which doesn't have to be CMake).
Also [see my work on autoconf etc.](https://github.com/JayFoxRox/nxdk/pull/83/files) - while CMake support is part of nxdk, it's trivial to use other build-systems after the existing refactors to the build-system (which only added CMake, for better or worse)

share/cmake/Modules/FindSDL2.cmake
RadWolfie: It is not a typo, you can find it at https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules . Since the function's name is already plural, then there shouldn't be an s after apostrophe?

share/toolchain-nxdk.cmake
RadWolfie: It is only triggered when user call add_executable function. Though, I will check with forked NevolutionX repo with fixed CMake support to be sure it doesn't return back to user's CMakeList. It may also apply to `GEN_XISO` variable too.

share/toolchain-nxdk.cmake
JayFoxRox: There isn't a good reason why it should always be linked (for CMake and the legacy nxdk makefile alike).

I can only speculate, but I assume the intention in #575 was to always build the lib, so the nxdk makefile is used for actually building nxdk (= libraries usable by pkg-config / CMake / ..), instead of building a single application (= decision which libraries to include); the PR description says:

> To make building the libraries easier, the network code is now always built.

Maybe @thrimbor can clarify?

I believe, in your case, the CMake user should now choose which libraries to include (but CMake will take care of the standard library, hence you need to specify those and dependencies for the C/C++ standard).
The fewer symbols our CMake brings, the better (smaller binaries, faster linking, less symbol name collisions, easier to customize your drivers/backends).

share/toolchain-nxdk.cmake
RadWolfie: Is there opened ticket discussed about it and/or have list of tasks for moving away from original nxdk makefile to have it done through toolchain script? I am currently unable to find the information in the repository. If there is or isn't, it will be helpful to follow through guideline to make a working toolchain as possible and have tasks done.

share/toolchain-nxdk.cmake
RadWolfie: One small thing I would like to add into JayFoxRox's comment. If download source code repo without build nxdk via make. Then use CMake, the toolchain, to build user's application will not setup the necessary build files due to nxdk did not get built. I took a punch in my own eye not realizing this and didn't think about build nxdk via make. Mainly because I thought it was already handled through nxdk's makefile from the toolchain script. I was completely wrong.

share/toolchain-nxdk.cmake
RadWolfie: I can confirm XBE_TITLE is stored back into user's CMakeList, will need to make a fix for this.

concept fix:
```cmake
  if(NOT XBE_TITLE)
    set(<prefix>_XBE_TITLE ${ARGV0})
  else()
    set(<prefix>_XBE_TITLE ${XBE_TITLE})
  endif()
```
As for prefix, not sure if we should use NXDK or CMAKE prefix.

RadWolfie:
@thrimbor, can we get a [public guideline](https://github.com/XboxDev/nxdk/pull/631#discussion_r1224811437) what toolchain should do? There's no information about it. Such as `GEN_XISO` should be replace as a function to permit title developers to create their own xiso image. Using reference of makefile, both main and samples, files from nxdk here isn't helpful. Plus there are some unresolved discussion that are welcome further discussions. 🙏

share/toolchain-nxdk.cmake
thrimbor: The changes in #575 regarding building were mostly to simplify building nxdk itself - adjusting compiler flags and include directories based on make vars is a hot mess when building the libraries.

Regarding the change above, I think I'm ok with it *for now*. Ultimately I'd like to see a winsock library (next lwIP release should hopefully bring us closer to being able to implement this), and then we can have a `ws2_32.lib` that I'd expect the user to have to explicitly specify, which can then pull in the other networking libraries via `#pragma comment(lib,"libnxdk_net.lib")` (this is done a lot on Windows). Atm our messy buildsystem relies on dependencies being explicit, so we can't really utilize this as long as we have a makefile that's expected to handle both building nxdk itself *and* applications.

In the long term I'd like to see the low-level networking and USB libs removed from the standard libs and have them explicitly linked against if the app touches low-level stuff directly, or automatically pulled in by higher level libraries that are explicitly specified by the user such as SDL or winsock. But as long as we can't have that, imho linking them in might be the lesser of two evils.

thrimbor:
The closest thing to a guideline I have is "don't rely on anything the makefile does for you". It didn't really start out as the base for a toolchain, the goal probably was to have something that makes it as simple as possible to build the example it came with. As such it brings a few convenience functions, but with limitations and and reliance on specifics of flawed tools.

Xiso creation is one of those - it gives very little control over the end result (on-disc layout, non-1:1 directory mapping etc.), and *technically* isn't required to get something running on the Xbox, so imho we can leave that out the CMake toolchain file. It's still frequently used for quick emulator testing, so my suggestion would be to have a wiki entry "Creating an xiso from CMake" or something that supplies code that can just be dropped in.

Feature parity between our current makefile and CMake to ease the switch is not a big concern to me - getting rid of the way we currently build things will break applications relying on it and require major adjustments anyway.

thrimbor:
I'm not close to having reviewed the whole PR yet (my free time is rather fragmented), but regarding https://github.com/XboxDev/nxdk/pull/631/commits/702b32e96476083e20fd6bb0eb89fbc09b944097:

Is that change required by CMake? As it is now that commit breaks direct usage of `nxdk-pkg-config` as `lld-link` does not understand `-l` parameters (it takes `link.exe` syntax as opposed to `ld`)

RadWolfie:
When I was debugging the issues for CMake problems from codespace I'll quote what I had found from OP post:
> * Whole *.pc files were missing -l, lowercase L, before each library files in `Libs:` variable. I had confirmed the correction now appear in `<PREFIX>_LIBRARIES` and `<PREFIX>_LINK_LIBRARIES` from output message test in cmake.

`Libs:` accept `-L<path to lib dir>` and `-l<path to lib file>` according to [pkg-config documentation](https://people.freedesktop.org/~dbn/pkg-config-guide.html#using) I had looked at. If that's the wrong source to look at the documentation, please direct me to the right source. Otherwise, without -l argument included as I quoted will not work correctly with CMake.

thrimbor:
Like that guide says it's for Unix-like systems. `-L` and `-l` are flags for `ld` and not compatible with `link.exe` so we cannot use them.

> Otherwise, without -l argument included as I quoted will not work correctly with CMake.

What will not work? What was the issue and how can I reproduce it?

RadWolfie:
> What will not work? What was the issue and how can I reproduce it?

You can use https://github.com/RadWolfie/NevolutionX/tree/dev-container branch, select green code button dropdown menu > three dots (...) > + New with options... > it will open a page with options. The only thing need to change is Dev container configuration to "Xbox Development". Once done, you can press "Create codespace". From there, you can test compile build out of the box which is currently using my own nxdk package. The modified I made to CMakeLists.txt file to show the output are:
```
# Place this after "pkg_search_module(SDL2TTF REQUIRED SDL2_ttf)" line.
message("SDL2TTF libraries: ${SDL2TTF_LIBRARIES}")
message("SDL2TTF link libraries: ${SDL2TTF_LINK_LIBRARIES}")
message("SDL2TTF cflags: ${SDL2TTF_CFLAGS}")
```
which output:
```
[cmake] SDL2TTF libraries: /usr/src/nxdk/lib/libSDL_ttf.lib;/usr/src/nxdk/lib/libfreetype.lib;/usr/src/nxdk/lib/libSDL2.lib
[cmake] SDL2TTF link libraries: /usr/src/nxdk/lib/libSDL_ttf.lib;/usr/src/nxdk/lib/libfreetype.lib;/usr/src/nxdk/lib/libSDL2.lib
[cmake] SDL2TTF cflags: -I/usr/src/nxdk/lib/sdl;-I/usr/src/nxdk/lib/sdl/SDL_ttf;-I/usr/src/nxdk/lib/sdl/SDL2/include;-DXBOX
````

To reproduce the issue from upstream. You will need go into `.devcontainer/xbox/Dockerfile` file and change very first line to `FROM ghcr.io/xboxdev/nxdk:latest`. After the change click "Codespaces" from bottom left corner, select "Full rebuild container". Then you will want to comment out `find_package(Threads REQUIRED)` and `find_package(SDL2 REQUIRED)` lines. As they will block continue run generate build setup.
After that and do the build, the output will be:
```
[cmake] SDL2TTF libraries: 
[cmake] SDL2TTF link libraries: 
[cmake] SDL2TTF cflags: -I/usr/src/nxdk/lib/sdl;-I/usr/src/nxdk/lib/sdl/SDL_ttf;-I/usr/src/nxdk/lib/sdl/SDL2/include;-DXBOX
```

As for cannot have -L / -l is a bit confusing... I ran a CI test for pkg-config command on all three available platforms from GitHub action: https://github.com/RadWolfie/cross-platform-pkg-config-test/actions/runs/5634597949
On Windows, it even says `-lcrypto`, `-lpng16 -lz`, and `-ljpeg` via doing the commands below:
```
pkg-config libcrypto --libs
pkg-config libpng --libs
pkg-config libjpeg --libs
```
🤔
After your edited comment, I append another job test except all has failed for `--msvc-syntax`. I have not used pkg-config on Windows platform.

***on side note:*** I didn't know it is possible to compile nxdk with msvc?

**Comment by thrimbor** [comment]
> ```
> [cmake] SDL2TTF libraries: 
> [cmake] SDL2TTF link libraries: 
> [cmake] SDL2TTF cflags: -I/usr/src/nxdk/lib/sdl;-I/usr/src/nxdk/lib/sdl/SDL_ttf;-I/usr/src/nxdk/lib/sdl/SDL2/include;-DXBOX
> ```

I see. Annoying that CMake seems to mess with these files directly, but I guess we'll need to keep CMake happy.

> As for cannot have -L / -l is a bit confusing... I ran a CI test for pkg-config command on all three available platforms from GitHub action: https://github.com/RadWolfie/cross-platform-pkg-config-test/actions/runs/5634597949 On Windows, it even says `-lcrypto`, `-lpng16 -lz`, and `-ljpeg` via doing the commands below:

That's expected, pkg-config defaults to assuming a Unix-like environment (mysys/Cygwin on Windows).

> thinking After your edited comment, I append another job test except all has failed for `--msvc-syntax`. I have not used pkg-config on Windows platform.

That's annoying - there are three different implementations of pkg-config:
* pkg-config: The original implementation that supports `--msvc-syntax`
* pkgconf: Fully compatible reimplementation that seems to get preferred by Linux distros nowadays
* PkgConfig: A reimplementation in Perl that is **not** fully compatible and does not support `--msvc-syntax`

For some reason you seem to get `PkgConfig` in your tests, which sucks. On my system (Arch Linux), I get `pkgconf` (version 1.8.1), which works just fine:
image

If you want to experiment you could try calling `pkgconf` instead of  `pkg-config`, maybe it's installed but not the default. In the worst case we'll have to ship `pkgconf` with nxdk.

> _**on side note:**_ I didn't know it is possible to compile nxdk with msvc?

It's not, but the tools we use (`lld-link`) are designed as drop-in replacements for MSVC tools and follow the same cli syntax. We also use the same ABI so in theory you should be able to mix and match LLVM and MSVC when building something that uses nxdk.

**Comment by RadWolfie** [comment]
From codespace using the docker image, it is using pkgconf (version 1.9.4) when I call pkg-config -h and pkg-config --version. To be sure it's not a fluke, I can confirm it is the same via call pkgconf. I can also confirm `--msvc-syntax` does appear from -h argument. I took a look in CMakeCache.txt file which shows `FIND_PACKAGE_MESSAGE_DETAILS_PkgConfig:INTERNAL=[/usr/src/nxdk/bin/nxdk-pkg-config][v1.9.4()]` meaning it does use pkgconf.

<details>
<summary>I tried downgrade pkgconf to 1.8.1 yet still it gives me blank library inputs.</summary>

```
message("nxdk-pkg-config version:")
execute_process(COMMAND nxdk-pkg-config --version
)
message("nxdk-pkg-config SDL2_image --libs outputs:")
execute_process(COMMAND nxdk-pkg-config SDL2_image --libs
)
message("nxdk-pkg-config SDL2_image --libs --msvc-syntax outputs:")
execute_process(COMMAND nxdk-pkg-config SDL2_image --libs --msvc-syntax
)

message("using CMake's PkgConfig module: pkg_search_module")
pkg_search_module(SDL2IMAGE REQUIRED SDL2_image)
message("SDL2IMAGE libraries: ${SDL2IMAGE_LIBRARIES}")
message("SDL2IMAGE link libraries: ${SDL2IMAGE_LINK_LIBRARIES}")
message("SDL2IMAGE cflags: ${SDL2IMAGE_CFLAGS}")
```
</details>
<details>
<summary>The output will be:</summary>

```
[cmake] nxdk-pkg-config version:
[cmake] 1.8.1
[cmake] nxdk-pkg-config SDL2_image --libs outputs:
[cmake] /usr/src/nxdk/lib/libSDL2_image.lib /usr/src/nxdk/lib/libSDL2.lib /usr/src/nxdk/lib/libjpeg.lib /usr/src/nxdk/lib/libpng.lib /usr/src/nxdk/lib/libzlib.lib 
[cmake] nxdk-pkg-config SDL2_image --libs --msvc-syntax outputs:
[cmake] 
[cmake] using CMake's PkgConfig module: pkg_search_module
[cmake] SDL2IMAGE libraries: 
[cmake] SDL2IMAGE link libraries: 
[cmake] SDL2IMAGE cflags: -I/usr/src/nxdk/lib/sdl/SDL2_image;-I/usr/src/nxdk/lib/sdl/SDL2/include;-DXBOX;-I/usr/src/nxdk/lib/libjpeg/libjpeg-turbo;-I$(NXDK_DIR)/lib/libjpeg;-I/usr/src/nxdk/lib/libpng;-I/usr/src/nxdk/lib/libpng/libpng;-I/usr/src/nxdk/lib/zlib/zlib;-DZ_SOLO
```
</details>
<details>
<summary>If I change SDL2_image.pc to include `-l` before library file. Then it outputs:</summary>

```
[cmake] nxdk-pkg-config version:
[cmake] 1.8.1
[cmake] nxdk-pkg-config SDL2_image --libs outputs:
[cmake] -l/usr/src/nxdk/lib/libSDL2_image.lib /usr/src/nxdk/lib/libSDL2.lib /usr/src/nxdk/lib/libjpeg.lib /usr/src/nxdk/lib/libpng.lib /usr/src/nxdk/lib/libzlib.lib 
[cmake] nxdk-pkg-config SDL2_image --libs --msvc-syntax outputs:
[cmake] /usr/src/nxdk/lib/libSDL2_image.lib.lib 
[cmake] -- Checking for one of the modules 'SDL2_image'
[cmake] using CMake's PkgConfig module: pkg_search_module
[cmake] SDL2IMAGE libraries: /usr/src/nxdk/lib/libSDL2_image.lib
[cmake] SDL2IMAGE link libraries: /usr/src/nxdk/lib/libSDL2_image.lib
[cmake] SDL2IMAGE cflags: -I/usr/src/nxdk/lib/sdl/SDL2_image;-I/usr/src/nxdk/lib/sdl/SDL2/include;-DXBOX;-I/usr/src/nxdk/lib/libjpeg/libjpeg-turbo;-I$(NXDK_DIR)/lib/libjpeg;-I/usr/src/nxdk/lib/libpng;-I/usr/src/nxdk/lib/libpng/libpng;-I/usr/src/nxdk/lib/zlib/zlib;-DZ_SOLO
```
</details>
<details>
<summary>So, in summary if not include `-l` before library files will also not output SDL2_image including whole library dependencies. With the correction from PR outputs:</summary>

```
[cmake] nxdk-pkg-config version:
[cmake] 1.8.1
[cmake] nxdk-pkg-config SDL2_image --libs outputs:
[cmake] -l/usr/src/nxdk/lib/libSDL2_image.lib -l/usr/src/nxdk/lib/libSDL2.lib -l/usr/src/nxdk/lib/libjpeg.lib -l/usr/src/nxdk/lib/libpng.lib -l/usr/src/nxdk/lib/libzlib.lib
[cmake] nxdk-pkg-config SDL2_image --libs --msvc-syntax outputs:
[cmake] /usr/src/nxdk/lib/libSDL2_image.lib.lib /usr/src/nxdk/lib/libSDL2.lib.lib /usr/src/nxdk/lib/libjpeg.lib.lib /usr/src/nxdk/lib/libpng.lib.lib /usr/src/nxdk/lib/libzlib.lib.lib 
[cmake] using CMake's PkgConfig module: pkg_search_module
[cmake] SDL2IMAGE libraries: /usr/src/nxdk/lib/libSDL2_image.lib;/usr/src/nxdk/lib/libSDL2.lib;/usr/src/nxdk/lib/libjpeg.lib;/usr/src/nxdk/lib/libpng.lib;/usr/src/nxdk/lib/libzlib.lib
[cmake] SDL2IMAGE link libraries: /usr/src/nxdk/lib/libSDL2_image.lib;/usr/src/nxdk/lib/libSDL2.lib;/usr/src/nxdk/lib/libjpeg.lib;/usr/src/nxdk/lib/libpng.lib;/usr/src/nxdk/lib/libzlib.lib
[cmake] SDL2IMAGE cflags: -I/usr/src/nxdk/lib/sdl/SDL2_image;-I/usr/src/nxdk/lib/sdl/SDL2/include;-DXBOX;-I/usr/src/nxdk/lib/libjpeg/libjpeg-turbo;-I/usr/src/nxdk/lib/libjpeg;-I/usr/src/nxdk/lib/libpng;-I/usr/src/nxdk/lib/libpng/libpng;-I/usr/src/nxdk/lib/zlib/zlib;-DZ_SOLO
```
</details>

TLDR version: Alpine docker's pkgconf didn't output the libraries with `--msvc-syntax`. With the correction from this pull request, it does. 🤷 I think it's better to communicate with pkgconf developers for better understanding the specification of *.pc files for `libs:` field.

**Comment by thrimbor** [comment]
> TLDR version: Alpine docker's pkgconf didn't output the libraries with `--msvc-syntax`. With the correction from this pull request, it does. 🤷 I think it's better to communicate with pkgconf developers for better understanding the specification of *.pc files for `libs:` field.

There seems to be a misunderstanding here, let me clarify what I meant.

The way we currently have things set up makes pkg-config emit link.exe-style parameters which is fine for using them directly or with meson, but doesn't work with CMake because it expects to see ld-style parameters.
With your changes CMake is happy because it now gets ld-style parameters, but direct usage doesn't work (and I assume meson doesn't either).
With `--msvc-syntax`, we can keep your changes and still get link.exe-style parameters because pkg-config will convert them for us when passing that. For direct usage the user will have to specify that parameter, for meson we can probably add it to the toolchain file.

The only requirement is that we have a compatible implementation of pkg-config. On our docker image that's not a problem because we control what software it contains, and I don't worry about Linux distros either.
GH's runners may be problematic though because they ship with that shitty Perl version.

**Review by thrimbor — COMMENTED** [review]
I've gone through the code, and as far as I can see apart from the bits I'm mentioning here all issues have already been raised by jfr.

File: bin/nxdk-cc [hunk 5]
```diff
@@ -17,6 +17,9 @@ clang \
     -I${NXDK_DIR}/lib/pdclib/platform/xbox/include \
     -I${NXDK_DIR}/lib/winapi \
     -I${NXDK_DIR}/lib/xboxrt/vcruntime \
+    -I${NXDK_DIR}/lib/net/lwip/src/include \
+    -I${NXDK_DIR}/lib/net/nforceif/include \
+    -I${NXDK_DIR}/lib/net/nvnetdrv \
```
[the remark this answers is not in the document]
RadWolfie: When the option of NXDK_NET was dropped in main makefile, These includes came from these lines.
https://github.com/XboxDev/nxdk/blob/da6a0bd73ef141dca46bd69defba482ebdbc0dd3/lib/net/Makefile#L22-L27

Do I know which includes folder actually used for application? No, I didn't have time to check all other applications using nxdk. Since I was too busy figuring out what are the major problems using CMake in codespace environment. I can recheck it later for what's really unnecessary for user-level application development.

share/toolchain-nxdk.cmake
RadWolfie: > Could we also do an `unset` when we're done with it?

That's the plan. I must had forgot to mentioned that in my previous comment.

bin/nxdk-cc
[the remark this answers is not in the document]
thrimbor: You don't need to check all applications. You seem to be testing with NevolutionX, which includes network support, so you can check whether it builds without the `nvnetdrv` and `nforceif` directories, and if it does, you can remove them. I assume only the lwip directory is necessary because that supplies the actual socket layer for the application, the other two should only be necessary for building nxdk itself.

share/toolchain-nxdk.cmake
RadWolfie: > Xiso creation is one of those - it gives very little control over the end result (on-disc layout, non-1:1 directory mapping etc.), and technically isn't required to get something running on the Xbox, so imho we can leave that out the CMake toolchain file. It's still frequently used for quick emulator testing, so my suggestion would be to have a wiki entry "Creating an xiso from CMake" or something that supplies code that can just be dropped in.

[Quote from thrimbor](https://github.com/XboxDev/nxdk/pull/631#issuecomment-1646725220)

In summary, it's better off to perform xiso creation from application preferred target's POST_BUILD process with CMake's add_custom_command function; via nxdk wiki CMake's getting started/howto.

sharkwouter:
The 2 pkg-config related commits in this PR should be cherry picked into master. They fix an issue.

PR #634, opened 2023-05-27
xboxkrnl: add OBJECT_HEADER structure

RadWolfie:
Missing addition of OBJECT_HEADER in nxdk repo will allow me to prepare upload xbox_object_directory_list project on GitHub. Having access to OBJECT_HEADER allow me access to pointer of Type member variable to determine which type of object it is.

### DIFF
--- a/lib/xboxkrnl/xboxdef.h
+++ b/lib/xboxkrnl/xboxdef.h
@@ -41,6 +41,8 @@ typedef unsigned int SIZE_T, *PSIZE_T;
 typedef int BOOL, *PBOOL;
 typedef const char *PCSZ, *PCSTR, *LPCSTR;
 
+typedef ULONGLONG QUAD;
+
 typedef ULONG ULONG_PTR;
 typedef LONG LONG_PTR;
 
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -891,6 +891,16 @@ typedef struct _OBJECT_TYPE
     ULONG PoolTag;
 } OBJECT_TYPE, *POBJECT_TYPE;
 
+typedef struct _OBJECT_HEADER {
+	LONG PointerCount;
+	LONG HandleCount;
+	POBJECT_TYPE Type;
+	ULONG Flags;
+	QUAD Body;
+} OBJECT_HEADER, *POBJECT_HEADER;
+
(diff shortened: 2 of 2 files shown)


lib/xboxkrnl/xboxdef.h
```diff
@@ -41,6 +41,8 @@ typedef unsigned int SIZE_T, *PSIZE_T;
 typedef int BOOL, *PBOOL;
 typedef const char *PCSZ, *PCSTR, *LPCSTR;
 
+typedef unsigned __int64 QUAD;
```
dracc, line 44: What's wrong with `LONGLONG`?
I haven't tested, but `__int64` looks to be a MSVC thing?

RadWolfie: Sorry about that, I had a brain fart when I was doing experimental accessing to objects. But yes, it is MSVC feature. Even weird thing that `__int128` is mentioned on [cppreference.com](https://en.cppreference.com/w/cpp/language/integer_literal) site.

GXTX: We use it already in nxdk bits. <https://github.com/XboxDev/nxdk/blob/a48ed10ae6e6e5affe090677b62d765a09cada9e/lib/winapi/winnt.h#L16>

dracc: @RadWolfie It's an extended integer type, which GCC [doesn't support](https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html). It seems it's not officially supported in clang either, but [prefers `_BitInt(N)` instead](https://clang.llvm.org/docs/LanguageExtensions.html#extended-integer-types).

@GXTX Intriguing. See above.

RadWolfie:
If looking for simpler way to test this. You can use ObOpenObjectByName for `\\??\\D:` path to obtain handle. Then call `ObReferenceObjectByHandle` to obtain object. With the object obtained you can use `OBJECT_TO_OBJECT_HEADER` and test Type pointer, or PoolTag from Type pointer, is the same as `ObSymbolicLinkObjectType`, or `bmyS` for PoolTag.

Ryzee119:
> If looking for simpler way to test this. You can use ObOpenObjectByName for `\\??\\D:` path to obtain handle. Then call `ObReferenceObjectByHandle` to obtain object. With the object obtained you can use `OBJECT_TO_OBJECT_HEADER` and test Type pointer, or PoolTag from Type pointer, is the same as `ObSymbolicLinkObjectType`, or `bmyS` for PoolTag.

Took me a while to decrpyt this test example but worked as expect. Thanks

PR #635, opened 2023-06-09
Add GetLocalTime function

Ryzee119:
This PR adds [GetLocalTime](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getlocaltime). This requires [GetTimeZoneInformation](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-gettimezoneinformation) so is also added in this PR.

`GetTimeZoneInformation` reads the timezone biases, and daylight savings start/end times from the Xbox EEPROM and calculates whether the user is in DST or not and returns a TIME_ZONE_INFORMATION struct with all the timezone information. These EEPROM values are written by MSDash when you set your timezone.

This is used by `GetLocalTime` to determine the users current time based on the SystemTime.

DST is automatically adjusted for when the corresponding setting in MSDash is enabled (and your timezone supports it).

As this relies on hardcoded time info within MSDash that is ~20 years old some discrepencies exist with modern timezones and daylight savings. But I have ran this against about 10 timezones and the output matches what I expect given the inputs.

This is admittedly hard to test for all cases but to try get some edge cases I ran a loop that increments time by 30 minutes over a whole year over a few timezones to confirm the transitions to DST occur at the correct times which outputs as expected.

This simple test will print your current timezone info and localtime (Ran with `xemu.exe -device lpc47m157 -serial stdio` and tested on my 1.6 Xbox.)
```
const CHAR *tz_strs[] = {"TIME_ZONE_ID_UNKNOWN", "TIME_ZONE_ID_STANDARD", "TIME_ZONE_ID_DAYLIGHT"};

TIME_ZONE_INFORMATION tzInfo;
DWORD result = GetTimeZoneInformation(&tzInfo);
DbgPrint("TIMEZONE RESULT %s\n", tz_strs[result]);
DbgPrint("Time Zone Information:\n");
DbgPrint("Standard Name: %ls\n", tzInfo.StandardName);
DbgPrint("Daylight Name: %ls\n", tzInfo.DaylightName);
DbgPrint("Bias: %ld minutes\n", tzInfo.Bias);
DbgPrint("Standard Bias: %ld minutes\n", tzInfo.StandardBias);
DbgPrint("Daylight Bias: %ld minutes\n", tzInfo.DaylightBias);
DbgPrint("Standard Date: Month: %ld, Day: %ld, Time: %02ld:%02ld:%02ld\n",
       tzInfo.StandardDate.wMonth, tzInfo.StandardDate.wDay,
       tzInfo.StandardDate.wHour, tzInfo.StandardDate.wMinute, tzInfo.StandardDate.wSecond);
DbgPrint("Daylight Date: Month: %ld, Day: %ld, Time: %02ld:%02ld:%02ld\n",
       tzInfo.DaylightDate.wMonth, tzInfo.DaylightDate.wDay,
       tzInfo.DaylightDate.wHour, tzInfo.DaylightDate.wMinute, tzInfo.DaylightDate.wSecond);

SYSTEMTIME st;
GetLocalTime(&st);
DbgPrint("Local time %04d-%02d-%02d %02d:%02d:%02d\n",
    st.wYear, st.wMonth, st.wDay, st.wHour, st.wMinute, st.wSecond);
```

This should output something like this:
```
TIMEZONE RESULT TIME_ZONE_ID_STANDARD
Time Zone Information:
Standard Name: ACST
Daylight Name: ACDT
Bias: -570 minutes
Standard Bias: 0 minutes
Daylight Bias: -60 minutes
Standard Date: Month: 3, Day: 5, Time: 02:00:00
Daylight Date: Month: 10, Day: 5, Time: 02:00:00
Local time 2023-06-09 09:02:50
```

### DIFF
--- a/lib/winapi/timezoneapi.c
+++ b/lib/winapi/timezoneapi.c
@@ -1,0 +1,207 @@
+// SPDX-License-Identifier: MIT
+
+// SPDX-FileCopyrightText: 2023 Ryan Wendland
+
+#include <timezoneapi.h>
+#include <winbase.h>
+#include <assert.h>
+#include <xboxkrnl/xboxkrnl.h>
+#include <string.h>
+
+typedef struct
+{
+    UCHAR Month;     // 1-12, a 0 indicates there is no timezone information
+    UCHAR Day;       // 1 = 1st occurrence of DayOfWeek up to 5 (5th or last)
+    UCHAR DayOfWeek; // 0 = Sunday to 6 = Saturday
+    UCHAR Hour;
+} XBOX_TZ_STRUCT;
+
+// Determine the day of the week given a year, month and day
+// https://en.wikipedia.org/wiki/Determination_of_the_day_of_the_week#Methods_in_computer_code
+static UCHAR GetDayOfWeek (SHORT year, UCHAR month, UCHAR day)
+{
+    return (day += month < 3 ? year-- : year - 2, 23 *
+            month / 9 + day + 4 + year / 4 - year / 100 + year / 400) % 7;
+}
+
+static UCHAR GetDaysInMonth (SHORT year, UCHAR month)
+{
+    static const UCHAR daysPerMonth[] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
+    UCHAR days;
+
+    assert(month >= 1);
+    assert(month <= 12);
+
+    if (month == 2) {
+        if ((year % 4 == 0 && year % 100 != 0) || (year % 400 == 0)) {
+            days = 29;
+        } else {
+            days = 28;
+        }
+    } else {
+        days = daysPerMonth[month - 1];
+    }
+
+    return days;
+}
+
+// Given a year, month, day of the week, and the n-th occurrence this returns the day in the month
+// i.e The 2nd Wednesday in November 2023 is the 8th November. This function would return 8
+static UCHAR GetDayOfMonth (SHORT year, XBOX_TZ_STRUCT *tzInfo)
+{
+    UCHAR dayOfWeek = GetDayOfWeek(year, tzInfo->Month, 1);
+    UCHAR daysInMonth = GetDaysInMonth(year, tzInfo->Month);
+    UCHAR dayTargetCount = tzInfo->Day - 1;
+    UCHAR dayActual = 1;
+
+    // Find first occurrence of the weekday
+    dayActual += (tzInfo->DayOfWeek < dayOfWeek) ? (tzInfo->DayOfWeek + 7 - dayOfWeek) : (tzInfo->DayOfWeek - dayOfWeek);
+
+    // Add remaining weeks
+    dayActual += dayTargetCount * 7;
+
+    // If we overrun, go back to the last occurrence in the month
+    while (dayActual > daysInMonth) {
+        dayActual -= 7;
+    }
+    return dayActual;
+}
+
+static void XboxTimeZoneToSystemTime (LPSYSTEMTIME lpSystemTime, XBOX_TZ_STRUCT *tzInfo)
+{
+    lpSystemTime->wMonth = tzInfo->Month;
+    lpSystemTime->wDayOfWeek = tzInfo->DayOfWeek;
+    lpSystemTime->wDay = tzInfo->Day;
+    lpSystemTime->wHour = tzInfo->Hour;
+}
+
+static BOOL XboxTimeZoneValid (XBOX_TZ_STRUCT *tzInfo)
+{
+    if (tzInfo->Month > 12 || tzInfo->DayOfWeek > 6 || tzInfo->Hour > 23) {
+        return FALSE;
+    }
+    return TRUE;
+}
+
+DWORD GetTimeZoneInformation (LPTIME_ZONE_INFORMATION lpTimeZoneInformation)
+{
+    assert(lpTimeZoneInformation != NULL);
+
+    LONG timeZoneBias, daylightBias, standardBias, daylightDisableFlag;
+    LONG timeNow, daylightStart, daylightEnd;
+    XBOX_TZ_STRUCT daylightStartDate, daylightEndDate;
+    CHAR daylightName[4], standardName[4];
+    NTSTATUS queryStatus;
+    TIME_FIELDS dateNow;
+    LARGE_INTEGER kTime;
+    ULONG type;
+
+    memset(lpTimeZoneInformation, 0, sizeof(TIME_ZONE_INFORMATION));
+
+    queryStatus = ExQueryNonVolatileSetting(XC_TIMEZONE_BIAS, &type, &timeZoneBias, sizeof(timeZoneBias), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    lpTimeZoneInformation->Bias = timeZoneBias;
+
+    // Check if daylight savings time (DST) adjustment is enabled
+    queryStatus = ExQueryNonVolatileSetting(XC_MISC, &type, &daylightDisableFlag, sizeof(daylightDisableFlag), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    // DST adjustment is disabled
+    if (daylightDisableFlag & XC_MISC_FLAG_DISABLE_DST) {
+        return TIME_ZONE_ID_UNKNOWN;
+    }
+
+    // Query all DST info from EEPROM
+    // FIXME: One large EEPROM access could be better
+    queryStatus = ExQueryNonVolatileSetting(XC_TZ_DLT_DATE, &type, &daylightStartDate, sizeof(daylightStartDate), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+    queryStatus = ExQueryNonVolatileSetting(XC_TZ_STD_DATE, &type, &daylightEndDate, sizeof(daylightEndDate), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+    queryStatus = ExQueryNonVolatileSetting(XC_TZ_DLT_BIAS, &type, &daylightBias, sizeof(daylightBias), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+    queryStatus = ExQueryNonVolatileSetting(XC_TZ_STD_BIAS, &type, &standardBias, sizeof(standardBias), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+    queryStatus = ExQueryNonVolatileSetting(XC_TZ_DLT_NAME, &type, daylightName, sizeof(daylightName), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+    queryStatus = ExQueryNonVolatileSetting(XC_TZ_STD_NAME, &type, standardName, sizeof(standardName), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    if ((XboxTimeZoneValid(&daylightStartDate) == FALSE) || (XboxTimeZoneValid(&daylightEndDate) == FALSE)) {
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    // Xbox stores up to 4 characters here
+    for (UCHAR i = 0; i < 4; i++) {
+        lpTimeZoneInformation->StandardName[i] = standardName[i];
+        lpTimeZoneInformation->DaylightName[i] = daylightName[i];
+    }
+
+    // There was no DST info
+    if (daylightEndDate.Month == 0 || daylightStartDate.Month == 0) {
+        return TIME_ZONE_ID_UNKNOWN;
+    }
+
+    XboxTimeZoneToSystemTime(&lpTimeZoneInformation->DaylightDate, &daylightStartDate);
+    XboxTimeZoneToSystemTime(&lpTimeZoneInformation->StandardDate, &daylightEndDate);
+    lpTimeZoneInformation->DaylightBias = daylightBias;
+    lpTimeZoneInformation->StandardBias = standardBias;
+
+    // Now we need to determine if the current time is within the DST range
+    KeQuerySystemTime(&kTime);
+    RtlTimeToTimeFields(&kTime, &dateNow);
+
+    // Determine what day in the month DST starts and ends
+    UCHAR daylightStartDay = GetDayOfMonth(dateNow.Year, &daylightStartDate);
+    UCHAR daylightEndDay = GetDayOfMonth(dateNow.Year, &daylightEndDate);
+
+    // Fix month wrapping. i.e if DST is from Oct(10) to Feb(2) and the current month is Jan(1) we should be in DST
+    // Applying a simple check would result is 10 < 1 < 2 = false and incorrect
+    // To account for this, we offset the months in the following year by 12
+    // This results in the comparison 10 < 13(12+1) < 14(12+2) = true and is correct
+    if (daylightEndDate.Month < daylightStartDate.Month) {
+        if (dateNow.Month <= daylightEndDate.Month) {
+            dateNow.Month += 12;
+        }
+        daylightEndDate.Month += 12;
+    }
+
+    // Now that month wrapping is fixed, we convert the dates to minutes from the start of the year corresponding to daylightStartDate
+    // The conversion to minutes allows for easier comparison. The minute resolution allows for timezones with 0.5 hour offsets.
+    const LONG mpd = 24 * 60;  // Minutes per day
+    const LONG mpm = 31 * mpd; // Minutes per month. 31 is ok providing it's consistent
+    timeNow = (dateNow.Month * mpm) + (dateNow.Day * mpd) + (dateNow.Hour * 60) + dateNow.Minute;
+    // DST change overs happen in local time, so we need to add the respective biases too
+    daylightStart = (daylightStartDate.Month * mpm) + (daylightStartDay * mpd) + (daylightStartDate.Hour * 60) + timeZoneBias + standardBias;
+    daylightEnd = (daylightEndDate.Month * mpm) + (daylightEndDay * mpd) + (daylightEndDate.Hour * 60) + timeZoneBias + daylightBias;
+
+    if (timeNow >= daylightStart && timeNow < daylightEnd) {
+        return TIME_ZONE_ID_DAYLIGHT;
+    } else {
+        return TIME_ZONE_ID_STANDARD;
+    }
+}
--- a/lib/winapi/timezoneapi.h
+++ b/lib/winapi/timezoneapi.h
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: MIT
+
+// SPDX-FileCopyrightText: 2023 Ryan Wendland
+
+#ifndef __TIMEZONEAPI_H__
+#define __TIMEZONEAPI_H__
+
+#include <minwinbase.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define TIME_ZONE_ID_UNKNOWN 0
+#define TIME_ZONE_ID_STANDARD 1
+#define TIME_ZONE_ID_DAYLIGHT 2
+#define TIME_ZONE_ID_INVALID ((DWORD)0xFFFFFFFF)
+
+typedef struct _TIME_ZONE_INFORMATION {
+    LONG Bias;
+    WCHAR StandardName[32];
+    SYSTEMTIME StandardDate;
+    LONG StandardBias;
+    WCHAR DaylightName[32];
+    SYSTEMTIME DaylightDate;
+    LONG DaylightBias;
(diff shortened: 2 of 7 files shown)


lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,194 @@
...
+    KeQuerySystemTime(&kTime);
+    RtlTimeToTimeFields(&kTime, &dateNow);
+
+    // Determine what day in the month DST starts and ends
+    UCHAR endDay = GetActualDay(dateNow.Year, &dayLightDate);
+    UCHAR startDay = GetActualDay(dateNow.Year, &standardDate);
+
+    // Fix month wrapping. i.e if DST is Oct to Feb and current month is Jan we should be in DST.
+    // Month range is 10-2, this becomes 10-14
+    // Current month becomes 13. So we compare 10 < 13 < 14 so it all works
```
JayFoxRox, line 171: I don't understand this comment. Currently it's switching between months and numbers (which aren't explained).

I think you should add numbers to the mentioned months  (`10[Oct] - 2[Feb], this becomes [WHY?!] 10[Oct]-14[Feb+???]`). Would also be fine to explain what each number represents first (instead of doing it inline).

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,194 @@
...
+    if (standardDate.Month < dayLightDate.Month)
+    {
+        if (dateNow.Month <= standardDate.Month) {
+            dateNow.Month += 12;
+        }
+        standardDate.Month += 12;
```
JayFoxRox, line 177: It's not clear why `.Month += 12` is used. Wouldn't that mean the Month is out of the 1-12 range? What is being calculated here?

Ryzee119: We ultimately need to determine if Start of DST < now < End of DST. 

In the case of DST going over new  years may want to calculate if January(1) is between Dec(12) and Feb(2); a simple range check would give us 12<1<2 which is comes out false which is wrong.

The +12 just handles this wrap around business, so ultimately we compare 12<13<14 which is correct.
This feeds into the calculation futher down which also has to consider Days, hours and minutes but only the month wrap around is the main problem.

Effectively, this just considers January the 13th Month  and Feb 14th Month since the start of the previous year

lib/winapi/timezoneapi.c
Ryzee119: I have reworded this comment to hopefully make it clearer. Thanks

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,196 @@
...
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    // DST adjustment is disabled.
+    if (dayLightEnFlag & 0x2) {
+        return TIME_ZONE_ID_UNKNOWN;
+    }
+
+    // Query all DST info from EEPROM. //FIXME. One large EEPROM access could be better
```
JayFoxRox, line 100: FIXME should be on next line?

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,196 @@
...
+            days = 28;
+        }
+    } else {
+        days = daysPerMonth[month - 1];
+    }
+
+    return days;
+}
+
+// Given a year, month, day of the week, and the nth occurrence this returns the day in the month.
```
JayFoxRox, line 42: nit: `n-th` usually looks better imho

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,196 @@
...
+    lpTimeZoneInformation->Bias = timeZoneBias;
+
+    // Check if DST adjustment is enabled
+    queryStatus = ExQueryNonVolatileSetting(XC_MISC, &type, &dayLightEnFlag, sizeof(dayLightEnFlag), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    // DST adjustment is disabled.
```
JayFoxRox, line 95: Some of the comments end with a period symbol, others don't. Copy/paste bug?

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,196 @@
...
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    // Xbox stores up to 4 characters here.
+    for (UCHAR i = 0; i < 4; i++) {
+        lpTimeZoneInformation->StandardName[i] = standardName[i];
+        lpTimeZoneInformation->DaylightName[i] = dayLightName[i];
+    }
+
+    // There was no Daylight savings date info
```
JayFoxRox, line 144: Sometimes it's "Daylight savings", and sometimes it's "day light savings".

Should probably be "daylight savings" ?

lib/winapi/timezoneapi.c
JayFoxRox: Yes, thanks. This is much better :)

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,196 @@
...
+    lpTimeZoneInformation->DaylightDate.wHour = dayLightDate.Hour;
+    lpTimeZoneInformation->DaylightBias = dayLightBias;
+
+    // Now we need to determine if the current time is within the day light savings range
+    KeQuerySystemTime(&kTime);
+    RtlTimeToTimeFields(&kTime, &dateNow);
+
+    // Determine what day in the month DST starts and ends
+    UCHAR startDay = GetActualDay(dateNow.Year, &dayLightDate);
+    UCHAR endDay = GetActualDay(dateNow.Year, &standardDate);
```
JayFoxRox, line 167: nit: The variable names for `dayLightDate` and `standardDate` confused me in the section below and also make these names inconsistent (should be `dayLightDay` and `standardDay`).

However, I think having `daylightStartDay` from `daylightStartDate` & `daylightEndDay` from `daylightStartDate` would make this entire function clearer, even if this conflicts with the Xbox EEPROM naming convention.

lib/winapi/timezoneapi.c
Ryzee119: Done

lib/winapi/timezoneapi.c
Ryzee119: Done

lib/winapi/timezoneapi.c
Ryzee119: I just removed all full stops so its consistent. Looking at some other files we didnt seem to use them often.

lib/winapi/timezoneapi.c
Ryzee119: Made it daylight savings time but now use the abbreviation DST is most places now

lib/winapi/timezoneapi.c
Ryzee119: I think ive done this. Its now all daylightStartDate, daylightEndDate, daylightStartDay, daylightEndDay etc.

lib/winapi/timezoneapi.c
JayFoxRox: new nit (sorry, didn't notice earlier): most of our existing FIXME start with `// FIXME: ` [space after `//`, and colon after `FIXME`]

lib/winapi/timezoneapi.c
JayFoxRox: nit: Would have preferred `daylight` over `dayLight` to be consistent with existing Xbox names (`Daylight`). But it's just nitpicking, so you can resolve this conversation (with or without change).

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+    const LONG MPM = 31 * MPD; //  31 is ok providing it's consistent
+    timeNow = (dateNow.Month * MPM) + (dateNow.Day * MPD) + (dateNow.Hour * 60) + dateNow.Minute;
+    // DST change overs happen in local time, so we need to add the respective biases too
+    dayLightStart = (dayLightStartDate.Month * MPM) + (dayLightStartDay * MPD) + (dayLightStartDate.Hour * 60) + timeZoneBias + standardBias;
+    dayLightEnd = (dayLightEndDate.Month * MPM) + (dayLightEndDay * MPD) + (dayLightEndDate.Hour * 60) + timeZoneBias + dayLightBias;
+
+    if (timeNow >= dayLightStart && timeNow < dayLightEnd) {
+        return TIME_ZONE_ID_DAYLIGHT;
+    }
+    else {
```
JayFoxRox, line 194: nit: Formatting in other files (also within the nxdk winapi) has `} else {`. Not sure about winapi or if there's a code convention for nxdk now (clang-format?)

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+
+    // Find first occurrence of the weekday
+    dayActual += (tzInfo->DayOfWeek < dayOfWeek) ? (tzInfo->DayOfWeek + 7 - dayOfWeek) : (tzInfo->DayOfWeek - dayOfWeek);
+
+    // Add remaining weeks
+    dayActual += dayTargetCnt * 7;
+
+    // If we overrun, go back to the last occurrence in the month
+    while (dayActual > daysinMonth)
+    {
```
JayFoxRox, line 59: nit: I believe we usually put `while (...) {` in a single line

lib/winapi/timezoneapi.h
```diff
@@ -0,0 +1,35 @@
...
+
+typedef struct _TIME_ZONE_INFORMATION {
+  LONG       Bias;
+  WCHAR      StandardName[32];
+  SYSTEMTIME StandardDate;
+  LONG       StandardBias;
+  WCHAR      DaylightName[32];
+  SYSTEMTIME DaylightDate;
+  LONG       DaylightBias;
+} TIME_ZONE_INFORMATION, *PTIME_ZONE_INFORMATION, *LPTIME_ZONE_INFORMATION;
```
JayFoxRox, line 27: nit: The source file has 4 spaces for indentation, this has 2 spaces

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+        if (dateNow.Month <= dayLightEndDate.Month) {
+            dateNow.Month += 12;
+        }
+        dayLightEndDate.Month += 12;
+    }
+
+    // Now that month wrapping is fixed, we convert the dates to minutes from the start of the year corresponding to dayLightStartDate
+    // The conversion to minutes allows for easier comparison. The minute resolution allows for timezones with 0.5 hour offsets.
+    const LONG MPD = 24 * 60;
+    const LONG MPM = 31 * MPD; //  31 is ok providing it's consistent
```
JayFoxRox, line 185: nit: upper-case variable names should probably be reserved for macros.

It's also not clear what MPD and MPM stand for (although it's probably minutes-per-day and minutes-per-month)

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+// https://en.wikipedia.org/wiki/Determination_of_the_day_of_the_week#Methods_in_computer_code
+static UCHAR GetDayofWeek(SHORT y, UCHAR m, UCHAR d) {
+    return (d += m < 3 ? y-- : y - 2, 23 * m / 9 + d + 4 + y / 4 - y / 100 + y / 400) % 7;
+}
+
+static UCHAR GetDaysinMonth(SHORT year, UCHAR month) {
+    static const UCHAR daysPerMonth[] = {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31};
+    UCHAR days;
+
+    assert (month <= 12);
```
JayFoxRox, line 27: This probably also has to `assert(month >= 1)`, otherwise the index access below could crash.

nit: I don't think we usually do space after `assert`

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+    UCHAR Hour;
+} XBOX_TZ_STRUCT;
+
+// Determine the day of the week given a year, month and day
+// https://en.wikipedia.org/wiki/Determination_of_the_day_of_the_week#Methods_in_computer_code
+static UCHAR GetDayofWeek(SHORT y, UCHAR m, UCHAR d) {
+    return (d += m < 3 ? y-- : y - 2, 23 * m / 9 + d + 4 + y / 4 - y / 100 + y / 400) % 7;
+}
+
+static UCHAR GetDaysinMonth(SHORT year, UCHAR month) {
```
JayFoxRox, line 23: `GetDaysInMonth`? (camel-case `In`)

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+typedef struct {
+    UCHAR Month;     // 1-12
+    UCHAR Day;       // 1 = 1st occurrence of DayOfWeek up to 5 (5th or last)
+    UCHAR DayOfWeek; // 0 = Sunday to 6 = Saturday
+    UCHAR Hour;
+} XBOX_TZ_STRUCT;
+
+// Determine the day of the week given a year, month and day
+// https://en.wikipedia.org/wiki/Determination_of_the_day_of_the_week#Methods_in_computer_code
+static UCHAR GetDayofWeek(SHORT y, UCHAR m, UCHAR d) {
```
JayFoxRox, line 19: `GetDayOfWeek`? (camel-case `Of`)

JayFoxRox, line 19: nit: `GetDaysinMonth` functions use `year`, `month` spelled out; should be more consistent?

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+    return days;
+}
+
+// Given a year, month, day of the week, and the n-th occurrence this returns the day in the month
+// i.e The 2nd Wednesday in November 2023 is the 8th November. This function would return 8
+static UCHAR GetActualDay(SHORT year, XBOX_TZ_STRUCT *tzInfo)
+{
+    UCHAR dayOfWeek = GetDayofWeek(year, tzInfo->Month, 1);
+    UCHAR daysinMonth = GetDaysinMonth(year, tzInfo->Month);
+    UCHAR dayTargetCnt = tzInfo->Day - 1;
```
JayFoxRox, line 48: nit: Prefer `Count` spelled out

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+
+    return days;
+}
+
+// Given a year, month, day of the week, and the n-th occurrence this returns the day in the month
+// i.e The 2nd Wednesday in November 2023 is the 8th November. This function would return 8
+static UCHAR GetActualDay(SHORT year, XBOX_TZ_STRUCT *tzInfo)
+{
+    UCHAR dayOfWeek = GetDayofWeek(year, tzInfo->Month, 1);
+    UCHAR daysinMonth = GetDaysinMonth(year, tzInfo->Month);
```
JayFoxRox, line 47: `daysInMonth` (camel-case `In`); it's already camel-cased in `dayOfWeek`

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    if (dayLightEndDate.Month > 12 || dayLightStartDate.Month > 12 ||
+        dayLightEndDate.DayOfWeek > 6 || dayLightStartDate.DayOfWeek > 6 ||
+        dayLightEndDate.Hour > 23 || dayLightStartDate.Hour > 23) {
+        return TIME_ZONE_ID_INVALID;
+    }
```
JayFoxRox, line 137: This would also have to check if `Month < 1`?

nit: You might also want to move this into a helper function to check if a date is valid.

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+
+    // There was no DST info
+    if (dayLightEndDate.Month == 0 || dayLightStartDate.Month == 0) {
+        return TIME_ZONE_ID_UNKNOWN;
+    }
+
+    lpTimeZoneInformation->StandardDate.wMonth = dayLightEndDate.Month;
+    lpTimeZoneInformation->StandardDate.wDayOfWeek = dayLightEndDate.DayOfWeek;
+    lpTimeZoneInformation->StandardDate.wDay = dayLightEndDate.Day;
+    lpTimeZoneInformation->StandardDate.wHour = dayLightEndDate.Hour;
```
JayFoxRox, line 153: nit: Might want to copy this into something which copies dates, or use the MS type?

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,197 @@
...
+    LONG timeZoneBias, dayLightBias, standardBias, dayLightEnFlag;
+    LONG timeNow, dayLightStart, dayLightEnd;
+    XBOX_TZ_STRUCT dayLightStartDate, dayLightEndDate;
+    CHAR dayLightName[4], standardName[4];
+    NTSTATUS queryStatus;
+    TIME_FIELDS dateNow;
+    LARGE_INTEGER kTime;
+    ULONG type;
+
+    RtlZeroMemory(lpTimeZoneInformation, sizeof(TIME_ZONE_INFORMATION));
```
JayFoxRox, line 78: nit: I believe we mostly use `memset`, because it can be optimized better (unlike a kernel import)

lib/winapi/timezoneapi.c
Ryzee119: When month == 0 does not return TIME_ZONE_ID_INVALID, it returns TIME_ZONE_ID_UNKNOWN.

So the check for month ==0 is done further down in the code. On xbox it also appears to still attempt to copy the standardName and dayLightName even if month == 0 so my code does that too.

lib/winapi/timezoneapi.c
Ryzee119: Done

lib/winapi/timezoneapi.c
Ryzee119: Done, all daylight

lib/winapi/timezoneapi.c
Ryzee119: missed this formatting issue. Done

lib/winapi/timezoneapi.c
Ryzee119: Done

lib/winapi/timezoneapi.h
Ryzee119: Done

lib/winapi/timezoneapi.c
Ryzee119: made lower case. Yes minutes per day and per month

lib/winapi/timezoneapi.c
Ryzee119: done

lib/winapi/timezoneapi.c
Ryzee119: done

lib/winapi/timezoneapi.c
Ryzee119: done

Ryzee119: Done

lib/winapi/timezoneapi.c
Ryzee119: done

lib/winapi/timezoneapi.c
Ryzee119: done

lib/winapi/timezoneapi.c
Ryzee119: moved it to a helper function

lib/winapi/timezoneapi.c
Ryzee119: split into a function

lib/winapi/timezoneapi.c
Ryzee119: done

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,205 @@
...
+    // Determine what day in the month DST starts and ends
+    UCHAR daylightStartDay = GetActualDay(dateNow.Year, &daylightStartDate);
+    UCHAR daylightEndDay = GetActualDay(dateNow.Year, &daylightEndDate);
+
+    // Fix month wrapping. i.e if DST is from Oct(10) to Feb(2) and the current month is Jan(1) we should be in DST
+    // Applying a simple check would result is 10 < 1 < 2 = false and incorrect
+    // To account for this, we offset the months in the following year by 12
+    // This results in the comparison 10 < 13(12+1) < 14(12+2) = true and is correct
+    if (daylightEndDate.Month < daylightStartDate.Month)
+    {
```
JayFoxRox, line 184: nit: Inconsistent wether `{` is on new line

Ryzee119: done

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,206 @@
...
+// SPDX-FileCopyrightText: 2023 Ryan Wendland
+
+#include <timezoneapi.h>
+#include <winbase.h>
+#include <assert.h>
+#include <xboxkrnl/xboxkrnl.h>
+#include <string.h>
+
+typedef struct {
+    UCHAR Month;     // 1-12
```
mborgerson, line 12: Can you add a note here that `0` is a valid value representing unknown time zone

mborgerson (changes requested):

lib/winapi/timezoneapi.c
Ryzee119: Note added

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,206 @@
...
+
+    // Check if daylight savings time (DST) adjustment is enabled
+    queryStatus = ExQueryNonVolatileSetting(XC_MISC, &type, &daylightEnFlag, sizeof(daylightEnFlag), NULL);
+    if (!NT_SUCCESS(queryStatus)) {
+        SetLastError(RtlNtStatusToDosError(queryStatus));
+        return TIME_ZONE_ID_INVALID;
+    }
+
+    // DST adjustment is disabled
+    if (daylightEnFlag & 0x2) {
```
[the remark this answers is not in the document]
Ryzee119: Ive added a define `XC_MISC_FLAG_DISABLE_DST` in `xboxkrnl.h`, as I consider this the most appropriate location.

lib/winapi/timezoneapi.c
```diff
@@ -0,0 +1,206 @@
...
+    } else {
+        days = daysPerMonth[month - 1];
+    }
+
+    return days;
+}
+
+// Given a year, month, day of the week, and the n-th occurrence this returns the day in the month
+// i.e The 2nd Wednesday in November 2023 is the 8th November. This function would return 8
+static UCHAR GetActualDay (SHORT year, XBOX_TZ_STRUCT *tzInfo)
```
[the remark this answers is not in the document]
Ryzee119: Agree this is better, Ive changed this to `GetDayOfMonth`.

mborgerson:
Nice, thanks!

PR #636, opened 2023-06-10
Implement `direct.h` functions

### LINKED ISSUE #466: Missing direct.h contents
We do have a direct.h already, but it's empty.

I currently use this hack:

```
echo "#pragma once // Generated" > $NXDK_DIR/lib/xboxrt/libc_extensions/direct.h
echo "int _chdir(const char *dirname) { return 0; }" >> $NXDK_DIR/lib/xboxrt/libc_extensions/direct.h
echo "int _mkdir(const char *dirname) { return 0; }" >> $NXDK_DIR/lib/xboxrt/libc_extensions/direct.h
```

I hope that someone will implement this properly.

There's more functionality in this, but this is enough for my current projects

ExoSkye:
`fnmerge` and `fnsplit` haven't been implemented since I noticed inconsistencies in the documentation relating to them, for now I simply have empty function definitions in the form described by  https://www.digitalmars.com/rtl/direct.html (ie no `_` prefix)

 The status of other functions is as follows:
`_chdir`, `_chdrive`: impossible to implement, succeeds without an error code, or change of state in the Xbox 
`_getcwd`, `_getwd`: impossible to implement, return NULL pointers and sets `errno` to `-EINVAL` 
`_getdrive`: as before, impossible to implement - returns `0` and sets `errno` to `-EINVAL` (this is the error state described by a combo of MS public documentation and digital mars docs) 
`_searchpath`: impossible to implement since there is no PATH variable on the Xbox 
`_mkdir`, `_rmdir`: implemented, testing required, however

`MAXPATH`, `MAXDRIVE` defined, however these may be incorrect right now (they were taken from DOS-era documentation)

Should fix #466 when complete


ExoSkye:
I've since been informed that `getcwd` and `getwd` should be possible to implement by Andy - so I'll implement those when I get home

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,108 @@
...
+int _chdir(char* path) {
+    return 0;
+}
+
+int _chdrive(int drive) {
+    return 0;
+}
+
+/*
+ * The Xbox has no concept of current working directory, so these can't work
```
JayFoxRox, line 30: Pretty sure it does?

If not, we could still implement it. After all, the libc is implemented in nxdk.

So while this wouldn't work for kernel functions (although it might, because they must also do relative or absolute access at some point - both of which could be remapped), it would still work for the libc or our winapi.

I'm fine with stubbing these for now, but I think there's room for improvement.

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,108 @@
...
+    return NULL;
+}
+
+int _getdrive(void) {
+    errno = -EINVAL;
+    return 0;
+}
+
+/*
+ * There's no path variable on the Xbox, so this can't work
```
JayFoxRox, line 49: The linked docs (although not really a good source) say the cwd is searched first.

Is this function ever used though?
For functions which aren't really used, I think it's better to leave them out, rather than broken stubs.

We should implement (or stub, if necessary) those functions which are actually used in software that people want to use / port.

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,108 @@
...
+
+#include <errno.h>
+#include <string.h>
+#include <stdbool.h>
+#include <windows.h>
+
+// Made referencing https://www.digitalmars.com/rtl/direct.html, retrieved 2023-06-10, copyrighted 1999-2018 by Digital Mars
+
+#define MAXPATH 80
+#define MAXDRIVE 3
```
JayFoxRox, line 15: I assume `MAXDRIVE` would be much higher, because Xbox likes to assign up to Z.
How does Microsoft implement it for modern Windows?

ExoSkye: This is the character limit: it's 3 characters because it includes the nullbyte, the colon and the drive letter

lib/xboxrt/libc_extensions/direct.c
ExoSkye: I'm pretty sure it doesn't, based on what RadWolfie said and also that I can't find any kind of reference to it in the nxdk source tree currently

But it would be good to implement it eventually

lib/xboxrt/libc_extensions/direct.c
ExoSkye: Oh shit, I didn't realise that, I guess I need to do more research into it

I've never see any usage of this in the wild, however

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+void fnmerge(char* path, const char* drive, const char* dir, const char* name, const char* ext) {
+
+}
+
+void fnsplit(const char* path, char* drive, char* dir, char* name, char* ext) {
+
+}
```
RadWolfie, line 108: NIT: Missing newline at end of file, otherwise git isn't praising you.

lib/xboxrt/libc_extensions/direct.h
```diff
@@ -3,3 +3,13 @@
...
+
+int _chdir(char* path);
+int _chdrive(int drive);
+char* _getcwd(char* buffer, size_t length);
+char* _getwd(char* path_name);
+int _getdrive(void);
+int _mkdir(const char* pathname);
+int _rmdir(const char* pathname);
+void _fnmerge(char* path, const char* drive, const char* dir, const char* name, const char* ext);
+char* _searchpath(const char* file);
```
RadWolfie, line 15: NIT: Missing newline at end of file, otherwise git isn't praising you.

lib/xboxrt/libc_extensions/direct.c
JayFoxRox: Ohh stupid me. I thought it was for arrays which must be sized for each drive letter.
Thanks for clarifying 😊

GXTX: This doesn't take account of Chihiro partitions, with the largest being `mbupdate:\`.

ExoSkye: Honestly I'm half wondering if I should get rid of them altogether - i already got rid of some of the constants since i realised that digital mars was basically DOS documentation and it was saying that we'd have to abide by the 9-5 (filename + extension) rule, and I realised that's out of date
I personally haven't even used these in any of the implementation code and I don't see a reason I would need to, so I may just remove them

ExoSkye: On further inspection, `wine` doesn't define them whatsoever in it's source code so I'm pretty sure it's safe to remove them

lib/xboxrt/libc_extensions/direct.h
ExoSkye: Should be resolved by 0b37066

lib/xboxrt/libc_extensions/direct.c
ExoSkye: Should be resolved by 0b37066

lib/xboxrt/libc_extensions/direct.c
ExoSkye: Resolved by e2f40bf

lib/xboxrt/libc_extensions/direct.c
ExoSkye: This function doesn't exist in the `wine` project source code, so I'm going to remove it

ExoSkye: Resolved in 72ae765

lib/xboxrt/libc_extensions/direct.c
ExoSkye: I have a feeling that implementing relative path support (if there really is none) should be done in a separate PR, so I may just leave these stubbed for now

ExoSkye:
I'm going to set this as ready to review since I believe, other than certain stubbed functions, this is ready to go

ExoSkye:
I say this, and then completely forget to actually add the `.c` file to the makefile, whoops, i'll fix that rq

ExoSkye:
Can confirm that this built locally with no hiccups

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,85 @@
...
+#include <windows.h>
+
+// Made referencing https://www.digitalmars.com/rtl/direct.html, retrieved 2023-06-10, copyrighted 1999-2018 by Digital Mars
+
+
+/*
+ * TODO: Make a working directory system?
+ */
+
+int _chdir(char* path) {
```
Ryzee119, line 19: I'd suggest we either remove these stubs or atleast assert within them. We have `assert(0);` is other stubs.

Currently these will silently fail with a generic error code and the end user will have no idea why.

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,85 @@
...
+            errno = EACCES;
+        } else if (err == ERROR_PATH_NOT_FOUND) {
+            errno = ENOENT;
+        }
+
+        return -1;
+    }
+}
+
+int _rmdir(const char* pathname) {
```
Ryzee119, line 67: Need to handle ENOTEMPTY? i.e GetLastError == ERROR_DIRECTORY or ERROR_DIR_NOT_EMPTY

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,85 @@
...
+    BOOL result = RemoveDirectoryA(
+            pathname
+    );
+
+    if (result == true) {
+        return 0;
+    } else {
+        DWORD err = GetLastError();
+
+        if (err == ERROR_ALREADY_EXISTS) {
```
Ryzee119, line 77: Is `ERROR_ALREADY_EXISTS`  a valid error code for `RemoveDirectoryA`?

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,85 @@
...
+            pathname,
+            NULL
+    );
+
+    if (result == true) {
+        return 0;
+    } else {
+        DWORD err = GetLastError();
+
+        if (err == ERROR_ALREADY_EXISTS) {
```
Ryzee119, line 57: There's  alot of cases that are not covered from `GetLastError()`.  ERROR_DISK_FULL is just an obvious example.

Although these will fail, errno will not be set. Not sure if thats a problem

lib/xboxrt/libc_extensions/direct.c
ExoSkye: Good point, I'll do that

ExoSkye:
Ok, I've implemented a Win32 error code to errno function which just uses a switch statement (I felt that that would be easier for a compiler to optimise, there's even a chance that given there's no IO in it, it could optimise it away completely)

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -46,26 +46,28 @@ static int convert_error(DWORD winerror)
  */
 
 int _chdir(char* path) {
-    return 0;
+    assert(0);
+    return -1; // Unreachable
```
JayFoxRox, line 56: This *is* reachable when disabling assertions.
In that case, `errno` must be set, so I suggest setting it to `EACCES`?

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -46,26 +46,28 @@ static int convert_error(DWORD winerror)
...
-    return NULL;
+    assert(0);
+    return NULL; // Unreachable
 }
 
 char* _getwd(char* path_name) {
-    errno = -EINVAL;
-    return NULL;
+    assert(0);
+    return NULL; // Unreachable
```
JayFoxRox, line 71: The macOS man page says:

> The function getwd() is a compatibility routine which calls getcwd() with its buf argument and a size of MAXPATHLEN (as defined in the include file <sys/param.h>).  Obviously, buf should be at least MAXPATHLEN bytes in length.

Similar wording is in POSIX https://pubs.opengroup.org/onlinepubs/009696699/functions/getwd.html

which should be trivial to implement.

I can't find anything about it in the MSDN docs though. Is it contained in the Windows SDK / MS implementations?

lib/xboxrt/libc_extensions/direct.c
ExoSkye: Ah crap, forgot you could do that, I'll sort it out

lib/xboxrt/libc_extensions/direct.c
ExoSkye: Ah, yea, Wikipedia, and digitalmars state that getwd exists so I'm going on that
Should be trivial to implement though

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -46,26 +46,28 @@ static int convert_error(DWORD winerror)
...
-    return 0;
+    assert(0);
+    return -1; // Unreachable
 }
 
 char* _getcwd(char* buffer, size_t length) {
-    errno = -EINVAL;
-    return NULL;
+    assert(0);
+    return NULL; // Unreachable
```
JayFoxRox, line 66: I believe we can hard-code this for now, so at least apps which use this to find files will work (sometimes some SDK or API will require absolute paths, and then the application might query the base path through this).

I can see how that would be a lot of additional implementation effort, so it can be done in a follow-up PR.

---

If this follows the same convention, we might be able to use: https://github.com/XboxDev/nxdk/blob/500354ef264f35a46a31643876b06ed614b0aed8/lib/nxdk/path.c#L9-L14

Otherwise, we might be able to hardcode it to `D:\\` which is either (I don't remember) a kernel requirement or at least a dashboard- or self-imposed XBE controlled symlink?

Eitherway, we'd have to follow conventions of the MS implementation (regarding forward vs backward slash, device names vs driver letters etc.).

Similar reasoning for `_getdrive`, although I speculate that this is required much less.

lib/xboxrt/libc_extensions/direct.c
JayFoxRox: > I can't find anything about it in the MSDN docs though. Is it contained in the Windows SDK / MS implementations?

I don't see it in https://github.com/microsoft/win32metadata/blob/main/generation/WinSDK/RecompiledIdlHeaders/ucrt/direct.h either.

RadWolfie: Seems like getwd is a legacy function which was discontinued. The question is when did it get removed? As for digitalmars state it exists, I don't see getwd mentioned in the referenced link from the OP post. Heck, even did a search for whole site which also has no mention too.

lib/xboxrt/libc_extensions/direct.c
RadWolfie: D drive letter mount is handled by XDK on boot when D drive letter isn't set by the kernel. Otherwise, it is handled by the kernel on reboot.

Though, I highly recommended to check `LaunchDataPage->Header.szLaunchPath` first for semicolon character which is mount as working directory by the kernel itself. Then fallback to XeImageFileName with null terminate append at end of string.

fyi: I provided this type of information in XboxDev discord server based on what I had researched years ago.

JayFoxRox: I'd prefer `XeImageFileName` (which the kernel uses to talk to the app / after launching) over `LaunchDataPage` (which is used by the app to talk to the kernel / before launching)

RadWolfie: Yes, LaunchDataPage is used to talk to the kernel but that's only half correct. It is also used by the app as well hence the LaunchData member variable from LAUNCH_DATA_PAGE structure is used to tell the app, either from same xbe or another xbe, what to do next after reboot occur. The kernel doesn't modify LaunchDataPage created by the app. The kernel read from `LaunchDataPage->Header` structure to do what the app want the kernel to do on reboot.

Here's a log of how it works on first boot and reboot:
```
---- first boot ----
LaunchDataPage is NULL.
XeImageFileName: \Device\Harddisk0\Partition1\Games\reboot\default.xbe
\??\D: == \Device\Harddisk0\Partition1\Games\reboot
Relaunching... LaunchDataPage->Header.szLaunchPath: \Device\Harddisk0\Partition1\Games;reboot\default.xbe
============
------ reboot ------
LaunchDataPage->Header.szLaunchPath: \Device\Harddisk0\Partition1\Games;reboot\default.xbe
XeImageFileName: \Device\Harddisk0\Partition1\Games\reboot\default.xbe
\??\D: == \Device\Harddisk0\Partition1\Games
============
```

As you can see XeImageFileName doesn't include semicolon which break working directory mount from D: symlink pointing to a parent directory than the same directory according to reboot process.
This is the same/similar concept on Windows' command prompt or powershell usage. When you change directory from the terminal itself, i.e. `C:\logs>` . Then execute `D:\build\UnitTest.exe` which will create log file into `C:\logs` instead of `D:\build` directory.

In summary, this is how demo discs and even xbox live works when they focus on D:\ symlink as a working directory. Otherwise attempting to access the contents wouldn't work engineered by original xbox team. I hope this is more clarified for you and everyone else reading this.

thrimbor:
Merged as part of https://github.com/XboxDev/nxdk/pull/751

PR #640, opened 2023-06-24
winapi: Add Event functions

Ryzee119:
I need asynchronous File IO using overlapped functions which is [not currently supported](https://github.com/XboxDev/nxdk/blob/9daebb58281125c6e0b7be8ebd36e93be06db63f/lib/winapi/fileio.c#L107). Win32 uses  [Event Objects](https://learn.microsoft.com/en-us/windows/win32/sync/event-objects) to achieve this.

This is the first step to support overlapped IO.

Usage example below that simply has a sleeping thread wait for an event. Also tests some error scenarios.

```c
#include <windows.h>
#include <stdio.h>
#include <assert.h>
#include <hal/video.h>
#include <hal/debug.h>

static DWORD WINAPI EventThread(LPVOID lpParam)
{
    HANDLE hEvent = (HANDLE)lpParam;
    while (1) {
        if (WaitForSingleObject(hEvent, INFINITE) == WAIT_OBJECT_0) {
            debugPrint("Thread received the event!\n");
        }
    }
    return 0;
}

int main()
{
    HANDLE hEvent1, hEvent2, hMutex;

    // Create a named event. Then try create another event with the same name,
    // Also create a different object type with the same name and confirm errors match up
    hEvent1 = CreateEvent(NULL, FALSE, FALSE, "hello world");
    hEvent2 = CreateEvent(NULL, FALSE, FALSE, "hello world");
    assert(GetLastError() == ERROR_ALREADY_EXISTS);
    hMutex = CreateMutex(NULL, FALSE, "hello world");
    assert(GetLastError() == ERROR_INVALID_HANDLE);

    HANDLE hThread = CreateThread(NULL, 0, EventThread, hEvent1, 0, NULL);

    debugPrint("Setting the event in 2 seconds\n\n");
    Sleep(2000);
    assert(SetEvent(hEvent1) == TRUE);
    Sleep(500); //Manual Reset is FALSE, so make sure the thread isnt spamming while we sleep here.
    assert(ResetEvent(hEvent1) == TRUE);

    debugPrint("Pulsing the event 3 times in 2 seconds...\n\n");
    Sleep(2000);
    for (INT i = 0; i < 3; i++) {
        assert(PulseEvent(hEvent1) == TRUE);
        Sleep(100);
    }
    Sleep(100);
    debugPrint("Test Complete!\n");
    while(1) Sleep(1000);
}
```
 Ouputs below. I ran the same code natively on my windows PC and the output is the same.
xemu-2023-06-24-13-14-23

### DIFF
--- a/lib/winapi/sync.c
+++ b/lib/winapi/sync.c
@@ -580,3 +580,80 @@ BOOL ReleaseMutex (HANDLE hMutex)
     SetLastError(RtlNtStatusToDosError(status));
     return FALSE;
 }
+
+HANDLE CreateEventA (LPSECURITY_ATTRIBUTES lpEventAttributes, BOOL bManualReset, BOOL bInitialState, LPCSTR lpName)
+{
+    NTSTATUS status;
+    HANDLE handle;
+    ANSI_STRING obj_name;
+    OBJECT_ATTRIBUTES obj_attributes;
+    POBJECT_ATTRIBUTES obj_attributes_ptr;
+    EVENT_TYPE event;
+
+    if (lpName) {
+        RtlInitAnsiString(&obj_name, lpName);
+        InitializeObjectAttributes(&obj_attributes, &obj_name, OBJ_OPENIF, ObWin32NamedObjectsDirectory(), NULL);
+        obj_attributes_ptr = &obj_attributes;
+    } else {
+        obj_attributes_ptr = NULL;
+    }
+
+    if (bManualReset) {
+        event = NotificationEvent;
+    } else {
+        event = SynchronizationEvent;
+    }
+
+    status = NtCreateEvent(&handle, obj_attributes_ptr, event, bInitialState);
+    if (!NT_SUCCESS(status)) {
+        SetLastError(RtlNtStatusToDosError(status));
+        return NULL;
+    }
+
+    if (status == STATUS_OBJECT_NAME_EXISTS) {
+        SetLastError(ERROR_ALREADY_EXISTS);
+    } else {
+        SetLastError(0);
+    }
+
+    return handle;
+}
+
+BOOL SetEvent (HANDLE hEvent)
+{
+    NTSTATUS status;
+
+    status = NtSetEvent(hEvent, NULL);
+    if (NT_SUCCESS(status)) {
+        return TRUE;
+    }
+
+    SetLastError(RtlNtStatusToDosError(status));
+    return FALSE;
+}
+
+BOOL ResetEvent (HANDLE hEvent)
+{
+    NTSTATUS status;
+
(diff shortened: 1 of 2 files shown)


lib/winapi/sync.c
```diff
@@ -580,3 +580,80 @@ BOOL ReleaseMutex (HANDLE hMutex)
...
+    }
+
+    return handle;
+}
+
+BOOL SetEvent (HANDLE hEvent)
+{
+    NTSTATUS status;
+
+    status = NtSetEvent(hEvent,NULL);
```
RadWolfie, line 626: Question: Do need to keep variable declaration and the setter couple lines down separate? Seems like we can migrate into one line. (the same for below)

If maintainers say they shouldn't, then dismiss this question.

RadWolfie (changes requested):

RadWolfie:
I have verified these APIs from this pull request are identical to according decompiled titles with references of NTxxxxEvent usage:
- [X] CreateEventA
- [X] SetEvent
- [X] ResetEvent
- [X] PulseEvent

So, the only thing left are review remark(s).

lib/winapi/sync.c
RadWolfie: Hmm... I recently notice `ReleaseMutex` and `ReleaseSemaphore` functions does the same thing little bit up higher. Please dismiss this conversation.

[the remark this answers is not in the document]
JayFoxRox: We should start using CI for checking formatting / style in some folders; while there's no consistent style, there's typically a common style in subfolders (at least more recent ones like winapi or nxdk)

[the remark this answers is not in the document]
Ryzee119: Yes. I spent alot of time trying to get the spaces and commas right but still missed a spot. Tbh took longer formatting than coding 🙃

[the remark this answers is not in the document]
Ryzee119: Fixed this

RadWolfie (approved):

lib/winapi/sync.c
```diff
@@ -580,3 +580,80 @@ BOOL ReleaseMutex (HANDLE hMutex)
     SetLastError(RtlNtStatusToDosError(status));
     return FALSE;
 }
+
```
JayFoxRox, line 583: nit: Would expect commit title to be "winapi: Add Event functions", following scheme in https://github.com/XboxDev/nxdk/commit/69fbf00c959368989de458e4183dbddbd4b01b2d (note: "Add" vs "Implement" and "functions" vs "Functions")

Ryzee119: Ive changed the commit title

PR #641, opened 2023-06-25
cxbe: Fix sprintf compiler warnings

abaire:
Switches use of sprintf to snprintf to prevent a bunch of clang deprecation warnings.

### DIFF
--- a/tools/cxbe/Xbe.cpp
+++ b/tools/cxbe/Xbe.cpp
@@ -115,7 +115,11 @@ Xbe::Xbe(const char *x_szFilename)
 
             if(fread(&m_SectionHeader[v], sizeof(*m_SectionHeader), 1, XbeFile) != 1)
             {
-                sprintf(szBuffer, "Unexpected end of file while reading Xbe Section Header %d (%Xh)", v, v);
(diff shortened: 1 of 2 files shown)


tools/cxbe/Xbe.cpp
```diff
@@ -992,7 +1009,12 @@ void Xbe::Export(const char *x_szXbeFilename)
 
             if(fwrite(m_bzSection[v], RawSize, 1, XbeFile) != 1)
             {
-                sprintf(szBuffer, "Unexpected write error while writing Xbe Section %d (%Xh) (%s)", v, v, m_szSectionName[v]);
+                snprintf(szBuffer,
+                         sizeof(szBuffer),
+                         "Unexpected write error while writing Xbe Section %d (%Xh) (%s)",
+                         v,
+                         v,
+                         m_szSectionName[v]);
```
JayFoxRox, line 1017: nit: All of these edits are for `SetError` which internally has a maximum length anyway.
I think this change is good, but we should probably fix the underlying issue by having `SetError(bool x_bFatal, const char *x_szErrorFmt, ...)` which simplifies the code further.

PR #642, opened 2023-06-25
cxbe: Support non-kernel imports

abaire:
This implements handling for `dwNonKernelImportDirAddr` in DEBUG mode, allowing nxdk-based XBEs to make use of `xbdm.dll` and other non-kernel imports.

### DIFF
--- a/tools/cxbe/Xbe.cpp
+++ b/tools/cxbe/Xbe.cpp
@@ -16,6 +16,10 @@
 #include <cstring>
 #include <cstdio>
 
+
+static const char kKernelImageName[] = "xboxkrnl.exe";
+static uint32 CountNonKernelImportTableEntries(class Exe *x_Exe, uint32_t *extra_bytes);
+
 // construct via Xbe file
 Xbe::Xbe(const char *x_szFilename)
 {
@@ -346,7 +350,6 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
         // build time/date
         m_Header.dwTimeDate = CurrentTime;
 
-        // TODO: generate valid addr if necessary
         m_Header.dwNonKernelImportDirAddr = 0;
 
         // Need one or more library version, otherwise VerifierX on XDKs crashes
@@ -362,6 +365,7 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
     printf("Xbe::Xbe: Pass 2 (Calculating Requirements)...");
 
     // pass 2
+    uint32 non_kernel_import_table_bytes = 0;
     {
         // make-room cursor
         uint32 mrc = m_Header.dwBaseAddr + sizeof(m_Header);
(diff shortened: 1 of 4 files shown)


tools/cxbe/Xbe.cpp
```diff
@@ -1527,3 +1524,91 @@ uint08 *Xbe::GetLogoBitmap(uint32 x_dwSize)
 
     return 0;
 }
+
+// returns the count of import table entries other than xboxkrnl.exe.
```
JayFoxRox, line 1528: nit: avoid "xboxkrnl.exe" here, and instead refer just to the "kernel", so this remains valid even if `kKernelImageName` is modified

tools/cxbe/Xbe.cpp
```diff
@@ -1527,3 +1524,91 @@ uint08 *Xbe::GetLogoBitmap(uint32 x_dwSize)
...
+                *wideCharName++ = name[ch];
+                *wideCharName++ = 0;
+                nextImportNameAddr += 2;
+            }
+            ++xbeDescriptor;
+        }
+    }
+
+    if (kernel_thunk_addr == 0) {
+        printf("Xbe::ProcessImportTable: No imports from xboxkrnl.exe, exe may be invalid.\n");
```
JayFoxRox, line 1599: This should use `%s` and `kKernelImageName` so this doesn't break when modifying `kKernelImageName`.

tools/cxbe/Xbe.cpp
```diff
@@ -1527,3 +1524,91 @@ uint08 *Xbe::GetLogoBitmap(uint32 x_dwSize)
...
+    }
+
+    kernel_thunk_addr ^= (x_bRetail ? XOR_KT_RETAIL : XOR_KT_DEBUG);
+
+    m_Header.dwKernelImageThunkAddr = kernel_thunk_addr;
+
+    if (xbeDescriptor) {
+        if (x_bRetail) {
+            printf("Xbe::ProcessImportTable: Non-kernel imports found in retail image, exe may be invalid.\n");
```
JayFoxRox, line 1608: nit: The other message above doesn't use generic naming ("Non-kernel"), but explicit naming ("xboxkrnl.exe"). I think this should be more consistent.

JayFoxRox:
Interesting, I didn't realize we were missing this functionality, because I've already used xbdm before... from a dxt: https://github.com/JayFoxRox/nxdk/pull/10

tools/cxbe/Xbe.cpp
abaire: While I agree in principle, I think for `xboxkrnl.exe` to be safely modified it'd either need to be passed as a parameter to cxbe or there'd have to be a non-backwards compatible change accompanied by a change to the import lib in the nxdk: https://github.com/XboxDev/nxdk/blob/master/lib/xboxkrnl/xboxkrnl.exe.def.

I've gone ahead and made the language agnostic, but don't personally see a need for actually making it configurable so I'm leaving it there.

tools/cxbe/Xbe.cpp
abaire: Done.

tools/cxbe/Xbe.cpp
abaire: Modified the above message slightly to include the generic name in case somebody really does change the kernel name to something opaque. In this case the messaging reflects the header naming/behavior which differentiates solely between "kernel" and "everything else." While we could loop through the names again here and print out everything detected, presumably the vast majority of folks technical enough to add a non-kernel dep would also know how to look at the import table (or their linker call) to understand what they've included. 

My expectation is that this warning will primarily show up when somebody forgets to set `-MODE:DEBUG`. I'm also on the fence about whether it should be a warning or a fatal error. I don't think retail boxes look at the non-kernel import table at all, so it's likely any binary attempting to use those imports will just crash (which is what eventually led me down the path to adding support).

abaire:
> Interesting, I didn't realize we were missing this functionality, because I've already used xbdm before... from a dxt: [JayFoxRox#10](https://github.com/JayFoxRox/nxdk/pull/10)

Same here, it took me a day of pondering to even suspect the exe -> xbe conversion might be at fault. For my dxt-like loader I'd been using xbdm to resolve the imports before injecting the image.

I wonder if the DXT path just isn't the same as the XBE and it reads the normal PE import table? Given that the XBE import descriptor is dramatically different than for XBE's including the switch from `char` -> `char16_t` it seems like it'd be impossible for it to work any other way.

abaire:
Note: The macOS CI failure doesn't seem to be related (fails before even making it to the build)

PR #643, opened 2023-07-05
cxbe: Populate debug path header fields

abaire:


tools/cxbe/Exe.cpp
```diff
@@ -158,6 +158,8 @@ Exe::Exe(const char *x_szFilename)
         }
     }
 
+    m_debugPath = x_szFilename;
```
JayFoxRox, line 161: I think I'd prefer if this was an optional thing or user-controlled.

I'm concerned with leaking user-data (such as usernames) or storing unexpected data (which XBE parsers can't handle - such as non Windows paths).

abaire: Added `DEBUGPATH` option.

thrimbor:
I've reviewed (and tried) the code and everything seems to be fine. I'd like to see the fix in https://github.com/XboxDev/nxdk/pull/652 land first but once it's in I'll hit the merge button on this one.

PR #644, opened 2023-07-05
cxbe: Add tool to dump XBE headers

abaire:


tools/cxbe/Xbe.cpp
```diff
@@ -184,6 +239,10 @@ Xbe::Xbe(const char *x_szFilename)
                 goto cleanup;
             }
 
+            if (!strcmp("CXBE0", m_LibraryVersion[v].szName))
+            {
+                is_cxbe_generated = true;
+            }
```
JayFoxRox, line 245: I don't think we should track wether this was generated by `CXBE`. This library version header was also added rather late in CXBE development.

tools/cxbe/Xbe.cpp
```diff
@@ -193,20 +252,24 @@ Xbe::Xbe(const char *x_szFilename)
 
             if(m_Header.dwKernelLibraryVersionAddr == 0)
             {
-                SetError("Could not locate kernel library version", true);
-                goto cleanup;
+                if (!is_cxbe_generated)
+                {
+                    SetError("Could not locate kernel library version", true);
+                    goto cleanup;
+                }
```
JayFoxRox, line 259: I think this should warn unconditionally, but continue regardless.

tools/cxbe/.gitignore
```diff
@@ -1 +1,2 @@
 cxbe
+dumpxbe
```
JayFoxRox, line 2: nit: CXBE is kept rather minimalistic for nxdk (intentionally maybe?) to handle EXE > XBE conversion.

FYI, there's also https://github.com/XboxDev/xbedump which might be more capable.

abaire: Neat, I didn't know that existed.

Is there any reason to keep the Xbe read constructor code in cxbe at all since it's otherwise not used and xbedump looks like it provides the same functionality?

mborgerson: > FYI, there's also https://github.com/XboxDev/xbedump which might be more capable.

Agree here. We have multiple tools to dump XBE info already. Let's avoid another

> Is there any reason to keep the Xbe read constructor code in cxbe at all since it's otherwise not used and xbedump looks like it provides the same functionality?

Probably not. This is a remnant of original XBE&rarr;EXE conversion.

abaire: Cool, will close this and send a PR to delete the dead code.

PR #645, opened 2023-07-05
cxbe: Populate XBOXKRNL and XAPILIB versions

abaire:


tools/cxbe/Xbe.cpp
```diff
@@ -747,6 +748,32 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
                 szBuffer += sizeof(*m_LibraryVersion);
                 hwc += sizeof(*m_LibraryVersion);
             }
+
+            m_KernelLibraryVersion = new LibraryVersion;
```
mborgerson, line 752: * `XBOXKRNL` and `XAPILIB` library version decls should be counted in `m_Header.dwLibraryVersions`
* Should eliminate the dummy CXBEx lib version if we are populating these ones
* Might be simpler to just increase the count, let the above code do the allocation, then fill in the info after

JayFoxRox:
> Populate XBOXKRNL and XAPILIB versions 

Why?

---

To my knowledge:
- `XBOXKRNL` is not part of the standard libraries (which is meant for non-kernel libraries).
- `XAPILIB` is not actually linked by nxdk, so I'm not sure if / why we should add it here.

The CXBE lib is just a stub, to avoid errors on XDKs (some default MS dxt plugin parses the header, and crashes if there's no valid pointer).

tools/cxbe/Xbe.cpp
JayFoxRox: > `XBOXKRNL` and `XAPILIB` library version decls should be counted in m_Header.dwLibraryVersions
> [...]
> Might be simpler to just increase the count, let the above code do the allocation, then fill in the info after

Did you check on MS XBEs? They differentiate between kernel and non-kernel libs (I didn't remember they also track XAPI separately), so I'm not sure if these 2 count.

> Should eliminate the dummy CXBEx lib version if we are populating these ones

I believe VerifierX still needs at least one non-kernel lib?
XDK always has at least one valid lib, because of D3D etc.

mborgerson: > Did you check on MS XBEs

Yes, here's Halo. These version pointers in the header simply point into the array of all library versions. Maybe there is an exception in some other XBEs, but this is how I've always seen it.

```python
In [3]: xbe = Xbe.from_file("default.xbe")

In [4]: xbe.header.lib_versions_count
Out[4]: 8

In [5]: xbe.libraries
Out[5]: 
{'XAPILIB': <XbeLibrary 'XAPILIB' (1.0.3911)>,
 'D3D8': <XbeLibrary 'D3D8' (1.0.3925)>,
 'D3DX8': <XbeLibrary 'D3DX8' (1.0.3911)>,
 'DSOUND': <XbeLibrary 'DSOUND' (1.0.3936)>,
 'DSOUNDH': <XbeLibrary 'DSOUNDH' (1.0.3937)>,
 'XNETS': <XbeLibrary 'XNETS' (1.0.3911)>,
 'XBOXKRNL': <XbeLibrary 'XBOXKRNL' (1.0.3911)>,
 'LIBC': <XbeLibrary 'LIBC' (1.0.3911)>}
```

abaire:
> _(Edit: I see now that these are standard fields in the header? Still not sure if we should stub them like this)_

Yeah, the fact that they're explicitly in the header is what prompted me to add them. I was running into a crash when loading a binary linked with xbdm dependencies and noticed that it wasn't printing the usual "This XBE was linked with ..." messages. I glanced at the header and erroneously attributed the crash to these versions not being populated.

I think it's probably fine to leave them unpopulated but don't know if there's any devkit tooling that actually expects valid values.

I'm happy to either drop this PR or do the fixup @mborgerson suggested, just let me know what's preferred.

mborgerson:
> I'm happy to either drop this PR or do the fixup @mborgerson suggested, just let me know what's preferred.

If it's not required for debug enabling, let's keep it simple and stash the change for now

abaire:
Closing as the lack of these values doesn't seem to cause any concrete issue.

PR #648, opened 2023-07-07
cxbe: Add clang-format

abaire:
The Microsoft style is fairly close to the prevailing cxbe style. 
I set the line length to 100 chars as that covers most of the long cases.
Other tweaks are made in an attempt to minimize delta.


mborgerson:
Spending fewer cycles formatting code is a huge win, and I appreciate the effort in delta minimization. Thanks!

PR #651, opened 2023-07-20
Added various features

PQCraft:
`tools/cxbe`:
- Added the ability to handle long section names (moved the limit from 8 chars to 255)
- Added an option to set the XBE's title ID to either a hex code (like `FFFF0002`) or a human-readable code (like `CX-9999`)
- Added an option to set the XBE's region flags
- Added an option to set the XBE's version (can be dec, hex, or oct)

`Makefile`:
- Added `XBE_TITLEID` to define a title ID (goes to cxbe as `-TITLEID:$(XBE_TITLEID)`)
- Added `XBE_REGION` to define a title ID (goes to cxbe as `-REGION:$(XBE_REGION)`)
- Added `XBE_VERSION` to define a title ID (goes to cxbe as `-VERSION:$(XBE_VERSION)`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,6 @@ else
 main.exe: $(OBJS) $(NXDK_DIR)/lib/xboxkrnl/libxboxkrnl.lib
 	@echo "[ LD       ] $@"
 	$(VE) $(LD) $(NXDK_LDFLAGS) $(LDFLAGS) -out:'$@' $^
-	$(VE) objcopy --rename-section 'XTIMAGE=$$$$XTIMAGE' --rename-section 'XSIMAGE=$$$$XSIMAGE' $@ || exit 0
 endif
 
 %.lib:
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,6 @@ endif
 ifeq ($(XBE_TITLE),)
 XBE_TITLE = nxdk_app
 endif
-ifeq ($(XBE_TITLEID),)
-XBE_TITLEID = FFFF0002
-endif
 
 ifeq ($(OUTPUT_DIR),)
 OUTPUT_DIR = bin
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ endif
 ifeq ($(XBE_TITLE),)
 XBE_TITLE = nxdk_app
 endif
+ifeq ($(XBE_TITLEID),)
+XBE_TITLEID = FFFF0002
+endif
 
 ifeq ($(OUTPUT_DIR),)
 OUTPUT_DIR = bin
@@ -98,7 +101,8 @@ DEPS += $(filter %.cpp.d, $(SRCS:.cpp=.cpp.d))
 
 $(OUTPUT_DIR)/default.xbe: main.exe $(OUTPUT_DIR) $(CXBE)
 	@echo "[ CXBE     ] $@"
-	$(VE)$(CXBE) -OUT:$@ -TITLE:$(XBE_TITLE) $< $(QUIET)
+	$(VE)$(CXBE) -OUT:$@ -TITLE:$(XBE_TITLE) -TITLEID:$(XBE_TITLEID) \
+	    -REGION:$(XBE_REGION) -VERSION:$(XBE_VERSION) $< $(QUIET)
 
 $(OUTPUT_DIR):
 	@mkdir -p $(OUTPUT_DIR);
@@ -118,6 +122,7 @@ else
 main.exe: $(OBJS) $(NXDK_DIR)/lib/xboxkrnl/libxboxkrnl.lib
 	@echo "[ LD       ] $@"
 	$(VE) $(LD) $(NXDK_LDFLAGS) $(LDFLAGS) -out:'$@' $^
+	$(VE) objcopy --rename-section 'XTIMAGE=$$$$XTIMAGE' --rename-section 'XSIMAGE=$$$$XSIMAGE' $@ || exit 0
 endif
 
 %.lib:
--- a/tools/cxbe/Exe.cpp
+++ b/tools/cxbe/Exe.cpp
@@ -122,8 +122,10 @@ Exe::Exe(const char *x_szFilename)
                 for(uint32 i = 1; i < 8; ++i)
                 {
                     char c = m_SectionHeader[v].m_name[i];
-                    if(c < '0' || c > '9')
+                    if(!c)
                         break;
+                    if(c < '0' || c > '9') // not a long section after all?
+                        goto notlong;
                     m_SectionHeader_longname[v].m_offset *= 10;
                     m_SectionHeader_longname[v].m_offset += c - '0';
                 }
@@ -133,14 +135,13 @@ Exe::Exe(const char *x_szFilename)
                 fseek(ExeFile, m_Header.m_symbol_table_addr + m_SectionHeader_longname[v].m_offset,
                       SEEK_SET);
 
-                uint32 i = 0;
-                while(i < 255)
+                uint32 i;
+                for(i = 0; i < 255; ++i)
                 {
                     int c = fgetc(ExeFile);
                     if(!c || c == EOF)
                         break;
                     m_SectionHeader_longname[v].m_longname[i] = c;
-                    ++i;
                 }
                 m_SectionHeader_longname[v].m_longname[i] = 0;
 
@@ -150,6 +151,7 @@ Exe::Exe(const char *x_szFilename)
             }
             else
             {
+            notlong:;
                 m_SectionHeader_longname[v].m_longname = NULL;
             }
 
--- a/tools/cxbe/Exe.cpp
+++ b/tools/cxbe/Exe.cpp
@@ -146,16 +146,15 @@ Exe::Exe(const char *x_szFilename)
                 m_SectionHeader_longname[v].m_longname[i] = 0;
 
                 fseek(ExeFile, tmppos, SEEK_SET);
-                printf(" (long: offset=%u name=%s) ", m_SectionHeader_longname[v].m_offset,
+                printf("OK %d (long)\n", v, m_SectionHeader_longname[v].m_offset,
                        m_SectionHeader_longname[v].m_longname);
             }
             else
             {
             notlong:;
                 m_SectionHeader_longname[v].m_longname = NULL;
+                printf("OK %d\n", v);
             }
-
-            printf("OK %d\n", v);
         }
     }
 
--- a/tools/cxbe/Exe.cpp
+++ b/tools/cxbe/Exe.cpp
@@ -99,6 +99,7 @@ Exe::Exe(const char *x_szFilename)
     // read section headers
     {
         m_SectionHeader = new SectionHeader[m_Header.m_sections];
+        m_SectionHeader_longname = new SectionHeader_longname[m_Header.m_sections];
 
         printf("Exe::Exe: Reading Section Headers...\n");
 
@@ -112,6 +113,44 @@ Exe::Exe(const char *x_szFilename)
                 snprintf(buffer, sizeof(buffer), "Could not read PE section header %d (%Xh)", v, v);
                 SetError(buffer, true);
                 goto cleanup;
+            }
+
+            // interpret long section names
+            if(m_SectionHeader[v].m_name[0] == '/')
+            {
+                m_SectionHeader_longname[v].m_offset = 0;
+                for(uint32 i = 1; i < 8; ++i)
+                {
+                    char c = m_SectionHeader[v].m_name[i];
+                    if(c < '0' || c > '9')
+                        break;
+                    m_SectionHeader_longname[v].m_offset *= 10;
+                    m_SectionHeader_longname[v].m_offset += c - '0';
+                }
+                m_SectionHeader_longname[v].m_longname = new char[256]();
+
+                long tmppos = ftell(ExeFile);
+                fseek(ExeFile, m_Header.m_symbol_table_addr + m_SectionHeader_longname[v].m_offset,
+                      SEEK_SET);
+
+                uint32 i = 0;
+                while(i < 255)
+                {
+                    int c = fgetc(ExeFile);
+                    if(!c || c == EOF)
+                        break;
+                    m_SectionHeader_longname[v].m_longname[i] = c;
+                    ++i;
+                }
+                m_SectionHeader_longname[v].m_longname[i] = 0;
+
+                fseek(ExeFile, tmppos, SEEK_SET);
+                printf(" (long: offset=%u name=%s) ", m_SectionHeader_longname[v].m_offset,
+                       m_SectionHeader_longname[v].m_longname);
+            }
+            else
+            {
+                m_SectionHeader_longname[v].m_longname = NULL;
             }
 
             printf("OK %d\n", v);
@@ -181,6 +220,7 @@ cleanup:
 void Exe::ConstructorInit()
 {
     m_SectionHeader = NULL;
+    m_SectionHeader_longname = NULL;
     m_bzSection = NULL;
 }
 
@@ -190,12 +230,16 @@ Exe::~Exe()
     if(m_bzSection != 0)
     {
         for(uint32 v = 0; v < m_Header.m_sections; v++)
+        {
+            delete[] m_SectionHeader_longname[v].m_longname;
             delete[] m_bzSection[v];
+        }
 
         delete[] m_bzSection;
     }
 
     delete[] m_SectionHeader;
+    delete[] m_SectionHeader_longname;
 }
 
 // export to Exe file
--- a/tools/cxbe/Main.cpp
+++ b/tools/cxbe/Main.cpp
@@ -27,7 +27,7 @@ int main(int argc, char *argv[])
     char szLogo[OPTION_LEN + 1] = "";
 
     bool bRetail;
-    uint32 dwTitleId = 0x4358270F; // CX-9999
+    uint32 dwTitleId = 0xFFFF0002;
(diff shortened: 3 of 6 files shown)


thrimbor:
I haven't given this an in-depth review yet, but here's some preliminary feedback:



> `tools/cxbe`:

This commit contains multiple separate changes and should be split up, with each commit introducing one feature and a short but descriptive commit title following the style of previous commits.

> `lib/pbkit`:

The printing functionality in pbkit is a bit of a legacy feature - pbkit is currently a wild mix of high- and low-level code, and the printing code is most likely going to be removed soon-ish, so imho it doesn't make much sense to add features to it. It's only still here because it can help with debugging.
If you need a way to print text, I'd suggest having your own printing code in your app (potentially using the current pbkit code as a starting point).

> `Makefile`:

The makefile is a remnant from the early days of nxdk, I'd actually recommend against using it for anything other than building nxdk itself. We've been working on removing the need for it by adding wrapper scripts for the compiler and CMake support, and we'll eventually remove it.
For the moment, if you rely on the features you added, I'd recommend adding them as overrides to your own makefile and including the nxdk makefile like the samples do, but the most future proof solution is supplying your own independent makefile.

>     * Added an `objcopy` line after the linking stage to redirect `XTIMAGE` to `$$XTIMAGE` and `XSIMAGE` to `$$XSIMAGE` (it is impossible to define these sections otherwise)

Hmmm, interesting. I'll try to find some time to experiment with this, but if there really is no other way we should probably put that info into the nxdk wiki.

This review might sound a bit negative, but be assured that I really do appreciate people upstreaming fixes and features 👍

PQCraft:
Splitting up the commits for cxbe might take a bit since I will need to go into the code and manually undo the changes in reverse order then redo them

PQCraft:
For pbkit, I'll probably steal pbkit's text functions directly and paste them into my rendering code lol

For the Makefile, I am relying on calling it from my actual Makefile using this little snippet:
```
    ...
    else ifeq ($(CROSS),xbox)
        ifndef NXDK_DIR
            .PHONY: error
            error:
	            @echo Please define the NXDK_DIR environment variable
	            @exit 1
        endif
        ifneq ($(MODULE),engine)
            .PHONY: error
            error:
	            @echo Invalid module: $(MODULE)
	            @exit 1
        endif

        PLATFORM := Xbox

        export XBE_TITLE := PlatinumSrc
        export XBE_TITLEID := PQ-001
        export XBE_VERSION := $(shell grep '#define PSRC_BUILD ' src/version.h | sed 's/#define .* //')
        export GEN_XISO := $(XBE_TITLE).xiso.iso
        export NXDK_SDL := y
        SRCS := $(wildcard $(SRCDIR)/psrc_aux/*.c)
        SRCS := $(SRCS) $(wildcard $(SRCDIR)/psrc_engine/*.c)
        SRCS := $(SRCS) $(wildcard $(SRCDIR)/psrc_enginemain/*.c)
        SRCS := $(SRCS) $(wildcard $(SRCDIR)/psrc_server/*.c)
        SRCS := $(SRCS) $(wildcard $(SRCDIR)/stb/*.c)
        export SRCS
        export NXDK_SDL = y

        MKENV := ${MKENV} OUTPUT_DIR=xbox
        CPPFLAGS := $(CPPFLAGS) -DSTBI_NO_SIMD -DPB_HAL_FONT
        ifdef DEBUG
            CFLAGS := $(CFLAGS) -Og -g
            CPPFLAGS := $(CPPFLAGS) -DDBGLVL=$(DEBUG)
            MKENV := ${MKENV} DEBUG=y
        else
            ifndef O
                O = 2
            endif
            CFLAGS := $(CFLAGS) -O$(O)
            ifndef NOLTO
                MKENV := ${MKENV} LTO=y
            endif
        endif

        export CFLAGS := $(CFLAGS) $(CPPFLAGS)
        export LDFLAGS

        xbox-build:
	        @$(MAKE) --no-print-directory -f $(NXDK_DIR)/Makefile ${MKENV}

        xbox-clean:
	        @$(MAKE) --no-print-directory -f $(NXDK_DIR)/Makefile ${MKENV} clean

        xbox-distclean:
	        @$(MAKE) --no-print-directory -f $(NXDK_DIR)/Makefile ${MKENV} distclean

        .PHONY: xbox-build xbox-clean xbox-distclean
    else
    ...
```

PQCraft:
For the `$$XTIMAGE` and `$$XSIMAGE` sections, putting a `$` in the section name causes the linker to remove everything after the `$` and the `$` itself (`my$sect` -> `my`). I think it's for some special Windows stuff with shared libraries but the linker does it regardless of the output. Also the linker will truncate any section longer than 8 chars and these sections are both 9 chars. It isn't possible to tell the linker to stop modifying section names or to allow long section names so you have to edit the binary after linking using an external tool (like `objcopy`).

PQCraft:
Ok @thrimbor I split up the cxbe commits and also removed the changes to pbkit :+1:

tools/cxbe/Xbe.h
```diff
@@ -171,8 +171,10 @@ class Xbe : public Error
         uint32 dwCharacteristics; // characteristics
     } __attribute((packed)) * m_TLS;
 
-    // Xbe section names, each 8 bytes max and null terminated
-    char (*m_szSectionName)[9];
+    // Xbe section names, each 255 bytes max and null terminated
+    // Old value was 8 which caused problems
+    // For example, $$XTIMAGE and $$XSIMAGE are 9 bytes long not including null
```
JayFoxRox, line 176: Old value does not matter; if people care about it, they can check the blame.
Such comments are best left in the PR review, but not in the code.

tools/cxbe/Exe.cpp
```diff
@@ -114,6 +115,44 @@ Exe::Exe(const char *x_szFilename)
...
 
+            // interpret long section names
+            if(m_SectionHeader[v].m_name[0] == '/')
+            {
+                m_SectionHeader_longname[v].m_offset = 0;
+                for(uint32 i = 1; i < 8; ++i)
+                {
+                    char c = m_SectionHeader[v].m_name[i];
+                    if(c < '0' || c > '9')
+                        break;
```
JayFoxRox, line 126: I think there should be a warning or even an error here.

tools/cxbe/Exe.cpp
```diff
@@ -114,6 +115,44 @@ Exe::Exe(const char *x_szFilename)
...
+                    m_SectionHeader_longname[v].m_offset += c - '0';
+                }
+                m_SectionHeader_longname[v].m_longname = new char[256]();
+
+                long tmppos = ftell(ExeFile);
+                fseek(ExeFile, m_Header.m_symbol_table_addr + m_SectionHeader_longname[v].m_offset,
+                      SEEK_SET);
+
+                uint32 i = 0;
+                while(i < 255)
```
JayFoxRox, line 137: nit: This feels like a `for` loop

tools/cxbe/Exe.cpp
```diff
@@ -114,6 +115,44 @@ Exe::Exe(const char *x_szFilename)
...
+                    int c = fgetc(ExeFile);
+                    if(!c || c == EOF)
+                        break;
+                    m_SectionHeader_longname[v].m_longname[i] = c;
+                    ++i;
+                }
+                m_SectionHeader_longname[v].m_longname[i] = 0;
+
+                fseek(ExeFile, tmppos, SEEK_SET);
+                printf(" (long: offset=%u name=%s) ", m_SectionHeader_longname[v].m_offset,
```
JayFoxRox, line 148: I'm not sure how well this `printf` message fits into the output. Do you have example output?

tools/cxbe/Main.cpp
```diff
@@ -19,19 +20,33 @@ int main(int argc, char *argv[])
...
     char szDumpFilename[OPTION_LEN + 1] = { 0 };
     char szXbeTitle[OPTION_LEN + 1] = "Untitled";
+    char szXbeTitleID[OPTION_LEN + 1] = "";
+    char szXbeRegions[OPTION_LEN + 1] = "";
+    char szXbeVersion[OPTION_LEN + 1] = "";
     char szMode[OPTION_LEN + 1] = "retail";
     char szLogo[OPTION_LEN + 1] = "";
+
     bool bRetail;
+    uint32 dwTitleId = 0x4358270F; // CX-9999
```
JayFoxRox, line 30: nit: I'd suggest to stick to 0xFFFF0002 which is more obviously a fake ID and has been used for years in Cxbe. If we add CX-9999 it will just taint another valid XBE Title-ID.

tools/cxbe/Exe.cpp
PQCraft:
```
Exe::Exe: Opening Exe file...OK
Exe::Exe: Reading DOS stub...Found, Ignoring...OK
Exe::Exe: Reading PE header...OK
Exe::Exe: Reading Optional Header...OK
Exe::Exe: Reading Section Headers...
Exe::Exe: Reading Section Header 0x0000...OK 0
Exe::Exe: Reading Section Header 0x0001...OK 1
Exe::Exe: Reading Section Header 0x0002...OK 2
Exe::Exe: Reading Section Header 0x0003...OK 3
Exe::Exe: Reading Section Header 0x0004... (long: offset=4 name=$$XTIMAGE) OK 4
Exe::Exe: Reading Section Header 0x0005... (long: offset=14 name=.debug_abbrev) OK 5
Exe::Exe: Reading Section Header 0x0006... (long: offset=28 name=.debug_aranges) OK 6
Exe::Exe: Reading Section Header 0x0007... (long: offset=43 name=.debug_info) OK 7
Exe::Exe: Reading Section Header 0x0008... (long: offset=55 name=.debug_line) OK 8
Exe::Exe: Reading Section Header 0x0009... (long: offset=67 name=.debug_loc) OK 9
Exe::Exe: Reading Section Header 0x000A... (long: offset=78 name=.debug_ranges) OK 10
Exe::Exe: Reading Section Header 0x000B... (long: offset=92 name=.debug_str) OK 11
Exe::Exe: Reading Sections...
Exe::Exe: Reading Section 0x0000...OK
Exe::Exe: Reading Section 0x0001...OK
Exe::Exe: Reading Section 0x0002...OK
...
```

PQCraft: I could move it to after the OK

tools/cxbe/Main.cpp
```diff
@@ -54,6 +69,87 @@ int main(int argc, char *argv[])
...
+    // interpret region flags
+    if(szXbeRegions[0])
+    {
+        char c;
+        for(int i = 0; (c = szXbeRegions[i]); ++i)
+        {
+            switch(c)
+            {
+                case '-':;
+                    dwRegions = 0;
```
JayFoxRox, line 111: This shouldn't have to set to zero?
I'd expect that the user has to add regions like `-REGION:nj` or similar. I'd expect default to be `njwm`.

PQCraft: `-` means no regions and an empty string means all

tools/cxbe/Main.cpp
```diff
@@ -54,6 +69,87 @@ int main(int argc, char *argv[])
...
+                    goto breakfor;
+                case 'a':;
+                    dwRegions = XBEIMAGE_GAME_REGION_NA | XBEIMAGE_GAME_REGION_JAPAN |
+                                XBEIMAGE_GAME_REGION_RESTOFWORLD |
+                                XBEIMAGE_GAME_REGION_MANUFACTURING;
```
JayFoxRox, line 116: `a` feels like an unnecessary shortcut. Especially if it's `njwm` by default, then it doesn't matter. Most people will keep the default, but edit it on demand

PQCraft: Aye

tools/cxbe/Main.cpp
```diff
@@ -19,19 +20,33 @@ int main(int argc, char *argv[])
...
-                         { szXbeTitle, "TITLE", "title" },
-                         { szMode, "MODE", "{debug|retail}" },
-                         { szLogo, "LOGO", "filename" },
-                         { NULL } };
+    Option options[] = {
+        { szExeFilename, NULL, "exefile" },
+        { szXbeFilename, "OUT", "filename" },
+        { szDumpFilename, "DUMPINFO", "filename" },
+        { szXbeTitle, "TITLE", "title" },
+        { szXbeTitleID, "TITLEID", "{%c%c-%u|%x}" },
```
JayFoxRox, line 41: What does the `%` do here?

PQCraft: Nothing
It's to let the user know what format to write the ID in (<char><char>-<uint> or <hex>)

tools/cxbe/Main.cpp
```diff
@@ -54,6 +69,87 @@ int main(int argc, char *argv[])
...
+        }
+        if(hex)
+        {
+            sscanf(szXbeTitleID, "%x", &dwTitleId);
+        }
+        else
+        {
+            char titlechar[2] = { 'C', 'X' };
+            unsigned titleno = 9999;
+            sscanf(szXbeTitleID, "%c%c-%u", &titlechar[0], &titlechar[1], &titleno);
```
JayFoxRox, line 92: This should just check if each component was scan'd correctly and return an error otherwise.

We shouldn't have to default to CX-9999, either the user provides a fully valid XBE Title ID or they don't.

tools/cxbe/Main.cpp
```diff
@@ -54,6 +69,87 @@ int main(int argc, char *argv[])
...
+                    printf("WARNING: Invalid region char: %c\n", c);
+                    break;
+            }
+        }
+    breakfor:;
+    }
+    else
+    {
+        dwRegions = XBEIMAGE_GAME_REGION_NA | XBEIMAGE_GAME_REGION_JAPAN |
+                    XBEIMAGE_GAME_REGION_RESTOFWORLD | XBEIMAGE_GAME_REGION_MANUFACTURING;
```
JayFoxRox, line 140: Rather than explicitly doing this if `REGION` was not provided, I think it would be better to assign the default value for `REGION`, so it *always* exists, then use the code above the `else`.

tools/cxbe/Xbe.cpp
```diff
@@ -277,19 +285,15 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
...
 
             // TODO: allow configuration
             m_Certificate.dwGameRatings = 0xFFFFFFFF;
 
             // always disk 0, AFAIK
             m_Certificate.dwDiskNumber = 0;
 
-            // TODO: allow configuration
-            m_Certificate.dwVersion = 0;
+            m_Certificate.dwVersion = x_dwVersion;
```
JayFoxRox, line 296: What does this affect?

I believe different XBE versions require different header lengths, but I don't remember the details; is it this field which differentiates?

Makefile
```diff
@@ -118,6 +122,7 @@ else
 main.exe: $(OBJS) $(NXDK_DIR)/lib/xboxkrnl/libxboxkrnl.lib
 	@echo "[ LD       ] $@"
 	$(VE) $(LD) $(NXDK_LDFLAGS) $(LDFLAGS) -out:'$@' $^
+	$(VE) objcopy --rename-section 'XTIMAGE=$$$$XTIMAGE' --rename-section 'XSIMAGE=$$$$XSIMAGE' $@ || exit 0
```
JayFoxRox, line 125: As thrimbor said, this probably doesn't need to be here because the makefile should never really build XBEs in the future.

Alternatively we could extend LLVM to allow such section names.

However, I think most people will just their own binutils line to inject a new section from a raw file, because to make *proper* use of these sections you'll want to use some tool which can encode the Xbox D3D image for these sections (so I doubt they exist as part of an EXE already, unless you have a specific case where you want to run CXBE on some XDK EXE.. but I'm not sure if anybody ever plans to do that or what those XDK EXE look like [if they have these sections for example]).

tools/cxbe/Exe.cpp
JayFoxRox: I don't think the message adds anything (but can quickly become an issue if someone sneaks bad symbols into section names, such as `\n` or `\r`, VT markup etc.).

There's also a inconsistency that section names are not shown for short sections.

So I'd remove the message for now.

tools/cxbe/Main.cpp
JayFoxRox: Wouldn't `-REGION: -TITLE:Foo` be empty? or is that an issue with the CLI arg parser?

tools/cxbe/Xbe.cpp
PQCraft: I couldn't find any docs on that
I assumed it was the version of your program though due to the comment

Makefile
PQCraft: I currently rely on calling NXDK's Makefile from my Makefile so I kind of need this
Also I made my icon by exporting the logo from GIMP as a .dds and overwriting the data from offset 0x800 in an existing .xpr image extracted from an xbe. I will probably make a tool to do this at some point.
I embed the icon by doing this in my main.c:
```
#if PLATFORM == PLAT_XBOX
__asm__ (
    ".section \"XTIMAGE\"\n"
    ".incbin \"icons/engine.xpr\""
);
#endif
```
The linker is a pain and will truncate section names to 8 chars and do some special stuff with `$`. Something like `my$ect` would turn into `my` and `verylongsectname` would turn into `verylong`.

PQCraft:
@JayFoxRox I fixed all of the issues I think

PQCraft:
IDK how to make the "This branch cannot be rebased due to conflicts" message go away so I'm going to open a new PR and see if that works.

PR #652, opened 2023-07-23
cxbe: Fix oob section reads

thrimbor:
Fixes regression introduced by https://github.com/XboxDev/nxdk/commit/d6f0f3a24b35f086889c8d8c2c591656e759fdbb.

The underlying issue is that trailing null bytes are not stored in the exe and thus are not part of `dwSizeofRaw`, causing the null terminator of the `xboxkrnl.exe`-string to get cut off resulting in an oob-read and invalid string comparison, which ultimately produces an xbe with an incorrect import directory (the `hello` sample can be used to test this).

I've changed the exe loader and xbe constructor to instead keep the full section in memory so that valid reads from sections can no longer be out of bounds. This slightly increases memory consumption during the conversion, but that shouldn't be an issue.

### DIFF
--- a/tools/cxbe/Exe.cpp
+++ b/tools/cxbe/Exe.cpp
@@ -7,6 +7,7 @@
 
 #include "Exe.h"
 
+#include <algorithm>
 #include <memory.h>
 #include <stdio.h>
 
@@ -130,10 +131,12 @@ Exe::Exe(const char *x_szFilename)
 
             uint32 raw_size = m_SectionHeader[v].m_sizeof_raw;
             uint32 raw_addr = m_SectionHeader[v].m_raw_addr;
+            uint32 virt_size = m_SectionHeader[v].m_virtual_size;
+            uint32 max_size = std::max(virt_size, raw_size);
 
-            m_bzSection[v] = new uint08[raw_size];
+            m_bzSection[v] = new uint08[max_size];
 
-            memset(m_bzSection[v], 0, raw_size);
+            memset(m_bzSection[v], 0, max_size);
 
             if(raw_size == 0)
             {
--- a/tools/cxbe/Xbe.cpp
+++ b/tools/cxbe/Xbe.cpp
@@ -9,6 +9,7 @@
 #include "Xbe.h"
 #include "Exe.h"
 
+#include <algorithm>
 #include <cstdio>
 #include <cstring>
 #include <locale.h>
@@ -513,8 +514,12 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
                 printf("Xbe::Xbe: Generating Section %.04X...", v);
 
                 uint32 RawSize = m_SectionHeader[v].dwSizeofRaw;
+                uint32 VirtSize = m_SectionHeader[v].dwVirtualSize;
+                uint32 maxSize = std::max(VirtSize, RawSize);
 
-                m_bzSection[v] = new uint08[RawSize];
+                m_bzSection[v] = new uint08[maxSize];
+
+                memset(m_bzSection[v], 0, maxSize);
 
                 memcpy(m_bzSection[v], x_Exe->m_bzSection[v], RawSize);
 


tools/cxbe/Exe.cpp
```diff
@@ -130,10 +131,11 @@ Exe::Exe(const char *x_szFilename)
 
             uint32 raw_size = m_SectionHeader[v].m_sizeof_raw;
             uint32 raw_addr = m_SectionHeader[v].m_raw_addr;
+            uint32 virt_size = m_SectionHeader[v].m_virtual_size;
 
-            m_bzSection[v] = new uint08[raw_size];
+            m_bzSection[v] = new uint08[std::max(virt_size, raw_size)];
 
-            memset(m_bzSection[v], 0, raw_size);
+            memset(m_bzSection[v], 0, virt_size);
```
abaire, line 138: Any reason not to use the actual allocated size when clearing (i.e., the result of the `max`)? At a glance it looks like a potential source of error, even if it's safe in practice.

thrimbor: We'd just clear more bytes than necessary -  `virt_size > raw_size` will result in the buffer getting completely zeroed, `virt_size < raw_size` will result in the buffer being completely filled with a copy of what's in the file.

abaire: Understood that it'd have no actual effect, it's more that it looks potentially erroneous so somebody trying to debug a similar issue to the one that led to this fix might waste time thinking that uninitialized section data was in play.

thrimbor: Personally I'd prefer it to be like what the actual loader at runtime does, but I see what you mean. I've changed it, the performance difference is insignificant anyway.

tools/cxbe/Xbe.cpp
```diff
@@ -513,8 +514,12 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
                 printf("Xbe::Xbe: Generating Section %.04X...", v);
 
                 uint32 RawSize = m_SectionHeader[v].dwSizeofRaw;
+                uint32 VirtSize = m_SectionHeader[v].dwVirtualSize;
+                uint32 maxSize = std::max(VirtSize, RawSize);
 
-                m_bzSection[v] = new uint08[RawSize];
+                m_bzSection[v] = new uint08[maxSize];
+
+                memset(m_bzSection[v], 0, maxSize);
```
RadWolfie, line 522: imo, this should be after copy raw section (couple lines below) then perform `memset(m_bzSection[v] + RawSize, 0, VirtSize);`. This way, it will reduce unnecessary set the memory twice. Raw section is always written to memory. Whilst virtual block of the section is always set to 0.

thrimbor: That would cause an out of bounds write. The virtual size is not a block of zeroes, it's the size of the section when it's loaded into memory.

RadWolfie: I forgot to include ` - RawSize` after VirtSize, so it has been edited. Anyway, yes the section in virtual memory is VirtSize, but the RawSize is from data stored inside xbe file. When xbe's section is load into memory VirtSize is allocated, only first part of the raw section from xbe is written to memory. The remaining size, aka `VirtSize - RawSize` are set to zero'd.

thrimbor: We can't do that either, there's no guarantee that the raw size is smaller than the virtual size. In fact the exe produced by lld contains sections where this is not the case.

RadWolfie: Then wouldn't that be invalid exe binary? Otherwise it wouldn't work on any OSes, including Windows/Xbox. Raw size is always equal or less than virtual size no matter what.

thrimbor: No, having the raw size larger than the virtual size is perfectly fine and does work.

RadWolfie: > The size of the section (for object files) or the size of the initialized data on disk (for image files). For executable images, this must be a multiple of FileAlignment from the optional header. If this is less than VirtualSize, the remainder of the section is zero-filled. Because the SizeOfRawData field is rounded but the VirtualSize field is not, it is possible for SizeOfRawData to be greater than VirtualSize as well. When a section contains only uninitialized data, this field should be zero.

source: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers

Ah, I see raw size could be a bit larger than virtual size. Never mind then. (You can close this discussion as resolve.)

thrimbor:
As it's a somewhat critical bugfix and hasn't gotten any more feedback I'll merge it now.

PR #654, opened 2023-07-28
Added various features

PQCraft:
`tools/cxbe`:
- Added the ability to handle long section names (moved the limit from 8 chars to 255)
- Added an option to set the XBE's title ID to either a hex code (like `FFFF0002`) or a human-readable code (like `CX-9999`)
- Added an option to set the XBE's region flags
- Added an option to set the XBE's version (can be dec, hex, or oct)

`Makefile`:
- Added `XBE_TITLEID` to define a title ID (goes to cxbe as `-TITLEID:$(XBE_TITLEID)`)
- Added `XBE_REGION` to define a title ID (goes to cxbe as `-REGION:$(XBE_REGION)`)
- Added `XBE_VERSION` to define a title ID (goes to cxbe as `-VERSION:$(XBE_VERSION)`)


PQCraft:
NVM the message is still here

PR #655, opened 2023-07-28
Added various features

PQCraft:
#651 died because I suck at using git

Changelog copied from #651:
> `tools/cxbe`:
> - Added the ability to handle long section names (moved the limit from 8 chars to 255)
> - Added an option to set the XBE's title ID to either a hex code (like `FFFF0002`) or a human-readable code (like `CX-9999`)
> - Added an option to set the XBE's region flags
> - Added an option to set the XBE's version (can be dec, hex, or oct)
> 
> `Makefile`:
> - Added `XBE_TITLEID` to define a title ID (goes to cxbe as `-TITLEID:$(XBE_TITLEID)`)
> - Added `XBE_REGION` to define a title ID (goes to cxbe as `-REGION:$(XBE_REGION)`)
> - Added `XBE_VERSION` to define a title ID (goes to cxbe as `-VERSION:$(XBE_VERSION)`)


thrimbor:
This needs some more cleanup work:

* Commit titles don't match our pattern (`cxbe` instead of `tools/cxbe`, `Add` instead of `Added`, missing prefix etc.)
* Several fix-up commits that undo changes made in previous commits (https://github.com/XboxDev/nxdk/pull/655/commits/8bc1c97ef9b97287eb5c01a86e8d5f9d2947e72b) are not squashed into the commit that was faulty
* https://github.com/XboxDev/nxdk/pull/655/commits/fbd77464eaa41636456a3e129010133b84c26913 does not need to be a separate commit
* https://github.com/XboxDev/nxdk/pull/655/commits/2dd28baed327dc8adb68b3164cc0b36b5f54b20f has a rather non-descriptive commit title
* https://github.com/XboxDev/nxdk/pull/655/commits/44069a50cb6ce8bcbc4a8f5fe7274776a2b448a4 commit title is too long
* Unrelated changes in the same commit that should instead get split-up and squashed (https://github.com/XboxDev/nxdk/pull/655/commits/303a4999e2c4a3a2b611adedaef63dd32f22cafd)
* There still are Makefile changes in this PR
* Commits that don't match the formatting style (https://github.com/XboxDev/nxdk/pull/655/commits/46b848978a917ad5299e4ea46fc5955c9abe202d), this has to be fixed in the faulty commits themselves, not with a separate commit at the end. Each commit should pass a check with clang-format.

We also usually don't use commit messages unless they're absolutely necessary to provide context.

tools/cxbe/Main.cpp
```diff
@@ -54,6 +62,40 @@ int main(int argc, char *argv[])
...
+            }
+        }
+        if(hex)
+        {
+            sscanf(szXbeTitleID, "%x", &dwTitleId);
+        }
+        else
+        {
+            char titlechar[2];
+            unsigned titleno;
```
thrimbor, line 94: Using a not explicitly unsigned char can lead to incorrect results when shifting, better use the types defined in `Cxbx.h` or `stdint.h` for these variables.

tools/cxbe/Main.cpp
```diff
@@ -54,6 +62,40 @@ int main(int argc, char *argv[])
...
+            if (sscanf(szXbeTitleID, "%c%c-%u", &titlechar[0], &titlechar[1], &titleno) != 3)
```
thrimbor: This line doesn't follow clang-format rules.

tools/cxbe/Xbe.cpp
```diff
@@ -305,10 +304,7 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
                 XBEIMAGE_MEDIA_TYPE_MEDIA_BOARD | XBEIMAGE_MEDIA_TYPE_NONSECURE_HARD_DISK |
                 XBEIMAGE_MEDIA_TYPE_NONSECURE_MODE;
 
-            // TODO: allow configuration
-            m_Certificate.dwGameRegion = XBEIMAGE_GAME_REGION_MANUFACTURING |
-                                         XBEIMAGE_GAME_REGION_NA | XBEIMAGE_GAME_REGION_JAPAN |
-                                         XBEIMAGE_GAME_REGION_RESTOFWORLD;
+            m_Certificate.dwGameRegion = x_dwRegions;
```
thrimbor, line 308: This change probably belongs to a different commit and breaks building this one.

tools/cxbe/Main.cpp
```diff
@@ -20,12 +20,14 @@ int main(int argc, char *argv[])
...
     char szXbeTitle[OPTION_LEN + 1] = "Untitled";
     char szXbeTitleID[OPTION_LEN + 1] = "";
+    char szXbeRegions[OPTION_LEN + 1] = "";
     char szMode[OPTION_LEN + 1] = "retail";
     char szLogo[OPTION_LEN + 1] = "";
     char szDebugPath[OPTION_LEN + 1] = "";
 
     bool bRetail;
     uint32 dwTitleId = 0xFFFF0002;
+    uint32 dwRegions;
```
thrimbor: This needs to get initialized or lines 121+ will read undefined values.

tools/cxbe/Main.cpp
```diff
@@ -132,7 +178,7 @@ int main(int argc, char *argv[])
             LogoPtr = &logo;
         }
 
-        Xbe *XbeFile = new Xbe(ExeFile, szXbeTitle, dwTitleId, bRetail, LogoPtr, szDebugPath);
+        Xbe *XbeFile = new Xbe(ExeFile, szXbeTitle, dwTitleId, dwRegions, bRetail, LogoPtr, szDebugPath);
```
thrimbor: This does violate the clang-format rules (as do some of the changes carried over from the last commit).

tools/cxbe/Main.cpp
```diff
@@ -178,7 +192,8 @@ int main(int argc, char *argv[])
             LogoPtr = &logo;
         }
 
-        Xbe *XbeFile = new Xbe(ExeFile, szXbeTitle, dwTitleId, dwRegions, bRetail, LogoPtr, szDebugPath);
+        Xbe *XbeFile =
+            new Xbe(ExeFile, szXbeTitle, dwTitleId, dwRegions, dwVersion, bRetail, LogoPtr, szDebugPath);
```
thrimbor: This formatting change does not belong in this commit and it violates the clang-format rules.

tools/cxbe/Xbe.cpp
```diff
@@ -33,7 +33,7 @@ static size_t BasenameOffset(const std::string &path)
 
 // construct via Exe file object
 Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, uint32 x_dwTitleID, uint32 x_dwRegions,
-         bool x_bRetail, const std::vector<uint08> *logo, const char *x_szDebugPath)
+         uint32 x_dwVersion, bool x_bRetail, const std::vector<uint08> *logo, const char *x_szDebugPath)
```
thrimbor: This does not follow clang-format rules.

tools/cxbe/Xbe.cpp
```diff
@@ -376,14 +376,27 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, uint32 x_dwTitleID, uint32 x_d
...
-                if(characteristics & IMAGE_SCN_MEM_WRITE)
-                    m_SectionHeader[v].dwFlags.bWritable = true;
+                // check for $$XTIMAGE or $$XSIMAGE and set the correct flags
+                if(x_Exe->m_SectionHeader_longname[v].m_longname &&
+                   (!strcmp(x_Exe->m_SectionHeader_longname[v].m_longname, "$$XTIMAGE") ||
+                    !strcmp(x_Exe->m_SectionHeader_longname[v].m_longname, "$$XSIMAGE")))
+                {
+                    m_SectionHeader[v].dwFlags.bInsertedFile = true;
+                    m_SectionHeader[v].dwFlags.bHeadPageRO = true;
+                    m_SectionHeader[v].dwFlags.bTailPageRO = true;
```
thrimbor, line 387: How were these flag settings determined? They don't match Halo 2.

tools/cxbe/Xbe.cpp
```diff
@@ -376,14 +376,27 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, uint32 x_dwTitleID, uint32 x_d
 
                 memset(&m_SectionHeader[v].dwFlags, 0, sizeof(m_SectionHeader->dwFlags));
 
-                if(characteristics & IMAGE_SCN_MEM_WRITE)
-                    m_SectionHeader[v].dwFlags.bWritable = true;
+                // check for $$XTIMAGE or $$XSIMAGE and set the correct flags
+                if(x_Exe->m_SectionHeader_longname[v].m_longname &&
+                   (!strcmp(x_Exe->m_SectionHeader_longname[v].m_longname, "$$XTIMAGE") ||
+                    !strcmp(x_Exe->m_SectionHeader_longname[v].m_longname, "$$XSIMAGE")))
```
thrimbor, line 383: I'm not a fan of this hack. I'd rather like to see proper file insertion support in cxbe and something like `-TITLEIMAGE:` / `-SAVEIMAGE:` or similar.

tools/cxbe/Main.cpp
```diff
@@ -91,7 +92,7 @@ int main(int argc, char *argv[])
         {
             char titlechar[2];
             unsigned titleno;
-            if (sscanf(szXbeTitleID, "%c%c-%u", &titlechar[0], &titlechar[1], &titleno) != 3)
+            if(sscanf(szXbeTitleID, "%c%c-%u", &titlechar[0], &titlechar[1], &titleno) != 3)
```
thrimbor: This formatting change does not belong here.

tools/cxbe/Main.cpp
```diff
@@ -29,7 +29,8 @@ int main(int argc, char *argv[])
 
     bool bRetail;
     uint32 dwTitleId = 0xFFFF0002;
-    uint32 dwRegions;
+    uint32 dwRegions = XBEIMAGE_GAME_REGION_NA | XBEIMAGE_GAME_REGION_JAPAN |
+                       XBEIMAGE_GAME_REGION_RESTOFWORLD | XBEIMAGE_GAME_REGION_MANUFACTURING;
```
thrimbor: This probably belongs into the commit that introduced being able to specify the region flags.

tools/cxbe/Main.cpp
```diff
@@ -192,8 +193,8 @@ int main(int argc, char *argv[])
             LogoPtr = &logo;
         }
 
-        Xbe *XbeFile =
-            new Xbe(ExeFile, szXbeTitle, dwTitleId, dwRegions, dwVersion, bRetail, LogoPtr, szDebugPath);
+        Xbe *XbeFile = new Xbe(
+            ExeFile, szXbeTitle, dwTitleId, dwRegions, dwVersion, bRetail, LogoPtr, szDebugPath);
```
thrimbor, line 197: This formatting change does not belong here.

tools/cxbe/Xbe.cpp
```diff
@@ -33,7 +33,8 @@ static size_t BasenameOffset(const std::string &path)
 
 // construct via Exe file object
 Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, uint32 x_dwTitleID, uint32 x_dwRegions,
-         uint32 x_dwVersion, bool x_bRetail, const std::vector<uint08> *logo, const char *x_szDebugPath)
+         uint32 x_dwVersion, bool x_bRetail, const std::vector<uint08> *logo,
+         const char *x_szDebugPath)
```
thrimbor: This formatting change does not belong here.

tools/cxbe/Xbe.cpp
```diff
@@ -394,7 +395,11 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, uint32 x_dwTitleID, uint32 x_d
                        (characteristics & IMAGE_SCN_CNT_CODE))
                         m_SectionHeader[v].dwFlags.bExecutable = true;
 
-                    m_SectionHeader[v].dwFlags.bPreload = true;
+                    char *name = (x_Exe->m_SectionHeader_longname[v].m_longname)
+                                     ? x_Exe->m_SectionHeader_longname[v].m_longname
+                                     : (char *)x_Exe->m_SectionHeader[v].m_name;
+                    m_SectionHeader[v].dwFlags.bPreload =
+                        (strcmp(name, ".debug") && strncmp(name, ".debug_", 7));
```
thrimbor, line 402: Why not just `strncmp(name, ".debug", 6)`?

tools/cxbe/Xbe.h
```diff
@@ -27,7 +27,8 @@ class Xbe : public Error
   public:
     // construct via Exe file object
     Xbe(class Exe *x_Exe, const char *x_szTitle, uint32 x_dwTitleID, uint32 x_dwRegions,
-        uint32 x_dwVersion, bool x_bRetail, const std::vector<uint08> *logo = nullptr, const char *x_szDebugPath = nullptr);
+        uint32 x_dwVersion, bool x_bRetail, const std::vector<uint08> *logo = nullptr,
+        const char *x_szDebugPath = nullptr);
```
thrimbor: This formatting change does not belong here.

Makefile
```diff
@@ -98,7 +98,8 @@ DEPS += $(filter %.cpp.d, $(SRCS:.cpp=.cpp.d))
 
 $(OUTPUT_DIR)/default.xbe: main.exe $(OUTPUT_DIR) $(CXBE)
 	@echo "[ CXBE     ] $@"
-	$(VE)$(CXBE) -OUT:$@ -TITLE:$(XBE_TITLE) $< $(QUIET)
+	$(VE)$(CXBE) -OUT:$@ -TITLE:$(XBE_TITLE) -TITLEID:$(XBE_TITLEID) \
+	    -REGION:$(XBE_REGION) -VERSION:$(XBE_VERSION) $< $(QUIET)
```
thrimbor, line 102: As I said before, using this Makefile to build third-party applications should be considered deprecated. It will get removed and we won't add any features to it.

Using your own Makefile not relying on this one is the way to go. If you absolutely have to rely on this Makefile for now and want to use custom cxbe parameters, override this rule in your Makefile which includes this one.

thrimbor (changes requested):
Exams are over for now, so here's a more in-depth review.

There still are clang-format rule violations, formatting changes grouped into otherwise unrelated commits and even a commit that doesn't build.
The commit messages describe implementation details and imho are unnecessarily verbose, I'd recommend to just remove them all.
`cxbe: Set proper flags on debug sections` isn't a good commit title imho (what is "proper"?), I'd suggest something like `cxbe: Don't preload .debug sections`.
As previously said the Makefile changes won't get merged.

tools/cxbe/Exe.cpp
```diff
@@ -115,7 +116,46 @@ Exe::Exe(const char *x_szFilename)
                 goto cleanup;
             }
 
-            printf("OK %d\n", v);
+            // interpret long section names
+            if(m_SectionHeader[v].m_name[0] == '/')
```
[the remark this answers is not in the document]
PQCraft: I used objcopy to rename a section I made in some inline asm

tools/cxbe/Main.cpp
```diff
@@ -19,18 +19,26 @@ int main(int argc, char *argv[])
...
-        { szDebugPath, "DEBUGPATH", "path" },       { NULL }
+        { szExeFilename, NULL, "exefile" },
+        { szXbeFilename, "OUT", "filename" },
+        { szDumpFilename, "DUMPINFO", "filename" },
+        { szXbeTitle, "TITLE", "title" },
+        { szXbeTitleID, "TITLEID", "{%c%c-%u|%x}" },
+        { szMode, "MODE", "{debug|retail}" },
+        { szLogo, "LOGO", "filename" },
+        { szDebugPath, "DEBUGPATH", "path" },
+        { NULL }
```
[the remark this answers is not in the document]
PQCraft: This was actually done by clang format I think
I remember just adding it to the end

tools/cxbe/Main.cpp
PQCraft: Will adding `unsigned` do?
I prefer to use stdlib types with an stdlib function

tools/cxbe/Main.cpp
PQCraft: I'll see about doing a rebase and running clang-format on all the commits

tools/cxbe/Xbe.cpp
PQCraft: Probably
My rebase was kind of sketchy and one of my commits disappeared into thin air
This was probably a result of me trying to re-create the commit

tools/cxbe/Main.cpp
PQCraft: It does in line 141 under the `else`

tools/cxbe/Xbe.cpp
PQCraft: I used an XBE header dumper on a few XBEs. I think it was Doom 3 and a Quake homebrew XBE I had laying around

tools/cxbe/Main.cpp
thrimbor: That's a different control flow path, I'm talking about 121+

tools/cxbe/Xbe.cpp
PQCraft: There is currently no code to do section insertion as far as i'm aware, and that could probably be removed later once there is code made to add sections.

tools/cxbe/Main.cpp
PQCraft: Yeah more leftovers from the incinerated commit

tools/cxbe/Xbe.cpp
PQCraft: `.debugTest` is not a real debug section name while `.debug` is and anything starting with `.debug_` is

tools/cxbe/Exe.cpp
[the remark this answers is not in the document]
thrimbor: I have to say I'm not a fan, I'd rather stay a mile away from intentionally breaking the PE spec (or giving the impression that we support doing so).

tools/cxbe/Xbe.cpp
thrimbor: Which `.debugTest`? That is not a section I've ever seen nxdk produce (and nothing else should because the dot prefix is reserved for system sections)

Makefile
PQCraft: I really can't include your Makefile in mine. It has too many conflicts. It ended up causing a lot of issues with my Makefile especially with flags. Making another Makefile specifically for the NXDK would require maintaining 2 Makefiles and that would be an absolute pain thanks to the complexity of my Makefile. It turned out to be easier to just call `$(MAKE) -C $(NXDK_DIR)` and pass in stuff.

tools/cxbe/Xbe.cpp
thrimbor: There isn't, as I said I'd rather see it implemented than a name-based hack on top of violating the PE spec. nxdk does not require the usage of these sections nor does it implement the APIs relying on them yet, so we might as well take the time to do things properly.

tools/cxbe/Exe.cpp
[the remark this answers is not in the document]
PQCraft: The thing with `/` is documented in the spec on microsoft's page. While it says that long section names are not supported with executables, objcopy can still produce executables with long section names and they load and run just fine.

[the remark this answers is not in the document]
thrimbor: Yes, it is documented as not being valid for executable images. While it may be possible to hack around the linker for that, it *is* breaking the spec.

tools/cxbe/Xbe.cpp
PQCraft: It could be user generated. I only did it for completeness and it's not like it would hurt performance terribly. The more performance-oriented option would be to check if the first 6 matched `.debug` and then check if the 7th was `\0` or `_`

tools/cxbe/Exe.cpp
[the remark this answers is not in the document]
PQCraft: Also I added this because I actually *need* this to have a title image.
It would incorrectly add debug sections and the image sections as something like `/43` which ofc would not be picked up by the dashboard as an icon.

tools/cxbe/Main.cpp
PQCraft: Ah ok I see
The fix was put in the wrong commit I think

Makefile
thrimbor: I'm not suggesting you include the nxdk Makefile, I'm recommending the exact opposite. Don't rely on it, it's going to get removed.

I don't get why you would need two Makefiles either, you already seem to call cxbe yourself in your Makefile.

tools/cxbe/Xbe.cpp
PQCraft: I don't have the time to write something like that and I need the feature now. Loading the image sections takes up ram to do nothing. This code can be removed later when the better code is added.

tools/cxbe/Xbe.cpp
thrimbor: > It could be user generated

The dot prefix is reserved. I don't care about performance here as much as clean code.

Makefile
PQCraft: I currently use it to fetch all the includes from NXDK's lib dir and use it to build the tools.
I'll remove it since I'm assuming no one will use the new vars.

tools/cxbe/Xbe.cpp
thrimbor: > This code can be removed later when the better code is added.
I'm not really interested in nxdk acquiring *even more* technical debt. We haven't even gotten rid of the one inherited from OpenXDK yet.

tools/cxbe/Exe.cpp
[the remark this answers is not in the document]
JayFoxRox: 1. How does the XDK do it?
2. Can't we tell cxbe to add or rename sections? (Not that I'd like the added complexity, but it's probably better than breaking the spec)

PR #665, opened 2024-04-16
cxbe: Fix broken XBE sections sizes

LoveMHz:
A simple test with the hello world example is enough to generate an XBE with an invalid header.

```c
#include <hal/debug.h>
#include <hal/video.h>
#include <windows.h>

__attribute__((section("TEST")))
void test() {
    Sleep(1000);
}

int main(void) {

    while(1) {
        debugPrint("Hello nxdk!\n");
        Sleep(2000);
    }

    return 0;
}
```

This results in the final section, ``TEST``, being generated with a physical section size greater than the virtual size.

image

The physical size is calculated [here](https://github.com/XboxDev/nxdk/blob/ed21e83da43b8b8da14a9a78cf2528d9b5df86a7/tools/cxbe/Xbe.cpp#L397). The calculated physical size is 512 bytes in this case since the code section is filled with 0xCC. 

Since this is the last section in the PE, the virtual size is pulled from the PE directly. [here](https://github.com/XboxDev/nxdk/blob/ed21e83da43b8b8da14a9a78cf2528d9b5df86a7/tools/cxbe/Xbe.cpp#L392).

---

Calculating the physical size from the last zero-byte makes sense since the Xbox kernel will set the memory tail end of the section to zero based on the difference between physical and virtual size.

This PR attempts to fix this issue by ensuring the section's virtual size is at least as large as the physical size.

### DIFF
--- a/tools/cxbe/Xbe.cpp
+++ b/tools/cxbe/Xbe.cpp
@@ -384,14 +384,6 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
                 m_SectionHeader[v].dwVirtualAddr =
                     x_Exe->m_SectionHeader[v].m_virtual_addr + m_Header.dwPeBaseAddr;
 
-                if(v < m_Header.dwSections - 1)
-                    m_SectionHeader[v].dwVirtualSize =
-                        x_Exe->m_SectionHeader[v + 1].m_virtual_addr -
-                        x_Exe->m_SectionHeader[v].m_virtual_addr;
-                else
-                    m_SectionHeader[v].dwVirtualSize =
-                        RoundUp(x_Exe->m_SectionHeader[v].m_virtual_size, 4);
-
                 m_SectionHeader[v].dwRawAddr = SectionCursor;
 
                 // calculate sizeof_raw by locating the last non-zero value in the raw section data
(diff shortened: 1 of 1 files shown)


tools/cxbe/Xbe.cpp
```diff
@@ -17,6 +17,8 @@
 #include <stdlib.h>
 #include <time.h>
 
+#define MAX(x, y) (((x) > (y)) ? (x) : (y))
+
```
thrimbor, line 21: I'd prefer to just use `std::max` instead

tools/cxbe/Xbe.cpp
```diff
@@ -411,6 +405,17 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
                     m_SectionHeader[v].dwSizeofRaw = RoundUp(r + 2, 4);
                 }
 
+                // calculate virtual size
+                if(v < m_Header.dwSections - 1)
+                    m_SectionHeader[v].dwVirtualSize =
+                        x_Exe->m_SectionHeader[v + 1].m_virtual_addr -
+                        x_Exe->m_SectionHeader[v].m_virtual_addr;
```
thrimbor, line 412: I've never really noticed how weird and seemingly unnecessary this code is - unless it causes some breakage I'm not seeing rn we should imho just remove this part entirely.

tools/cxbe/Xbe.cpp
```diff
@@ -411,6 +405,17 @@ Xbe::Xbe(class Exe *x_Exe, const char *x_szTitle, bool x_bRetail, const std::vec
...
+                else
+                    m_SectionHeader[v].dwVirtualSize = MAX(
+                        RoundUp(x_Exe->m_SectionHeader[v].m_virtual_size, 4),
+                        m_SectionHeader[v].dwSizeofRaw
+                    );
```
thrimbor, line 417: I always kind of assumed `raw size > virtual size` being fine because it's allowed in the PE format (`SizeOfRawData ` is rounded to `FileAlignment`, `VirtualSize` is exact). Does it cause an issue in the XBE loader?

thrimbor (commented):
nit: The commit title prefix should be `cxbe` instead of `cxbx`

I think would be nice if we could lose some of the unnecessary `0xCC' in the file, but that's something we can think about in the future.
Code looks fine, I've left some comments about details.

tools/cxbe/Xbe.cpp
LoveMHz: It appears so. ``XepLoadSection`` takes in an XBE section and, after loading, attempts to clear the tail part of the section.

**Retail 5838**
image

tools/cxbe/Xbe.cpp
LoveMHz: I wouldn't be opposed to making this a separate issue/pr.

In my opinion, there's a lot more of cxbe that should be simplified to trust the generated PE from the linker.

tools/cxbe/Xbe.cpp
LoveMHz: Fair enough. It should be better with the latest commit. 

Feel free to nit style.

PR #670, opened 2024-06-15
winapi: Add clang-format file

Ryzee119:
Add a clang-format file specifcally to the winapi folder. The formatting across the `lib/` varies alot with formatting so this atleast targets this one folder that has bitten me a few times with formatting issues., although this also works pretty well for the `nxdk/` folder

This was based on the bare minimum custom options I could find whilst matching the existing format as best as I could.
This results in the following diff when ran against the winapi folder

~~https://github.com/XboxDev/nxdk/commit/2a2392ae51832295061856f5bf2e0779c88520b3~~ comments1
~~https://github.com/XboxDev/nxdk/commit/f22f6eef47f110caf1a733c1e338539b62f04381~~ comments2
~~https://github.com/XboxDev/nxdk/commit/1a2ab65e225508e3c97b6e9868ed7d974b7999fe~~ comments3
https://github.com/Ryzee119/nxdk/commit/0b314214961b8dec10bbc0b6af24096aac239a08

The majority of these diffs look like genuine formatting issues.

Unfortuntely it does not handle our calling conventions __cdecl or __stdcall and removes a space when it  shouldn't. I think it is related to this issue https://github.com/llvm/llvm-project/issues/34171 and nothing I could do seemed to bypass it.

Where there was a mix of formatting I went with the majority (ie `extern "c" {` vs `extern "c" \n {`

Also dont know if we want a `ColumnLimit` ?

### DIFF
--- a/lib/winapi/.clang-format
+++ b/lib/winapi/.clang-format
@@ -1,0 +1,21 @@
+BasedOnStyle: Microsoft
+ColumnLimit: 0
+IndentCaseLabels: true
+AlignTrailingComments: true
+AlignConsecutiveMacros: true
+InsertBraces: true
+
+SpaceBeforeParens: Custom
+SpaceBeforeParensOptions:
+  AfterControlStatements: true
+  AfterFunctionDefinitionName: true
+  AfterFunctionDeclarationName: true
+
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterControlStatement: false
+  AfterExternBlock: false
+  BeforeElse: false
+  AfterUnion: true
+
+TypeNames: [VOID, PVOID, LPVOID]


lib/winapi/.clang-format
```diff
@@ -0,0 +1,21 @@
+BasedOnStyle: Microsoft
+ColumnLimit: 0
+SortIncludes: false
```
thrimbor, line 3: We might actually want this. The way I prefer it is to separate my include statements into groups, and have each group alphabetically sorted, but the groups as a whole not reordered.

lib/winapi/.clang-format
```diff
@@ -0,0 +1,21 @@
...
+  AfterControlStatements: true
+  AfterFunctionDefinitionName: true
+  AfterFunctionDeclarationName: true
+
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterControlStatement: false
+  AfterExternBlock: false
+  BeforeElse: false
+  AfterUnion: false
```
thrimbor, line 18: Looks a bit weird imho to have `struct` and `union` formatted differently, consistency would be nicer. I used to prefer having opening braces on their own line, but I don't have strong feelings either way, so I'm fine with whatever style you choose.

thrimbor (commented):
Looks mostly good.

The only thing I really don't like is how it reformats some if statements:
image

We've been avoiding the latter because it can lead to things like Apple's "goto fail" bug. Do you know whether clang-format has an option to disallow using `if` without braces entirely?

Also, what happened here? :laughing: 
image

Ryzee119:
> Do you know whether clang-format has an option to disallow using if without braces entirely?

There is this option?
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

lib/winapi/.clang-format
Ryzee119: > AfterUnion: false

Ill make this true (i.e remove the line). I think this was just a case of a mixed bag

lib/winapi/.clang-format
Ryzee119: ill try it, not a fan tbh as not all includes behave nicely and ive had cases of code not compiling after running clang-format :P

If it doesnt break anything ill add it

Ryzee119:
Push new commit with fixes. New diff looks like

https://github.com/XboxDev/nxdk/commit/f22f6eef47f110caf1a733c1e338539b62f04381

Ill hightlight this oddity. Doesnt handle the short hand if/else. Could refactor or a `clang-format-off/on` wrap?

image

JayFoxRox:
I added some comments on https://github.com/XboxDev/nxdk/commit/2a2392ae51832295061856f5bf2e0779c88520b3 and only realized afterwards that thrimbor made some of those arguments already (and that it wasn't the latest revision).

Ryzee119:
I pushed another round. See top comment for new commit

Just need to agree on if we want to align #defines everywhere or not (winerror.h is not aligned, but minwinbase.h), or exclude this section only



Ill refactor this in a follow up before applying the formatting
image

RadWolfie:
> Just need to agree on if we want to align #defines everywhere or not (winerror.h is not aligned, but minwinbase.h), or exclude this section only

In my honest opinion, anything that is grouped together for list of `#define`s should be align for easier reading.

Ryzee119:
Pushed new commit to make #defines line up if in same group. See top comment for latest diff

JayFoxRox:
LGTM.

Some minor notes:

- I left a comment about `GetDayOfWeek` in https://github.com/Ryzee119/nxdk/commit/0b314214961b8dec10bbc0b6af24096aac239a08?diff=unified&w=0#r143421801.

- I'm also not sure if windows.h or other kernel headers have special requirements for `#include` order (as different headers like to add macros which windows.h will respect, or windows.h needs to be included before those macros are added). So I assume we might have to split into different include blocks in the future. However, if the code still works as-is, then it's fine with me.

- The alignment of `#define` blocks might need to a lot of reformatting diffs in the future (especially for `#define Function FunctionA` or constant tables where we manually add entries), so we should probably start https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view and ask users to reformat in a separate commit if necessary.

Issue #675, opened 2024-08-10
USB fails to read configuration of Chihiro Basebd


GXTX:
```
Connect device =>
New full-speed device address 4 assigned.
Manufacturer: SEGA
Product: BASEBD
Parse configuration 1 failed!
```

Ryzee root caused to `MAX_EP_PER_IFACE` requiring to be >10 with USB spec being 15.
https://github.com/XboxDev/nxdk/blob/d6e9c1c1e686e3e1f42f99cbd019bbc94ae1fd8e/lib/usb/libusbohci_xbox/usbh_config_xbox.h#L48

PR #676, opened 2024-09-06
usb: Support maximum endpoints per interface specified by USB spec

thrimbor:
The USB spec specifies a maximum of 15 endpoints per interface (excluding the control endpoint), this change makes us support that.

Fixes https://github.com/XboxDev/nxdk/issues/675

libusbohci preallocates a lot of stuff, so this increases memory usage per device by about 1KiB.


Ryzee119:
Thanks!, one day I will hopefully update to a stack that doesnt to do so much pre-allocation

PR #677, opened 2024-09-09
Winapi reformatting

thrimbor:
A bit of a follow-up to https://github.com/XboxDev/nxdk/pull/670, reformats all the winapi code to comply with the `.clang-format` Ryzee added.

Also utilizing `.git-blame-ignore-revs`, so this reformatting won't mess up the blame view (as long as this PR doesn't get rebased which would change the commit hash), https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view has more info on how that works.


lib/winapi/winbase.h
```diff
@@ -25,43 +24,44 @@ extern "C"
...
+#define INIT_ONCE_STATIC_INIT       {0}
 #define INIT_ONCE_CTX_RESERVED_BITS 2
-#define INIT_ONCE_CHECK_ONLY 0x00000001UL
-#define INIT_ONCE_ASYNC 0x00000002UL
-#define INIT_ONCE_INIT_FAILED 0x00000004UL
-typedef BOOL (CALLBACK *PINIT_ONCE_FN) (PINIT_ONCE, PVOID, PVOID *);
+#define INIT_ONCE_CHECK_ONLY        0x00000001UL
+#define INIT_ONCE_ASYNC             0x00000002UL
+#define INIT_ONCE_INIT_FAILED       0x00000004UL
+typedef BOOL(CALLBACK *PINIT_ONCE_FN)(PINIT_ONCE, PVOID, PVOID *);
```
dracc, line 51: Does removing the white-spaces here not change the behavior?

dracc (commented):
Looks good to me although one line raised some concern.

lib/winapi/winbase.h
thrimbor: That can be the case with `#define`s, where there must not be a whitespace between the name and parameter list for function-style macros, but typedef isn't sensitive to whitespace changes.

dracc: Well, in that case I have no qualms for a merge.

PR #678, opened 2024-10-19
feat: Add video mode sample

ReenigneArcher:
This PR adds a new sample that sets the Xbox to the "best" video mode available.

It was not abundantly clear to me if `XVideoListModes` will loop through in any specific order... if it does perhaps this can be simplified.

I tested this in Xemu, but only get two possible video modes so testing was not extensive.

Note: I'm quite new to nxdk so there may be better ways to accomplish this. I did find one other similar example (https://github.com/Ryzee119/LithiumX/blob/f4471d287d44abc84803d3b901bd4aa7ed459689/src/platform/xbox/platform.c#L99-L123) after I already wrote this, but feel my code is better for selecting the highest resolution available.

### DIFF
--- a/samples/set_best_video_mode/main.cpp
+++ b/samples/set_best_video_mode/main.cpp
@@ -1,0 +1,70 @@
+#include <hal/debug.h>
+#include <hal/video.h>
+#include <windows.h>
+#include <vector>
+
+// Screen dimension globals
+static int SCREEN_WIDTH = 640;
+static int SCREEN_HEIGHT = 480;
+
+int main(void)
+{
+    // set the default video mode
+    XVideoSetMode(SCREEN_WIDTH, SCREEN_HEIGHT, 32, REFRESH_DEFAULT);
+
+    // create an empty list for the available video modes
+    std::vector<VIDEO_MODE> availableVideoModes;
+
+    // save the best video mode here
+    VIDEO_MODE bestVideoMode = {0, 0, 0, 0};
+
+    VIDEO_MODE vm;
+    int bpp = 32;  // Bits per pixel
+    void *p = NULL;  // Initialize to NULL for the first call
+
+    // get the available video modes
+    while (XVideoListModes(&vm, bpp, REFRESH_DEFAULT, &p)) {
+        availableVideoModes.push_back(vm);
+
+        // ensure height is equal to or better than the current best video mode
+        if (vm.height < bestVideoMode.height) {
+            continue;
+        }
+
+        // ensure width is equal to or better than the current best video mode
+        if (vm.width < bestVideoMode.width) {
+            continue;
+        }
+
+        // ensure bpp is equal to or better than the current best video mode
+        if (vm.bpp < bestVideoMode.bpp) {
+            continue;
+        }
+
+        // ensure refresh is equal to or better than the current best video mode
+        if (vm.refresh < bestVideoMode.refresh) {
+            continue;
+        }
+
+        // save the best video mode
+        bestVideoMode = vm;
+    }
+
+    debugPrint("Available video modes:\n");
+    for (VIDEO_MODE vm : availableVideoModes) {
+        debugPrint("Width: %d, Height: %d, BPP: %d, Refresh: %d\n", vm.width, vm.height, vm.bpp, vm.refresh);
+    }
+    Sleep(2000);
+
+    // set the best video mode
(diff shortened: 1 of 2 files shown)


ReenigneArcher:
Any feedback?

samples/set_best_video_mode/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+        if (vm.refresh < bestVideoMode.refresh) {
+            continue;
+        }
+
+        // save the best video mode
+        bestVideoMode = vm;
+    }
+
+    SCREEN_WIDTH = bestVideoMode.width;
+    SCREEN_HEIGHT = bestVideoMode.height;
```
thrimbor, line 54: nit: Use `bestVideoMode.width` and `bestVideoMode.height` directly in the call to `XVideoSetMode` below instead of reassigning the static variables to reduce the potential for errors when someone is reusing this code

thrimbor (commented):
Looks mostly ok to me

samples/set_best_video_mode/main.cpp
ReenigneArcher: It's been a while since I did this, but I believe I wanted to set the variables in the outer scope in case they would be needed for another function call later. Doesn't matter for this example though, so I'm happy to adjust it.

samples/set_best_video_mode/main.cpp
```diff
@@ -0,0 +1,70 @@
...
+int main(void)
+{
+    // set the default video mode
+    XVideoSetMode(SCREEN_WIDTH, SCREEN_HEIGHT, 32, REFRESH_DEFAULT);
+
+    // create an empty list for the available video modes
+    std::vector<VIDEO_MODE> availableVideoModes;
+
+    // save the best video mode here
+    VIDEO_MODE bestVideoMode = {0, 0, 0, 0};
```
mborgerson, line 19: * "best" is subjective. Best in what way? "Highest quality" seems like goal here.
* This feels like it should be a library function so applications don't have to reproduce this mode selection logic

samples/set_best_video_mode/main.cpp
```diff
@@ -0,0 +1,70 @@
...
+    Sleep(2000);
+
+    // set the best video mode
+    XVideoSetMode(bestVideoMode.width, bestVideoMode.height, bestVideoMode.bpp, bestVideoMode.refresh);
+
+    debugPrint("Best video mode:\n");
+    debugPrint("Width: %d, Height: %d, BPP: %d, Refresh: %d\n", bestVideoMode.width, bestVideoMode.height, bestVideoMode.bpp, bestVideoMode.refresh);
+
+    while (1) {
+        Sleep(1000);
```
mborgerson, line 66: For a sample, it would be nice to allow the user to switch between multiple video modes, maybe with some visuals that clearly demonstrate the quality difference

ReenigneArcher: Would using a png/jpg be okay for the visuals?

samples/set_best_video_mode/main.cpp
mborgerson: > Resolution is the primary factor, but it also accounts for bpp and refresh rate.

We can consider bpp (color resolution) and refresh rate (temporal resolution) part of "highest resolution" selection

samples/set_best_video_mode/main.cpp
RadWolfie: For Matt's suggestion, I feel like the user's video selection should go into the "overall" display sample of 2D and 3D render with basic stats info such as how much ram is in use and frames per second. It would be neat to use abaire's 3D objects for misc interaction stuff, such as changing, moving, and rotating the object. At this time, it might be better to go with the basic draw render in 2D. With the interactive video mode selection of course.

mborgerson: fyi, we already have a sample with a 3d object: mesh

samples/set_best_video_mode/main.cpp
ReenigneArcher: Right, I was just using 1080p as an example. I know that resolution is not support by og Xbox.

ReenigneArcher:
Sorry, I do not have much time to work on this so I'm going to close this for now. If anyone else wants to continue this work, I don't have any issue with that.

PR #681, opened 2024-10-29
ci(trigger): fix ci triggers and concurrency

ReenigneArcher:
When working on this locally in my own branch, I noticed some issues with the workflows.

1. Duplicated jobs are run simultaneously due to accepting both `push` and `pull_request` events.
2. All types of pull_request events will trigger the workflow. This is generally not a good idea. For example, these workflows should not trigger if you label the workflow, or request a review. See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request
3. There is no concurrency control. Concurrency will allow you to cancel in process workflows in cases where a PR is updated again while the current workflows are running. See: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs

This PR fixes all of the above issues.

Additionally, many of the actions are also using very outdated versions. Generally, I use @dependabot to keep these updated through automatic PRs, I'd be happy to add a secondary PR to add this in. Also works for git submodules, which is pretty nice.

### DIFF
--- a/.github/workflows/build_docker_image.yml
+++ b/.github/workflows/build_docker_image.yml
@@ -1,6 +1,18 @@
 name: Build Docker image
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   main:
--- a/.github/workflows/build_samples.yml
+++ b/.github/workflows/build_samples.yml
@@ -1,6 +1,18 @@
 name: Build Samples
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   windows:


thrimbor:
>     1. Duplicated jobs are run simultaneously due to accepting both `push` and `pull_request` events.

This only happens when you push a branch that also has a pull-request open, and is afaik something that cannot be prevented properly, as it is once run in the context of the repository you're pushing to, and once in the context of the repository the pull request was filed against.
Your change would simply disable the workflow for all branches except the master branch, which is not good - checking that a commit builds properly should work for all branches, not only those that have a pull-request open.

>     2. All types of pull_request events will trigger the workflow. This is generally not a good idea. For example, these workflows should not trigger if you label the workflow, or request a review. See: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

This is not the case, as the documentation states:
> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened

So your changes have no effect.

>     3. There is no concurrency control. Concurrency will allow you to cancel in process workflows in cases where a PR is updated again while the current workflows are running. See: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs

While I would like to see strict ordering of workflow runs in a pull-request, I don't like them getting aborted. In fact I do the exact opposite on purpose, pushing in small steps to ensure CI completes for every commit (I do that *a lot* when cleaning up and reworking complex feature branches).


> Additionally, many of the actions are also using very outdated versions. Generally, I use @dependabot to keep these updated through automatic PRs, I'd be happy to add a secondary PR to add this in. Also works for git submodules, which is pretty nice.

If there's a solution to better handle outdated actions I'm open to considering it.

ReenigneArcher:
I did notice your main contributors often don't work inside of PRs (or using PRs in forks instead of to this repo) so I understand wanting to keep the push event for all branches. It can be really annoying to consume all of these runners though since GitHub only provides 20 for free, especially for anyone who maintains a lot of other repos.

These settings do work really well IF you work in PRs though. Typically that's what I do, often with draft PRs when there is still a lot of work to do.

> This is not the case, as the documentation states:

> >    By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened

> So your changes have no effect.

Good point. I'm usually explicit as I don't trust GitHub to not adjust defaults with a CoPilot typo by mistake.

I'll close this, but you'll have it for future reference if you later decide to use any of this.

PR #683, opened 2024-10-29
ci: Install lld explicitly on macOS

ReenigneArcher:
The build for macOS is currently failing due to missing `lld` dependency.

Note: If it's preferable I can combine my PRs into one, although it's probably easier to review them separately. Just let me know whatever you prefer.


thrimbor:
Thanks for the fix, but before merging I'd like to see the commit title changed to something that is more descriptive and follow the pattern of the commit titles we currently have in nxdk, I suggest something like `ci: Install lld explicitly on macOS`

ReenigneArcher:
Done.

PR #685, opened 2024-12-22
debug: Handle carriage return

Ryzee119:
Prevents CR characters being printed like below and is just generally useful to handle CR correctly.

image


JayFoxRox:
Looks good to me.

PR #686, opened 2024-12-24
nvnetdrv v2 part 1

Ryzee119:
This is a culmination of work over ~12 months or so.

With nvnetdrv being used in StellarOS, it has seen many hundreds/thousands of people using it through the online updater and FTP server and a few edges cases started arising.

As part of the debugging effort in collaboration with LoveMhz/StellarOS developers, a significant amount of refactoring and optimizations were done.

This has turned into a reasonably big PR (sorry!) so I'm calling it version 2, however the primary changes are

1. Refactoring of variable names
2. Reduced memory footprint significantly
3. Less threads/context switching
4. Fixes some bugs.

Hopefully I have broken up the commits into logical units.

My stress test involved 8 x parallel iperf connections, 10 parallel FTP connections (with hash check of files after), web server and sntp client at the same time.

### DIFF
--- a/lib/net/nvnetdrv/nvnetdrv.c
+++ b/lib/net/nvnetdrv/nvnetdrv.c
@@ -77,40 +77,40 @@ static HANDLE g_irqThread;
 static KEVENT g_irqEvent;
 
 // Manage RX ring
-static volatile struct descriptor_t *g_rxRingDescriptors;
+static volatile struct descriptor_t *g_rxRing;
 static KSEMAPHORE g_rxRingFreeDescriptors;
 static HANDLE g_rxRingRequeueThread;
-static size_t g_rxRingBeginIndex;
-static size_t g_rxRingEndIndex;
-static uint8_t *g_rxRingBuffers;
+static size_t g_rxRingHead;
+static size_t g_rxRingTail;
+static uint8_t *g_rxRingUserBuffers;
 static uint32_t g_rxRingBufferVtoP;
 
 // Manage RX buffer callbacks to user network stack
 static nvnetdrv_rx_callback_t g_rxCallback;
-static KSEMAPHORE g_rxCallbackQueued;
+static KSEMAPHORE g_rxPendingCount;
 static HANDLE g_rxCallbackThread;
 struct rx_misc_t *g_rxCallbackQueue;
-static size_t g_rxCallbackEndIndex;
+static size_t g_rxCallbackTail;
 
 // Manage RX buffer pool to supply RX ring
 static void **g_rxBuffPool;
 static size_t g_rxBuffPoolHead;
-static size_t g_rxBuffCnt;
+static size_t g_rxRingSize;
 static RTL_CRITICAL_SECTION g_rxBuffPoolLock;
 static KSEMAPHORE g_rxFreeBuffers;
 
 // Manage TX ring
-static volatile struct descriptor_t *g_txRingDescriptors;
-static size_t g_txRingBeginIndex;
-static atomic_size_t g_txRingEndIndex;
-static atomic_size_t g_txRingDescriptorsInUseCount;
-static KSEMAPHORE g_txRingFreeDescriptors;
-struct tx_misc_t tx_misc[TX_RING_SIZE];
+static volatile struct descriptor_t *g_txRing;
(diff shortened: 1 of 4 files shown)


lib/net/nvnetdrv/nvnetdrv.c
```diff
@@ -79,37 +79,34 @@ static KINTERRUPT g_interrupt;
...
+// Manage RX and TX rings
+static struct descriptor_t *g_rxRing;
+static struct descriptor_t *g_txRing;
+static size_t g_rxRingSize;
+static size_t g_txRingSize;
 static size_t g_rxRingHead;
-static size_t g_rxRingTail;
+static size_t g_txRingHead;
+static atomic_size_t g_txPendingCount;
+static atomic_size_t g_rxPendingCount;
```
thrimbor: The type of `g_rxPendingCount` seems to have been changed accidentally here

lib/net/nvnetdrv/nvnetdrv.c
```diff
@@ -201,7 +201,7 @@ static void nvnetdrv_handle_rx_irq (void)
             goto release_packet;
         }
 
-        if (!(flags & NV_RX_ERROR) || (flags & NV_RX_FRAMINGERR)) {
+        if (!(flags & NV_RX_ERROR) || ((flags & NV_RX_ERROR_MASK)) = NV_RX_FRAMINGERR)) {
```
thrimbor: Considering that `NV_RX_SUBTRACT1` isn't part of `NV_RX_ERROR_MASK` either, wouldn't it make more sense to simplify the whole check to something like `if ((flags & NV_RX_CRIT_ERRORS) == 0)` with `#define NV_RX_CRIT_ERRORS (NV_RX_ERROR1|NV_RX_ERROR2|NV_RX_ERROR3|NV_RX_ERROR4|NV_RX_CRCERR|NV_RX_OVERFLOW)`, or similar?

thrimbor (changes requested):
I've only managed to go through up to and including `nvnetdrv: Soft error if framing error only` and left some feedback.

When that is merge-able I'll probably split that off into a separate PR to make this a bit easier to review.

lib/net/nvnetdrv/nvnetdrv.c
Ryzee119: Fixed

lib/net/nvnetdrv/nvnetdrv.c
Ryzee119: Good idea, done

Ryzee119:
>When that is merge-able I'll probably split that off into a separate PR to make this a bit easier to review.
Ive renamed this to part one and dropped the other commits. Will raised another when merged and rebase the other commits

lib/net/nvnetdrv/nvnetdrv.c
```diff
@@ -79,37 +79,34 @@ static KINTERRUPT g_interrupt;
 static HANDLE g_irqThread;
 static KEVENT g_irqEvent;
 
-// Manage RX ring
-static volatile struct descriptor_t *g_rxRing;
-static KSEMAPHORE g_rxRingFreeDescriptors;
-static HANDLE g_rxRingRequeueThread;
+// Manage RX and TX rings
+static struct descriptor_t *g_rxRing;
+static struct descriptor_t *g_txRing;
```
thrimbor, line 84: Hm, I don't think we can drop the `volatile` here as the hw modifies the `flags` member. Interestingly this change doesn't lead to a bug even with LTO, probably because `nvnetdrv_handle_rx_irq` copies the address into a proper volatile pointer.

Imho the initialization of these should be changed, too, as it's currently casting to a non-volatile pointer, if you want to squeeze that into one of the commits in this PR:
```
// Setup the TX and RX ring descriptor pointers
    g_rxRing = (struct descriptor_t *)descriptors;
    g_txRing = (struct descriptor_t *)descriptors + RX_RING_SIZE;
```

Ryzee119: ah yep gotcha, fixed thanks

thrimbor (approved):

thrimbor:
Looks good to me, merging

PR #688, opened 2024-12-29
zlib: Update to 1.3.1

Ryzee119:
Currently building with nxdk results in warning spam from zlib

` warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C23 [-Wdeprecated-non-prototype]`

This PR just brings zlib up to 1.3.1 which fixes these issues.

Compiled all samples, and ran the sdl samples with no issues


thrimbor:
Seems to work fine (tested with a modified SDL_Image sample), merging

PR #689, opened 2024-12-29
nvnetdrv v2 part 2

Ryzee119:
Continuation from https://github.com/XboxDev/nxdk/pull/686

Will likely end up in 3 parts.

Primary change here are to pass ring sizes to the init function instead of #defines. Previously we could have different hardware ring size and queue size. Its actually much simpler to just have the hardware ring size exactly the same size as the queue. It also works better with LWIP as we can create pbuf pools to exactly match our requirement.

Most other commits are just some minor bug fixes and house keeping.

This fixes a bug that on soft reset or coming from XDK app where the network would not retrieve an IP address in some cases.

### DIFF
--- a/lib/net/nvnetdrv/nvnetdrv.c
+++ b/lib/net/nvnetdrv/nvnetdrv.c
@@ -716,8 +716,7 @@ void nvnetdrv_stop_txrx (void)
         KeDelayExecutionThread(KernelMode, FALSE, FIFTY_MICRO);
     }
 
-    reg32(NvRegLinkSpeed) = 0;
-    reg32(NvRegTransmitPoll) = 0;
+    reg32(NvRegTxRxControl) = 0;
 }
 
 int nvnetdrv_acquire_tx_descriptors (size_t count)
--- a/lib/net/nvnetdrv/nvnetdrv.c
+++ b/lib/net/nvnetdrv/nvnetdrv.c
@@ -498,22 +498,20 @@ int nvnetdrv_init (size_t rx_buffer_count, nvnetdrv_rx_callback_t rx_callback, s
     KeDelayExecutionThread(KernelMode, FALSE, TEN_MICRO);
     reg32(NvRegTxRxControl) = 0;
     KeDelayExecutionThread(KernelMode, FALSE, TEN_MICRO);
-
-    // Disable interrupts while we initialise the NIC
-    reg32(NvRegIrqMask) = 0;
     reg32(NvRegMIIMask) = 0;
+    reg32(NvRegIrqMask) = 0;
+    reg32(NvRegWakeUpFlags) = 0;
+    reg32(NvRegPollingControl) = 0;
+    reg32(NvRegTxRingPhysAddr) = 0;
+    reg32(NvRegRxRingPhysAddr) = 0;
+    reg32(NvRegTransmitPoll) = 0;
+    reg32(NvRegLinkSpeed) = 0;
 
     // Acknowledge any existing interrupts status bits
     reg32(NvRegTransmitterStatus) = reg32(NvRegTransmitterStatus);
     reg32(NvRegReceiverStatus) = reg32(NvRegReceiverStatus);
     reg32(NvRegIrqStatus) = reg32(NvRegIrqStatus);
     reg32(NvRegMIIStatus) = reg32(NvRegMIIStatus);
-
-    // Clear latent registers
-    reg32(NvRegWakeUpFlags) = 0;
-    reg32(NvRegPollingControl) = 0;
-    reg32(NvRegTransmitPoll) = 0;
-    reg32(NvRegLinkSpeed) = 0;
 
     // Reset local ring tracking variables
     g_rxRingHead = 0;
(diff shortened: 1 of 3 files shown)


lib/net/nvnetdrv/nvnetdrv.c
```diff
@@ -735,40 +733,21 @@ int nvnetdrv_acquire_tx_descriptors (size_t count)
...
+    for (size_t i = 0; i < count; i++) {
+        NTSTATUS status = KeWaitForSingleObject(&g_txRingFreeCount, Executive, KernelMode, FALSE, timeout);
+        if (!NT_SUCCESS(status) || !g_running) {
+            if (i > 0) {
+                KeReleaseSemaphore(&g_txRingFreeCount, IO_NETWORK_INCREMENT, i, FALSE);
             }
-        }
-
-        if (!g_running) {
+            assert(!g_running);
```
thrimbor, line 741: `nvnetdrv_stop` sets `g_running` to false before releasing the semaphore, so this could get hit if the driver gets stopped while this thread waits. I'd suggest keeping the `if (!g_running) return false;` instead.

lib/net/nvnetdrv/nvnetdrv.c
```diff
@@ -735,40 +733,21 @@ int nvnetdrv_acquire_tx_descriptors (size_t count)
...
             return false;
         }
-
-        // If we have claimed all the tx descriptors. We are done.
-        if (i == (count - 1)) {
-            break;
+        // Try again, but wait now
+        if (status == STATUS_TIMEOUT) {
+            i--;
+            timeout = NULL;
```
thrimbor, line 747: The intention behind my more complex code was to prevent the dining philosophers problem. If I understand your code correctly, that could actually now deadlock (imagine four descriptors, two threads grab two each, both wait indefinitely for a third one to become available and never back off, and no other thread can now send).
I'm not against simplifying the code, but what I would suggest instead is to choose a sensible timeout, always use the waiting version of the call (I don't see any benefit in trying twice), and fully back out and return false as soon as even a single wait fails. Or even use the timeout for just the first loop iteration.

thrimbor (commented):
Sorry for taking so long to get around to reviewing this.
Most of it is completely fine (as usual :smiley: ), but I left a bit of feedback on the changes to `nvnetdrv_acquire_tx_descriptors`.

lib/net/nvnetdrv/nvnetdrv.c
```diff
@@ -441,17 +447,25 @@ int nvnetdrv_init (size_t rx_buffer_count, nvnetdrv_rx_callback_t rx_callback)
     }
 
     // Allocate memory for TX and RX ring descriptors.
-    void *descriptors = MmAllocateContiguousMemoryEx((RX_RING_SIZE + TX_RING_SIZE) * sizeof(struct descriptor_t), 0,
+    void *descriptors = MmAllocateContiguousMemoryEx((g_rxRingSize + g_txRingSize) * sizeof(struct descriptor_t), 0,
```
disean, line 450: For performance reasons, you may want to align this pointer on a 64-byte boundary.

JayFoxRox: `MmAllocateContiguousMemoryEx` should allocate physical memory pages, so I'm not sure if this could ever *not* be aligned to 0x1000 (which is a multiple of 0x40)?

thrimbor: Yeah it's always page-aligned. `MmAllocateContiguousMemoryEx` has an alignment parameter, but that only does something if you want a POT alignment bigger than 4KiB.

lib/net/nvnetdrv/nvnetdrv.c
Ryzee119: Thanks this is a good point. I'm happy to bench this specific change and we can manage it in a separate PR if needed.
For the moment I have dropped this commit as the original code had no problems and it was purely a code cleanup attempt.

lib/net/nvnetdrv/nvnetdrv.c
Ryzee119: I believe this is now resolved after dropping the related commit. See [other comment](https://github.com/XboxDev/nxdk/pull/689#discussion_r1970718063)

lib/net/nvnetdrv/nvnetdrv.c
```diff
@@ -441,17 +441,25 @@ int nvnetdrv_init (size_t rx_buffer_count, nvnetdrv_rx_callback_t rx_callback)
...
 
     // Allocate memory for TX and RX ring descriptors.
-    void *descriptors = MmAllocateContiguousMemoryEx((RX_RING_SIZE + TX_RING_SIZE) * sizeof(struct descriptor_t), 0,
+    void *descriptors = MmAllocateContiguousMemoryEx((g_rxRingSize + g_txRingSize) * sizeof(struct descriptor_t), 0,
                                                      0xFFFFFFFF, 0, PAGE_READWRITE);
     if (!descriptors) {
         return NVNET_NO_MEM;
     }
 
+    // Allocate memory to store data associated with outgoing and incoming transfers
```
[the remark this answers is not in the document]
Ryzee119: Fixed

PR #693, opened 2025-01-10
Pkgconfig CMake compatibility & switch to pkgconf

thrimbor:
The last two commits are from https://github.com/XboxDev/nxdk/pull/631, with only the commit title updated. They make it possible for CMake to find the libraries to link against via pkg-config.

These two commits had the problem that they fix CMake, but by doing so are breaking direct usage of `nxdk-pkg-config` (and probably meson) because the emitted linker arguments are in `ld` syntax, while we rely on the syntax of `link.exe`.

This PR circumvents this issue by switching from whatever pkg-config the host OS provides to explicitly requiring `pkgconf` (which is the default `pkg-config` implementation on Arch, Alpine and probably others anyway).
`pkgconf` provides a `--msvc-syntax` parameter that can be used to get the required `link.exe`-style linker parameters. This would add an additional `.lib` suffix to the library paths (resulting in things like `libjpeg.lib.lib`), so I added a small `sed` command to fix that.

### DIFF
--- a/bin/nxdk-pkg-config
+++ b/bin/nxdk-pkg-config
@@ -9,9 +9,9 @@ export PKG_CONFIG_PATH=""
 export PKG_CONFIG_SYSROOT_DIR=""
 export PKG_CONFIG_LIBDIR=${NXDK_DIR}/lib/pkgconfig:${NXDK_DIR}/share/pkgconfig
 
-[ "$1" = '--version' ] && exec pkg-config --version
-exec pkg-config \
+[ "$1" = '--version' ] && exec pkgconf --version
+exec pkgconf \
     --define-variable=NXDK_DIR=${NXDK_DIR} \
     --define-prefix \
     --static \
-    "$@"
+    "$@" | sed 's/\.lib\.lib/\.lib/g'
--- a/.github/workflows/build_samples.yml
+++ b/.github/workflows/build_samples.yml
@@ -42,7 +42,7 @@ jobs:
           HOMEBREW_NO_ANALYTICS: 1
           HOMEBREW_NO_INSTALL_CLEANUP: 1
         run: |
-          brew install lld llvm coreutils
+          brew install lld llvm coreutils pkgconf
       - name: Build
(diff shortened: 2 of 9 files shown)


bin/nxdk-pkg-config
```diff
@@ -9,9 +9,9 @@ export PKG_CONFIG_PATH=""
...
 
-[ "$1" = '--version' ] && exec pkg-config --version
-exec pkg-config \
+[ "$1" = '--version' ] && exec pkgconf --version
+exec pkgconf \
     --define-variable=NXDK_DIR=${NXDK_DIR} \
     --define-prefix \
     --static \
-    "$@"
+    "$@" | sed 's/\.lib\.lib/\.lib/g'
```
JayFoxRox, line 17: What does this do?

JayFoxRox, line 17: I'm not sure I follow the PR description.

I thought the .lib.lib only happens when using `--msvc-syntax`, but now `--msvc-syntax` is absent?

thrimbor: > I thought the .lib.lib only happens when using --msvc-syntax, but now --msvc-syntax is absent?

It does, but passing in `--msvc-syntax` is up to the user. Generally, a user will want to specify it for the linker (which follows `link.exe` parameter syntax), but not for the compiler (which follows gcc/clang syntax), and not for CMake (which expects `ld` syntax).

> (Also, why does it produce .lib.lib? This feels like a bug - was it reported?)

Not a bug, [it matches freedesktop's `pkg-config` implementation](https://gitlab.freedesktop.org/pkg-config/pkg-config/-/blob/master/README.win32#L27).
To convert the `ld` to `link.exe` syntax, the `-l` is stripped and a `.lib` added (for example, on my host system `pkg-config libpng --libs` produces `-lpng16`, whereas with `--msvc-syntax` I get `png16.lib` instead, because with MSVC libraries by convention have a `.lib` ending).
We already have our parameters end in `.lib`, with the one added by pkgconf we get `.lib.lib`

PR #695, opened 2025-01-18
Use TSC for QueryPerformance* API

thrimbor:
The first of these two commits changes our implementation of `QueryPerformanceCounter` and `QueryPerformanceFrequency` to use the TSC (like later XDKs did) instead of the kernel's less accurate internal counter.

The second commit improves compatibility with non-stock Xbox systems by using a constructor to compute the TSC frequency on startup instead of using a hard-coded value.
It uses a lookup-table of known CPU clock multipliers and FSB divider and multiplier values from the northbridge to derive this frequency from the crystal frequency. This is done with 64 bit integer math and fractions of a Hz to not introduce any errors during the calculations.
The only way this could be wrong should be using a non-stock crystal frequency.

Using this code in xemu requires https://github.com/xemu-project/xemu/pull/1551 so any reasonably recent version should do.
The code is almost unchanged from my old branch that wasn't upstreamed yet (https://github.com/thrimbor/nxdk/tree/tsc).

Supersedes https://github.com/XboxDev/nxdk/pull/663

### DIFF
--- a/lib/winapi/profiling.c
+++ b/lib/winapi/profiling.c
@@ -3,9 +3,13 @@
 // SPDX-FileCopyrightText: 2019-2025 Stefan Schmidt
 
 #include <assert.h>
+#include <intrin.h>
 #include <profileapi.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <xboxkrnl/xboxkrnl.h>
+
+static LARGE_INTEGER tsc_freq = {.QuadPart = 733333333};
 
 BOOL QueryPerformanceCounter (LARGE_INTEGER *lpPerformanceCount)
 {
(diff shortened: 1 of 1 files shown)


lib/winapi/profiling.c
```diff
@@ -1,24 +1,95 @@
...
+
+static uint64_t get_cpu_multiplier (void)
+{
+    // Read EBL_CR_POWERON
+    const uint64_t poweron_reg = __readmsr(0x2A);
+
+    // Chapter 2.22 of https://www.intel.com/content/dam/develop/external/us/en/documents/335592-sdm-vol-4.pdf
+    const uint64_t multiplier_pattern = (poweron_reg >> 22) & 0b101111;
+
+    // https://web.archive.org/web/20040407043044/http://www.intel.com:80/design/mobile/datashts/29834006.pdf
```
GXTX, line 38: image

Neither document mentioned in this PR has information on >10X multipliers, this is something I brought up (and wasn't answered) in xemu-project/xemu/pull/1551, I also attempted to search Linux and they too seem not to mention support for >10X. Where are you getting these numbers?

thrimbor: From the CPUs, specifically SL6BY, SL6C6 and SL6C7.

lib/winapi/profiling.c
```diff
@@ -19,6 +23,73 @@ BOOL QueryPerformanceFrequency (LARGE_INTEGER *lpFrequency)
...
+    if (multiplier_10x == 0) {
+        // Could not determine clock multiplier, fall back to default TSC freq
+        return;
+    }
+
+    uint32_t fsbreg;
+    HalReadWritePCISpace(0, 0x60, 0x6C, &fsbreg, sizeof(fsbreg), FALSE);
+    const uint64_t fsb_divider = fsbreg & 0xFF;
+    const uint64_t fsb_multiplier = (fsbreg >> 8) & 0xFF;
+    const uint64_t fsb_freq = (xtal_freq_mhz / fsb_divider) * fsb_multiplier;
```
Ryzee119, line 88: Can fsb_divider *ever* be zero? (div 0 issue)

lib/winapi/profiling.c
```diff
@@ -19,6 +23,73 @@ BOOL QueryPerformanceFrequency (LARGE_INTEGER *lpFrequency)
...
+{
+    // Read EBL_CR_POWERON
+    const uint64_t poweron_reg = __readmsr(0x2A);
+
+    // Chapter 2.22 of https://www.intel.com/content/dam/develop/external/us/en/documents/335592-sdm-vol-4.pdf
+    const uint64_t multiplier_pattern = (poweron_reg >> 22) & 0b101111;
+
+    // https://web.archive.org/web/20040407043044/http://www.intel.com:80/design/mobile/datashts/29834006.pdf
+    // clang-format off
+    switch (multiplier_pattern) {
```
Ryzee119, line 40: Should we have a default case? No impact to functionality but might causes warnings with -wall?

lib/winapi/profiling.c
thrimbor: Interesting question, would be interesting to experiment with that value, but atm I don't know (and can't test). I added a check for zero that defaults to the 733MHz value.

mborgerson:
> The only way this could be wrong should be using a non-stock crystal frequency.

We can measure the tsc freq with a reference clock

Issue #696, opened 2025-01-18
Compiling with LTO strips automount_d_drive function


Ryzee119:
Simply compile the sdl_image sample `make CFLAGS=-flto`

![Image](https://github.com/user-attachments/assets/a8927475-bfa5-4ff1-8f4a-090b022e6f1b)

Tried a few common things, like adding volatiles, `__attribute__((used))` etc. But no luck and not immediately clear how to fix.

Results in code that works without LTO that breaks with LTO.

clang v19

Ryzee119:
Ah just figured it out  by editing `automount_d.c`

```diff
- __attribute__((section(".CRT$XIT"))) int (__cdecl *const automount_d_drive_p)(void) = automount_d_drive;
+ __attribute__((section(".CRT$XIT"), used)) int (__cdecl *const automount_d_drive_p)(void) = automount_d_drive;
```

ReenigneArcher:
Might not be entirely related, but I had to add this when using CMake.

https://github.com/LizardByte/Moonlight-XboxOG/blob/bfa16dd85d5e2a02559deac59eeb3898af2590ca/CMakeLists.txt#L32-L35

Ryzee119:
> Might not be entirely related, but I had to add this when using CMake.
> 
> https://github.com/LizardByte/Moonlight-XboxOG/blob/bfa16dd85d5e2a02559deac59eeb3898af2590ca/CMakeLists.txt#L32-L35

Right yea, that still will get stripped with LTO as that is exactly what scenario I have

PR #697, opened 2025-01-18
nxdk: Prevent LTO stripping automount_d_drive

Ryzee119:
Fixes https://github.com/XboxDev/nxdk/issues/696.


thrimbor:
Interesting, the MSDN documentation is imho a bit unclear on whether the symbol can still get removed somehow.
You might also want to report this upstream because it *could* be a deviation from what MSVC does (I didn't check myself).

PR #707, opened 2025-02-21
README.md: Fix Build Status badge

MasonT8198:


README.md
```diff
@@ -16,7 +16,7 @@ Notable features:
 
 Build Status
 ------
-![CI Status for "Build Samples"](https://github.com/XboxDev/nxdk/workflows/Build%20Samples/badge.svg?branch=master&event=push)
+![CI Status for "Build Samples"](https://github.com/XboxDev/nxdk/actions/workflows/build_samples.yml/badge.svg)
```
RadWolfie, line 19: For what reason have the parameters been removed? Having them removed will default to reading the last state activity from all branches and all events. Although, the absent of branch parameter in theory will use default branch used. Otherwise it will use all branches according to the [doc](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge).

MasonT8198: > For what reason have the parameters been removed? Having them removed will default to reading the last state activity from all branches and all events. Although, the absent of branch parameter in theory will use default branch used. Otherwise it will use all branches according to the [doc](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge).

In any instance with the old status badge it would just appear "No Status" for no apparent reason, I tried multiple permutations of the URL to resolve that issue to no avail, I decided to omit that segment of the badge and instead point it to the build_samples.yml in the action, if you want to revert it back to what it was before , by all means

RadWolfie: The old url link indeed is broken, if updated to the new url link. It is functional, below shows the new link is functional:
- "Build Samples" with branch and event (parameter with branch and event)
- "Build Samples" with branch only (parameter with branch only)
- "Build Samples" with event only (parameter with event only)

MasonT8198: Sounds good to me

PR #708, opened 2025-02-21
pbkit: Add various operations.

abaire:
Upstreams a bunch of operations from https://github.com/abaire/nxdk_pgraph_tests/blob/main/src/nxdk_ext.h


JayFoxRox:
nit: Please don't end commit titles in punctation

PR #712, opened 2025-02-26
pbkit: Fix pb_kill softlock in shutdown handler

### LINKED ISSUE #710: XBE's using pbkit sometimes lock up when rebooted via XBDM "reboot warm" or returning from `main`
When using a debug bios, executables that utilize pbkit will sometimes lock up during handling of a "reboot warm" command. This command is generally used when restarting an executable under a debugger (e.g., using `xbdm_gdb_bridge`).

Repro:
1) Build the "triangle" sample and copy it onto the Xbox harddrive (also confirmed reproducibility via xemu)
2) Connect to the Xbox via XBDM and launch the triangle sample
3) Send a "reboot warm" command

Sometimes (anecdotally ~5-10%) the reboot will freeze the running XBE as well as the XBDM command handler. A hard reboot is required to recover. Testing under xemu it seems that the CPU is still active in a tight loop.

Note that after a hard reboot, the boot process may stall for 30 seconds or so; this is likely due to the fact that the warm reboot attempts to reconnect XBDM notification sockets that were active prior to the reboot command.

So far I have been unsuccessful in reproducing this issue with the nxdk "hello" sample, and have also been unable to reproduce with either retail XBEs or non-nxdk-based homebrew binaries.

To reproduce with xbdm_gdb_bridge:
```
xbdm <xbox_ip_address> -v3 -s -- /launch e:\\nxdk_triangle

> reboot warm

> /launch e:\nxdk_triangle
> reboot warm

...
```

(You can also exit xbdm_gdb_bridge after reboots and still repro, in my case I always waited for the debug dashboard to be displayed after the reboot before relaunching the sample)

abaire:
Fixes #710 

Trivial program used to repro:
```

    if ((status = pb_init())) {
        debugPrint("pb_init Error %d\n", status);
        Sleep(2000);
        return 1;
    }

    pb_show_front_screen();

    pb_fill(0, 0, width, height, 0xff202020);
    pb_erase_text_screen();

    pb_print("Sleeping and rebooting...");
    pb_draw_text_screen();
    while(pb_busy()) {
        /* Wait for completion... */
    }

    /* Swap buffers (if we can) */
    while (pb_finished()) {
        /* Not ready to swap yet */
    }
    Sleep(500);
```

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -2551,6 +2551,8 @@ void pb_kill(void)
     while(pb_BackBufferbReady[pb_BackBufferNxt]);
 
     pb_running=0;
+    pb_uninstall_gpu_interrupt();
+    KeRemoveQueueDpc(&pb_DPCObject);
 
     if (pb_ExtraBuffersCount) MmFreeContiguousMemory((PVOID)pb_EXAddr[0]);
     if (pb_DepthStencilAddr) MmFreeContiguousMemory((PVOID)pb_DepthStencilAddr);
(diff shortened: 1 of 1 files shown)


abaire:
Also curious if `pb_running` should be `volatile` (though I confirmed that making it volatile is insufficient to fix this lockup without uninstalling the ISR handler)

PR #713, opened 2025-02-26
sdl: Update submodule to 5f6d52d

Ryzee119:
Fixes https://github.com/XboxDev/nxdk-sdl-old/issues/40

nxdk's SDL repository was forked from an unofficial mirror maintained on Github back when SDL was not on Github (officially)

Some years ago SDL officially migrated to Github and the SDL mirror was no longer maintained.

Unfortunately the official SDL repo had different git tree and so it was not possible for us to pull in upstream changes or rebase against upstream.

To fix this, I (painstakingly) cherry picked all of nxdk's SDL commits into the official SDL tree. I maintain author, commit author and timestamps etc so for all intents and purpose this is identical however it based off the official SDL git tree so commit hashes dont match.

This is the first step to bring up SDL to 2.3.x or SDL3

If merged, i think we should archive `XboxDev/nxdk-sdl-old`, and create a pinned issue in `XboxDev/nxdk-sdl` to reference old issues. There is one open PR but no activity in 5 years. This would no longer be mergable without a rebase.

Also we should change the default branch on XboxDev/nxdk-sdl to nxdk and enable issues on the fork repo


thrimbor:
I was about to hit the merge button, but I have one more (potentially super annoying) request. There is a merge commit in the history (that was done by me, oops), that's not only ugly but also can make later rebasing a huge headache, could you refactor that one out? That would involve cherry-picking faeb3d6 and 6878acc so they sit on top of fa1bd19, and then putting all the other commits on top again. I don't care about the authorship on the merge commit.
If you hit issues with this hit me up on Discord and I'll take over.


> Also we should change the default branch on XboxDev/nxdk-sdl to nxdk and enable issues on the fork repo

Done

> If merged, i think we should archive XboxDev/nxdk-sdl-old, and create a pinned issue in XboxDev/nxdk-sdl to reference old issues.

If you create the issue I'll archive the old repo when this is merged.

thrimbor:
Looks good, thank you! Merging now.

PR #715, opened 2025-03-01
cmake: Fix ninja depfile generation

abaire:
The cmake (3.30, at least) ninja buildsystem generation fails to create .d files when using the NXDK toolchain, leading to a lack of any dependencies on header files. This explicitly adds Clang generation commands.

Verified by building with `-d keepdepfile` and running `ninja -t deps` in the generated build dir.


thrimbor:
Ah, I remember hitting that issue a while back but didn't investigate. Tested your change it does indeed fix it, thank you!

PR #716, opened 2025-03-01
nvnetdrv v2 part 3

Ryzee119:
I think this will be my last major PR for this at this moment.

Primary change is significant optimizations to the RX ring. This drops the entire RX ring threading which in hindsight I'm not sure why I made it so complicated. This will reduce memory usage (atleast 2 or 3 pages) and improve performance.

Fixes a race during init because we setup hardware before lwip and if a packet arrived (via irq) before lwip was setup, it would try push it into the uninitialized network stack and could do strange things.

Updates to latest upstream LWIP. This fixes an actual issue we encountered with http downloads and timeouts not clearing correctly.

### DIFF
--- a/lib/net/nforceif/include/lwipopts.h
+++ b/lib/net/nforceif/include/lwipopts.h
@@ -178,7 +178,7 @@
  * a fragmented IP packet waits for all fragments to arrive. If not all fragments arrived
  * in this time, the whole packet is discarded.
  */
-#define IP_REASS_MAXAGE                 3
+#define IP_REASS_MAXAGE                 15
 
 /**
  * IP_REASS_MAX_PBUFS: Total maximum amount of pbufs waiting to be reassembled.
--- a/lib/net/nforceif/include/lwipopts.h
+++ b/lib/net/nforceif/include/lwipopts.h
@@ -83,7 +83,7 @@
  * ATTENTION: this does not work when tcpip_input() is called from
  * interrupt context!
  */
-#define LWIP_TCPIP_CORE_LOCKING_INPUT   1
+#define LWIP_TCPIP_CORE_LOCKING_INPUT   0
 
 /**
  * SYS_LIGHTWEIGHT_PROT==1: enable inter-task protection (and task-vs-interrupt
--- a/lib/net/nvnetdrv/nvnetdrv.c
+++ b/lib/net/nvnetdrv/nvnetdrv.c
@@ -30,13 +30,6 @@ struct tx_misc_t
     void *userdata;
 };
 
-// Struct to hold RX callback data
-struct rx_misc_t
-{
-    void *bufAddr;
-    size_t length;
-};
-
 #ifdef NVNETDRV_ENABLE_STATS
 struct nvnetdrv_stats_t
 {
@@ -86,7 +79,6 @@ static size_t g_txRingSize;
 static size_t g_rxRingHead;
 static size_t g_txRingHead;
 static atomic_size_t g_txPendingCount;
-static KSEMAPHORE g_rxPendingCount;
 static atomic_size_t g_rxRingTail;
 static atomic_size_t g_txRingTail;
 static KSEMAPHORE g_txRingFreeCount;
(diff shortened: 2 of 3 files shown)


thrimbor:
> This drops the entire RX ring threading which in hindsight I'm not sure why I made it so complicated.

I originally came up with that when I still tried to optimize pktdrv. Back then we had to poll for new packets, and I wanted to change that, but hit all kinds of issues, including random lockups.
Reading through my old discord conversations, it sounds like the core issue was memory allocation. We have our own pbuf pool for the rx buffers, so that's all good, but when passing the packet to the netif's input function we end up in `tcpip_inpkt`, which allocates a message and puts it into an mbox.
We have `MEMP_MEM_MALLOC` set (because preallocating pools for everything would be a waste when we already have `malloc`), but that means that the message allocation is also using `malloc`. If `malloc` doesn't have enough memory to satisfy the request, it calls `NtAllocateVirtualMemory`, and that's not allowed from a DPC because it can deadlock (I definitely hit that with my pktdrv work, iirc debug kernels even have an assert for that).

So I'd actually expect this PR to cause random hangs, but I tried hammering the http sample and didn't experience one.
I haven't yet looked into whether lwip had related changes done since I encountered these issues, but if it didn't (and I'm not mistaken in my conclusions) we might have to switch back to using pools (which sucks, but is probably better than the added complexity).

The other commits are fine, but I'll need to do some more reading and thinking (and then a proper review of the rx ring removal commit).

Ryzee119:
> The default mbox size it quite massive so we may not ever call the mbox allocator. But we probably don't want to rely on that because it may change in the future

We might just need to increase MEMP_NUM_TCPIP_MSG_INPKT variable to match or exceed rx ring size

Ryzee119:
Ive been digging into this more. 

I believe we need to do a few things:

1. `RX_BUFF_CNT` must be >= `MEMP_NUM_TCPIP_MSG_INPKT`
2. `RX_BUFF_CNT` must be able to encompass `TCP_WND`, which is set to `(45 * TCP_MSS)` in nxdk. So should be >= 45.
3. `TCPIP_MBOX_SIZE` should be some number greater than all of these. probably 2x or so. Currently it actually set to zero but our port ignores the size parameter and always uses 128 which should be okay.

Ill probably  make these changes:

Change `RX_BUFF_CNT ` to `#define RX_BUFF_CNT (TCP_WND / TCP_MSS)`.

Redefine `TCP_WND` from `(45 * TCP_MSS)` to `(MEMP_NUM_TCPIP_MSG_INPKT * TCP_MSS)`

Add #if and #error compile time sanity checks. 

`sys_mbox_trypost_fromisr` basically just calls `sys_mbox_trypost` in our port anyway. This does not block but will post a semapore which comes from our nvnet IRQ which is in a thread context. I think this is okay.

The `sys_mbox_trypost` call is guaranteed to succeed because `MEMP_NUM_TCPIP_MSG_INPKT`, `RX_BUFF_CNT` are all equal.

The pbuf allocator is guaranteed to succeed as we supply a custom PBUF pool that matches the `RX_BUFF_CNT`

I'll do some more testing, but I think this should actually help alot

thrimbor:
I'm not really concerned about the mbox (the way I see it, the worst it could do is drop packets when the mbox is full), but about the allocation of the messages. Because we have `MEMP_MEM_MALLOC` we don't use pools for that, so `MEMP_NUM_TCPIP_MSG_INPKT` has no effect, and we might hit a call to `NtAllocateVirtualMemory` from a DPC via malloc.
The way I see it, if we want to get rid of the rx thread, we need to disable `MEMP_MEM_MALLOC`, and live with the fact that we use pools even where we wouldn't need to - it doesn't look to me like we can use a pool for only this specific message type.

Ryzee119:
> `NtAllocateVirtualMemory` from a DPC via malloc

Its not directly in a DPC though. Our IRQ is handed in a thread triggered by the DPC. But yea gotcha memp_malloc uses libc malloc 🤔

thrimbor (approved):
Well, couldn't find anything wrong apart from a slightly misleading variable name that I've allowed myself to adjust (`paddr` -> `vaddr`).
Time to merge!

PR #718, opened 2025-03-07
CRLF check and cleanup

thrimbor:
This cleans up all CRLF line endings found in the nxdk repo itself.
We already had a `.editorconfig` file, but somehow a few still crept in or were present before. I converted all affected files to LF endings with `dos2unix`, no other changes were done to them.
I'm not sure what the consensus is on grouping these changes into a single commit - it's not really necessary to have them separated, and it'll be more verbose when we add it to `.git-blame-ignore-revs` later, but I couldn't think of a good prefix so I just kept them like this for now. If it's wished to squash them I'll happily do that.

This also adds another workflow for format checking. For now it only checks whether a commit in a PR is adding a CRLF line ending, but we can later expand that to checking actual code style as well (LLVM uses some clang-format integration for git that I want to take a look at).
Only added lines are checked (so removing any potentially overlooked CRLFs won't trigger it), and it checks each commit in the PR that is not part of `master` yet.

This is how it looks like:
image

/edit:
GH automatically inserts a fake merge commit (as can be seen in the screenshot) because the workflow is run from a merge branch it generates, I just force-pushed a small change that makes it use the actual HEAD of the PR instead.


Ryzee119:
Looks good. Happy to leave separate even if it means a few more lines in the blame ignore.

Few of these looks like I introduced them 🙈

mborgerson:
Can you add this change to .git-blame-ignore-revs

PR #719, opened 2025-03-07
hal: Handle LaunchData redirects to xonlinedash

abaire:
Adds support for values from https://xboxdevwiki.net/Kernel/LaunchDataPage#Type_0x00000001:_Switch_to_dashboard that are handled by xonlinedash.xbe

### DIFF
--- a/lib/hal/xbox.c
+++ b/lib/hal/xbox.c
@@ -3,7 +3,7 @@
 // SPDX-FileCopyrightText: 2004 Craig Edwards
 // SPDX-FileCopyrightText: 2006 Richard Osborne
 // SPDX-FileCopyrightText: 2017-2020 Stefan Schmidt
-// SPDX-FileCopyrightText: 2022 Erik Abair
+// SPDX-FileCopyrightText: 2022-2025 Erik Abair
 
 #include <string.h>
 // #include <xboxrt/stat.h>
@@ -17,6 +17,8 @@
 #define KernelMode 0
 
 #define LaunchDataPageSize 0x1000
+
+static const char *xonline_dash_path = "\\Device\\Harddisk0\\Partition2\\XODash\\xonlinedash.xbe";
 
 void XReboot()
 {
@@ -47,6 +49,31 @@ void XLaunchXBE(const char *xbePath)
     XLaunchXBEEx(xbePath, NULL);
 }
 
+static BOOL DoesFileExist(const char *lpPathName)
+{
+    ANSI_STRING path;
+    OBJECT_ATTRIBUTES objectAttributes;
+    IO_STATUS_BLOCK ioStatusBlock;
+
+    RtlInitAnsiString(&path, lpPathName);
+    InitializeObjectAttributes(&objectAttributes, &path, OBJ_CASE_INSENSITIVE, NULL, NULL);
+
+    HANDLE handle;
+    IO_STATUS_BLOCK statusBlock;
+    NTSTATUS status = NtOpenFile(&handle,
+                                 GENERIC_READ,
+                                 &objectAttributes,
+                                 &statusBlock,
+                                 FILE_SHARE_READ,
+                                 FILE_NON_DIRECTORY_FILE);
+    if (!NT_SUCCESS(status)) {
+        return FALSE;
+    }
+
+    NtClose(handle);
+    return TRUE;
+}
+
 void XLaunchXBEEx(const char *xbePath, const void *launchData)
 {
     if (LaunchDataPage == NULL) {
@@ -68,7 +95,16 @@ void XLaunchXBEEx(const char *xbePath, const void *launchData)
 
     if (!xbePath) {
         launchDataPage->Header.dwLaunchDataType = LDT_LAUNCH_DASHBOARD;
-    } else {
+
+        if (launchData && *(DWORD*)launchData >= LDT_LAUNCH_DASHBOARD_REASON_NETWORK) {
+            xbePath = xonline_dash_path;
+            if (!DoesFileExist(xbePath)) {
+                xbePath = NULL;
+            }
+        }
+    }
+
+    if (xbePath) {
         XConvertDOSFilenameToXBOX(xbePath, launchDataPage->Header.szLaunchPath);
 
         // one last thing... xbePath now looks like:
--- a/lib/xboxkrnl/xboxkrnl.h
+++ b/lib/xboxkrnl/xboxkrnl.h
@@ -3,7 +3,7 @@
 // SPDX-FileCopyrightText: 2017-2023 Stefan Schmidt
 // SPDX-FileCopyrightText: 2018-2021 Jannik Vogel
 // SPDX-FileCopyrightText: 2018 Sean Koppenhafer
-// SPDX-FileCopyrightText: 2022 Erik Abair
+// SPDX-FileCopyrightText: 2022-2025 Erik Abair
 
 /**
  * @file xboxkrnl.h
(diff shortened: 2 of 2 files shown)


lib/hal/xbox.c
```diff
@@ -18,6 +18,8 @@
 
 #define LaunchDataPageSize 0x1000
 
+static const char *xonline_dash_path = "\\Device\\Harddisk0\\partition2\\XODash\\xonlinedash.xbe";
```
GXTX, line 21: nit: p -> P

And is this path valid for all known online revisions?

abaire: This is the verbatim path from reversing a devkit sample XBE. 

If someone has a collection of all of the online revisions and it turns out that they use different paths they could be looped over until a working one is found (I don't have a list to use). 

The xbe also appeared to have handling to execute a dashboard upgrade or fallback to an `XODash` path on the DVD, neither of which seem appropriate for this method since an nxdk app would presumably never have an update to apply nor a bundled `xonlinedash.xbe` to defer to.

abaire: Capitalized the "P" and added a check to make sure xonlinedash exists. If it doesn't, this falls back to the current behavior of dropping into the dashboard.

Not sure if it'd be better to assert or just not patch up the xbe path and let the dashboard show an error since I believe the network settings are all expected to relaunch the title after completion.

LoveMHz:
Is ``XLaunchXBEEx`` the right place for this? This feels like it's trying to solve an issue that doesn't exist.

The kernel only cares if ``launch_data_type`` equals ``1`` to force load the dashboard.

All other contexts are up to the applications passing the launch data around (or dropping).

abaire:
> Is `XLaunchXBEEx` the right place for this? This feels like it's trying to solve an issue that doesn't exist.
> 
> The kernel only cares if `launch_data_type` equals `1` to force load the dashboard.
> 
> All other contexts are up to the applications passing the launch data around (or dropping).

My original intent for XLaunchXBEEx was to parallel the behavior of the XDK, where the redirect to `xonlinedash.xbe` happens behind the scenes. From the user perspective, asking to launch a NULL XBE indicates that you're passing control back to the dashboard (or "a dashboard" as it turns out), the caller is just responsible for setting the reason & associated params.

We could go the opposite route and force the caller to entirely specify all params, but then they'd need to be aware of the nuanced between reasons `< 6` and `>= 6`. I didn't know there was any difference until today, not sure if it's well documented someplace.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -3,7 +3,7 @@
 // SPDX-FileCopyrightText: 2017-2023 Stefan Schmidt
 // SPDX-FileCopyrightText: 2018-2021 Jannik Vogel
 // SPDX-FileCopyrightText: 2018 Sean Koppenhafer
-// SPDX-FileCopyrightText: 2022 Erik Abair
+// SPDX-FileCopyrightText: 2022-2025 Erik Abair
```
JayFoxRox, line 6: nit: I'm not sure if we also use separate years elsewhere or if mixing of ranges and separate is common (e.g: `x-y,z,u-v`), but *if* we do, I think `2022,2025` would be more faithful in this instance.

lib/hal/xbox.c
```diff
@@ -68,7 +95,17 @@ void XLaunchXBEEx(const char *xbePath, const void *launchData)
 
     if (!xbePath) {
         launchDataPage->Header.dwLaunchDataType = LDT_LAUNCH_DASHBOARD;
-    } else {
+
+        // Redirect to xonlinedash.xbe as needed.
```
JayFoxRox, line 99: nit: Start lowercase and avoid punctation at end of comment for consistency with surrounding code

lib/hal/xbox.c
```diff
@@ -68,7 +95,17 @@ void XLaunchXBEEx(const char *xbePath, const void *launchData)
...
+        if (launchData && *(int*)launchData >= LDT_LAUNCH_DASHBOARD_REASON_NETWORK) {
```
JayFoxRox, line 100: In `XGetLaunchInfo` we appear to use `*(unsigned long*)` for the `launchData` type field. From what I can tell, casting to `PLAUNCH_DATA_HEADER` and then accessing `dwLaunchDataType` would be even cleaner (`DWORD`).

lib/xboxkrnl/xboxkrnl.h
abaire: I'm fine just removing myself entirely, I would hope the copyright could be assigned to the project without the need to have individual names and worrying about assigning timestamps to contributions and whether they should be comma separated if intervening years had no commits, etc...

lib/hal/xbox.c
abaire: It's already inconsistent, the comment on line 85 is capitalized with punctuation. Removing the comment as it's obvious from context.

lib/hal/xbox.c
abaire: launchData is not a `PLAUNCH_DATA_HEADER`, it's the arbitrary user defined data after the header and padding. I cast to a DWORD, but this isn't a struct that's defined in the nxdk at all (but maybe it should be for this one magic use case?)

lib/xboxkrnl/xboxkrnl.h
thrimbor: The SPDX spec isn't very precise regarding copyright texts, so mixing is probably fine. Personally I don't really care whether it's 2022-2025 or 2022,2025 here, but please don't remove your name.

> copyright could be assigned to the project

Oh we don't do that. Giving away your copyright is unnecessary imho and doesn't even exist as a legal possibility in some countries, it's why some projects make you sign huge contributor license agreements that cover all kinds of strange cases. I stay as far away from such stuff as possible.

thrimbor:
@LoveMHz I thought this was strange too and didn't want to believe this was necessary, but it seems like it is and retail XBEs do it, too.

I looked at Halo 2, and its `XLaunchNewImageA` goes through an array of function pointers (most of which are zero, strange stuff) which contains `XoUpdateLaunchNewImageInternal`, which has the same `>= 6` check and depending on that modifies it to explicitly launch the online dash instead.
Very weird stuff, I would've expected the dash to just check for those values and forward them to the online dash if necessary, but I tried that and it doesn't seem to happen.
I wanted to look deeper into it, checking the contents of the launch data page when rebooting, but Arch has switched to SDL3 with the SDL2 compatibility layer, and somehow that causes xemu to hang on launch 90% of the time so I can't rn.

lib/xboxkrnl/xboxkrnl.h
RadWolfie: > nit: I'm not sure if we also use separate years elsewhere or if mixing of ranges and separate is common

https://github.com/search?q=repo%3AXboxDev%2Fnxdk+SPDX-FileCopyrightText&type=code

According to code search (see link above), they all contain dashes for an initial year to the "last time touched" current year. Although, it's kind of a moot point to update the "last time touched" current year when we have git blame to inspect who made the latest changes. General companies did drop the range of updated year anyway.

If it were up to me, I would just use the initial year only. That's my personal opinion.

lib/hal/xbox.c
RadWolfie: Actually, this will break any homebrew titles that may use `DWORD` size variable in the first offset of launchData. Reason? launchData is copy directly to launchDataPage's LaunchData offset. Therefore doesn't set any to launchDataPage's Header offset other than szLaunchPath member variable from the first parameter of XLaunchXBEEx function.

The parameter of launchData from XLaunchXBEEx function is intended to pass on directly without read from it. So, we shouldn't do this way.

lib/xboxkrnl/xboxkrnl.h
abaire: I'm happy with that last suggestion. I've restored the original state w/ just the original contribution date. 

It'd be really good to either clean out the dashes or provide some contributor instructions that indicate that people shouldn't bother extending ranges if they commit over years so it's clear what is expected.

lib/hal/xbox.c
abaire: This will only affect titles that use `XLaunchXBEEx` specifically to launch the dashboard by failing to pass an `xbePath`. At that point, I believe the structure of the launch data is (semi) well defined, since the target being launched sets the expectations for the content. This is why there's a `dwLaunchDataType` enumeration that calls out the fact that it's of "LAUNCH_DASHBOARD" type.

abaire:
> @LoveMHz I thought this was strange too and didn't want to believe this was necessary, but it seems like it is and retail XBEs do it, too.
> 
> I looked at Halo 2, and its `XLaunchNewImageA` goes through an array of function pointers (most of which are zero, strange stuff) which contains `XoUpdateLaunchNewImageInternal`, which has the same `>= 6` check and depending on that modifies it to explicitly launch the online dash instead. Very weird stuff, I would've expected the dash to just check for those values and forward them to the online dash if necessary, but I tried that and it doesn't seem to happen.

That's exactly consistent with what I saw. It was so unexpected that it took me longer than it should've to back my way up to that patch table and properly read the `> 5` check to realize that it was injecting an alternate XBE :)

lib/hal/xbox.c
RadWolfie: > And is this path valid for all known online revisions?

For the storage drive's path, yes. The capitalize P I had seen was from older online revisions. The latest doesn't have a capitalized letter since the file path is case-insensitive. Older online revisions had the folder named "XODASH" then later changed to "XODash". It doesn't matter to choose between the two as the kernel will be able to find the file.

I agree with abaire as there's no need to add a backup path to homebrew DVD containing `xonlinedash.xbe`.

---
NOTE: The reviewers not in XboxDev team are not able to close the review remark.

thrimbor:
I have squashed and rebased the changes and will merge as soon as the CI ran through.
I decided to keep the SPDX line as it was in your first commit, as we use ranges everywhere in the code and I've always interpreted it as min/max of the individual change dates.

PR #722, opened 2025-03-11
Remove nvnetdrv rx thread

thrimbor:
This reworks our lwIP arch abstraction code to remove the necessity to call it from a separate thread instead of a DPC, and then removes that thread from nvnetdrv.

There were three main issues that prevented this previously:
* The semaphore implementation used `NtWaitForSingleObject` which is not safe to call from a DPC, even if we can guarantee it wouldn't block
* The mbox implementation protects itself with a mutex, which will deadlock if it causes a call from a DPC to block
* lwIP used `malloc()` for everything, potentially calling `NtAllocateVirtualMemory` via `VirtualAlloc` from a DPC which is not safe

This PR fixes all three. The semaphores now use the kernel's implementation directly instead of the object layer abstraction version, the mbox now raises the IRQL instead of locking a mutex, and lwIP now gets its memory from the kernel's pool allocator instead (I got the idea from ReactOS though my implementation is not based on theirs).

While doing this I also changed the abstraction code to use in-place objects instead of pointers and malloc/free.

We don't commonly use commit messages, but I added one to the reimplementation commit because I felt it was beneficial to explain why the changes were done.

Bugs in this PR have the potentially to really break applications using the network, so any testing is highly appreciated.


lib/net/nforceif/src/sys_arch.c
```diff
@@ -91,209 +64,176 @@ sys_init(void)
...
-    return ts.tv_sec * 1000000000L + ts.tv_nsec;
-}
-#endif
-
-#if !NO_SYS
-
 sys_thread_t
 sys_thread_new(const char *name, lwip_thread_fn function, void *arg, int stacksize, int prio)
 {
     struct sys_thread *st = NULL;
```
ergo720, line 70: The variable `st` is no longer used and could be deleted.

lib/net/nforceif/src/sys_arch.c
```diff
@@ -91,209 +64,176 @@ sys_init(void)
...
-    if (!ext_sys_arch_sem_try(&mbox->write_sem)) {
+    if (!ext_sys_arch_sem_try(&mb->write_sem)) {
         return ERR_MEM;
     }
 
-    sys_arch_sem_wait(&mbox->mutex, 0);
-    mbox->msgs[mbox->last % SYS_MBOX_SIZE] = msg;
-    mbox->last++;
-    sys_sem_signal(&mbox->mutex);
+    KIRQL prev_irql = KeRaiseIrqlToSynchLevel();
```
ergo720, line 116: Is it necessary to raise the IRQL all the way up to sync level (=28), since that will block almost all other interrupts in the system. On the xbox, the NIC uses IRQ=4, which `HalGetInterruptVector` will convert to IRQL=22. Would that be enough here?

lib/net/nforceif/src/sys_arch.c
thrimbor: Removed.

lib/net/nforceif/src/sys_arch.c
thrimbor: Actually that was supposed to be `KeRaiseIrqlToDpcLevel`, I had increased it during debugging and apparently forgot to change it back.

MasonT8198: Happens to the best of us.

Ryzee119:
I ran these changes on my ftp server. I uploaded -> downloaded -> uploaded ... etc approx 1000 files 10 times. I then performed a CRC check against the original content to confirm no errors had occurred.
To stress test all the messaging mechanisms I used 9 concurrent clients. 
This resulted in 0 errors and a 100% CRC match. Im confident this change does not break anything.

I also ran the iperf lwip server sample with no issues.
```
Client connecting to 192.168.68.69, TCP port 5001
TCP window size: 64.0 KByte (default)
------------------------------------------------------------
[  1] local 192.168.68.68 port 58628 connected with 192.168.68.69 port 5001
[ ID] Interval       Transfer     Bandwidth
[  1] 0.00-10.03 sec   112 MBytes  93.5 Mbits/sec
```

My only nit would be to rename the commit `... remove rx thread` to `... remove irq thread` as this also handed the mii irq and tx irqs.

MasonT8198:
Did some testing myself too, works great, I think this can be merged with the changes that Ryzee suggested

thrimbor:
> My only nit would be to rename the commit ... remove rx thread to ... remove irq thread as this also handed the mii irq and tx irqs.

Done

PR #723, opened 2025-03-14
lib: Replace All Instances of RtlZeroMemory with Memset

MasonT8198:
Per discussion on the xbox dev discord

image


RadWolfie:
Things to consider:

1. Memset will be variety base on different compiler and flavor of optimization. Plus how much memory space it will use.
2. Memset's inline will cause file size increase vs a simple jump to the kernel's already made function. (memset inline method will use more RAM space.)
3. Performance wise, which one is more faster? Was there a proof of using memset is better?

Base on OP post, the above are missing the why this pull request should be accepted. I recommended do the comparison analysis first before make proposal change. In my opinion, performance wise is important. Some others may say having more free RAM space is important, which I can agree that 64MB of RAM is pretty small.

LoveMHz:
Interesting fact for the day... No retail software uses ``RtlZeroMemory``.

---

> * Memset will be variety base on different compiler and flavor of optimization. Plus how much memory space it will use.

By forcing a call into the kernel you're removing any chance of compiler optimizations.

> * Performance wise, which one is more faster? Was there a proof of using memset is better?

The only performance optimization in ``RtlZeroMemory`` is WORD aligned writes. The ``memset`` in ``pdclib`` has a lot of room for optimizations to take advantage of compile time optimizations provided by GCC/LLVM builtins (``__builtin_memset``) that could easily surpass ``RtlZeroMemory``. The same could be said about ``memcpy`` too.

> * Memset's inline will cause file size increase vs a simple jump to the kernel's already made function. (memset inline method will use more RAM space.)

Not quite simply a ``JMP``. Kernel exports are by reference so you would have two ``PUSH`` and an indirect ``CALL`` vs a ``STO REP``. But again, by using ``memset`` we would have more options.

So I would merge this in, as is, and consider creating another issue on memset/memcpy optimizations.

thrimbor:
> Interesting fact for the day... No retail software uses `RtlZeroMemory`.

Interesting detail. I would've assumed at least the driver parts do, but maybe the XDK actually provided it as a macro wrapper around `memset` instead like it seems to be done nowadays.

At the moment imitating that would be my preferred solution, that way we would keep the more clear expression of intent that `RtlZeroMemory` communicates, which is why I currently prefer it for the winapi/driver parts (that, and after 23 years of understanding C I still can't remember the parameter order of `memset` without looking it up).
I'm just not sure how we should handle the name conflict with the kernel export then - hide it behind some `#define` switch? If so, how should we name it? `XBOXKRNL_PREFER_EXPORTS`?

Overall using memset (at least internally) is better imho, so if my idea isn't liked I'd be fine with seeing this merged (after squashing and fixing up the commit title)

LoveMHz:
> Interesting detail. I would've assumed at least the driver parts do

Nope, not even there. It's one of the few parts of that kernel I consider dead.

PR #724, opened 2025-03-14
pbkit: Add perspective-correct texture mapping bit to control0

coldhex:
Bit 20 in NV097_SET_CONTROL0 enables perspective corrected texture mapping (and perspective corrected color interpolation D0, D1, B0 and B1). It is always enabled by DirectX.

Left image bit 20 unset, right image bit 20 set:

Quad using two triangles:
pers_bitri

Just quad:
pers_quad

(Images from Xbox HW and https://github.com/coldhex/nxdk_pgraph_tests/commit/aa83445899633cf2c2609850cd90f87d2422b8ba)


dracc:
LGTM

PR #729, opened 2025-04-17
build: Warn if clang version is in the 19.x.x-20.1.2 range

thrimbor:
There is a bug in these LLVM versions that can cause SSE2 instructions get emitted when optimizations are turned on despite the target not supporting SSE2.

Originally found by @LoveMHz, tracked in https://github.com/llvm/llvm-project/issues/134607 and fixed in https://github.com/llvm/llvm-project/pull/134547

Version 20.1.3 has the fix.

### DIFF
--- a/bin/activate
+++ b/bin/activate
@@ -2,7 +2,7 @@
 
 # SPDX-License-Identifier: CC0-1.0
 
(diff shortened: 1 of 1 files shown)


thrimbor:
Merging despite no reviews to avoid people building broken applications

PR #730, opened 2025-04-17
Revert "docker: Update to buildbase:git-ee8f698d (based on Alpine 3.21)"

thrimbor:
This reverts the docker image back to the Alpine 3.18-based base-image to avoid the compiler bug https://github.com/XboxDev/nxdk/pull/729 warns about.


thrimbor:
Merging despite no reviews to avoid people building broken applications

Issue #733, opened 2025-04-25
SDL submodule fails to build on MSYS2


astarivi:
SDL build fails on MSYS2 MINGW64:

```
[ CC       ] /d/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit.obj
[ CC       ] /d/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit_0.obj
[ CC       ] /d/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit_1.obj
[ CC       ] /d/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit_A.obj
D:/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit_A.c:343:15: error: always_inline function
      '_mm_setzero_si64' requires target feature 'sse2', but would be inlined into function
      'BlitRGBtoRGBPixelAlphaMMX' that is compiled without support for 'sse2'
  343 |     mm_zero = _mm_setzero_si64();       /* 0 -> mm_zero */
      |               ^
D:/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit_A.c:357:20: error: always_inline function
      '_mm_cvtsi32_si64' requires target feature 'sse2', but would be inlined into function
      'BlitRGBtoRGBPixelAlphaMMX' that is compiled without support for 'sse2'
  357 |             src1 = _mm_cvtsi32_si64(*srcp); /* src(ARGB) -> src1 (0000ARGB) */
      |                    ^
D:/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit_A.c:358:20: error: always_inline function
      '_mm_unpacklo_pi8' requires target feature 'sse2', but would be inlined into function
      'BlitRGBtoRGBPixelAlphaMMX' that is compiled without support for 'sse2'
  358 |             src1 = _mm_unpacklo_pi8(src1, mm_zero); /* 0A0R0G0B -> src1 */
      |                    ^
D:/xb_dev/nxdk2/nxdk/lib/sdl/SDL2/src/video/SDL_blit_A.c:360:20: error: always_inline function
      '_mm_cvtsi32_si64' requires target feature 'sse2', but would be inlined into function
      'BlitRGBtoRGBPixelAlphaMMX' that is compiled without support for 'sse2'
[...]

PR #734, opened 2025-04-27
build: Undefine __MMX__

thrimbor:
LLVM has dropped direct support of MMX intrinsics in version 20, instead implementing them as wrappers of SSE2 functionality (https://github.com/llvm/llvm-project/pull/96540).
As a non-SSE2 target, we now have no MMX support (which according to upstream was prone to issues anyway), so we need to lose the `__MMX__` define.

Fixes https://github.com/XboxDev/nxdk/issues/733

### DIFF
--- a/bin/nxdk-cc
+++ b/bin/nxdk-cc
@@ -21,4 +21,5 @@ clang \
     -DNXDK \
     -D__STDC__=1 \
     -U__STDC_NO_THREADS__ \
+    -U__MMX__ \
     "$@"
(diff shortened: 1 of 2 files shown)


Ryzee119:
We can still use MMX assembly as I understand? https://github.com/llvm/llvm-project/commit/b79568654e38a14ef921af932ed96abd8961b1ed

thrimbor:
Yeah (inline) assembly is still fine, but with MMX you'll have to be extra careful to not have your inline asm blocks reordered and always make sure to have an `emms` instruction.

Considering the upstream issue discussion, I'll probably end up closing this and submitting a PR to the SDL submodule instead.

PR #741, opened 2025-06-21
nvnetdrv: Implement link state change handling

Ryzee119:
Previously it was assumed the link state was up at all times.

This implements link state handling. You can now connect the ethernet cable anytime (or unplug) and lwip should handle things better.

### DIFF
--- a/lib/net/nvnetdrv/nvnetdrv_lwip.c
+++ b/lib/net/nvnetdrv/nvnetdrv_lwip.c
@@ -14,6 +14,7 @@
 #include "lwip/snmp.h"
 #include "lwip/stats.h"
 #include "lwip/sys.h"
+#include "lwip/tcpip.h"
 #include "netif/etharp.h"
 #include "netif/ppp/pppoe.h"
 #include "nvnetdrv.h"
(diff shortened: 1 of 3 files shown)


lib/net/nvnetdrv/nvnetdrv_lwip.c
```diff
@@ -123,6 +134,13 @@ static err_t low_level_init (struct netif *netif)
     /* set MAC hardware address */
     memcpy(netif->hwaddr, nvnetdrv_get_ethernet_addr(), 6);
 
+    // Let LWIP know the NIC is initialized and ready to use
+    netif_set_up(g_pnetif);
+
+    // Let LWIP know if a ethernet cable is currently connected
+    bool link_up = nvnetdrv_is_link_up();
+    change_link_state((void *)link_up);
```
JayFoxRox, line 142: `change_link_state` does `netif_set_up` internally (if a cable is connected); why is the `netif_set_up` before that necessary?
Is it so the NIC is detected at all in case the cable isn't connected at startup?

The comments seem to suggest this but I'm not sure I fully understand.

I'd guess that lwip might behave differently if there's no NIC, so this is probably done to ensure it sees one.. but that also doesn't make sense to me because the current code in master never seemed to call `netif_set_up` here in the first place - so it should have never worked?

JayFoxRox, line 142: Should (or can?) this happen before `netif_set_up`?

If the link state defaults to `up` and no cable is connected, lwip would first set up the NIC with default link state `up`, and then immediately change the link state `down` again.

mborgerson:
Nice, thanks!

PR #742, opened 2025-06-25
cmake: Auto discover NXDK_DIR

abaire:
Simplifies creation of projects in CLion


RadWolfie:
@abaire, have you tested this without NXDK_DIR environment variable set? I am a bit concerned it may not work accurately as it could mix up between installed and portable nxdk folder. Plus the fact that environment variable isn't set at all will cause nxdk's compiler to fail.

RadWolfie:
No one wants to mix and match parts of two nxdk folders. 🥲 Even though I do see the mistake occur in upstream's cmake will redirect to environment variable setter's nxdk folder. However this change from my view could cause even more mix up of using two nxdk folders. Since you have tested without environment variable and able to build an xbe successfully. The concern can be dismissed for this part.

One more question, did you test any of linkage library inside [pkgconfig](https://github.com/XboxDev/nxdk/tree/master/lib/pkgconfig) folder? Since they use environment variable too.

> \# Keep in sync with include paths in bin/nxdk-cc 
> ...

For whatever reason with IDE(s?, in my case, visual code) always show either blue or red underline names or unable to show list of possible matched names.

> Is there a reason someone would want to mix and match parts of two nxdk installs rather than just use the entire nxdk bundle?

As for your question, some of the people in the community do use nxdk as a submodule ([which can be found on GitHub repos](https://github.com/search?q=path%3A.gitmodules+nxdk&type=code)) for some odd reason. So, having 2+ nxdk folders for 2+ projects could have one of two folders being used as NXDK_DIR environment set. Plus the user may have the main nxdk installed as well (as NXDK_DIR environment set too). I can see these as mostly user errors. Yet a valid reason for such a thing to exist is experimental forked nxdk to verify improvement/feature works in their fork before create a pull request. One other possible reason is to choose a specific commit that is stable for their project(s). **What I'm trying to say is having two separate nxdk folders is intended to verify the code works from their respective nxdk folder rather than some files accidentally included from another nxdk folder unintentionally.**

I apologize for the rambling I may have made above. I tried to keep it short and simple. Yet it got longer and complex than I expected.

abaire:
> As for winmm library, as far as I remember on Windows, it should be done by the user. Either by `#pragma comment(lib, "winmm.lib")` in the source code or `target_link_libraries(<your_project_name or ${PROJECT_NAME}> winmm.lib)` call in CMake. So, it might be fine for a temporary workaround if unable to perform the two methods that I am familiar with until a permanent fix is made.

I don't think pkg-config does anything magic so it'd still require user action to wire it in. The pc file just allows you to glob nxdk features together in your CMakeLists if desired, you could also explicitly depend on `winmm.lib` with the appropriate linker path settings.

> I would not necessarily agree with "tests" stuff residing in the .github folder as there's a possibility contributor could run the test project(s) locally to be sure nothing is broken. The samples folder is in the root directory and the tests folder should be in the root directory to be easily recognized we do have sample/test projects that can be vetted for no breakage within nxdk tools.

I'd argue that samples are examples of how the nxdk is meant to be used whereas tests are there to validate changes to the project itself. In particular I would not want somebody to look at the hacky test target and think that that is the best way to set up a CMake-based project using the nxdk. I think the bar to finding that the tests exist and how to run them is pretty low for anybody doing a PR (if you break the test I believe the failure will provide sufficient info to discover and run it locally).

I'll leave it for the maintainers to chime in on their preference of location. I'll hold off on changing the workflow to include the tests until it's agreed that these tests are desirable at all.

> In my opinion, this pull request is no longer `cmake: Auto discover NXDK_DIR ` modification. Perhaps a change to the PR title or split them into separate pull requests. As this pull request is pretty much QoL improvements and added the missing stuff that wasn't there before.

Yeah, it's expanded a bit. I'll again let a maintainer decide if it's worth renaming or decomposing the commits further.

Thanks for your feedback, I had missed that the package config stuff existed and wouldn't have found it if you hadn't asked :)

PR #743, opened 2025-06-27
Added XID driver support for the Steel Battalion Controller

Quantx:
This will eventually be utilized by this PR: https://github.com/XboxDev/nxdk-sdl/pull/10

### DIFF
--- a/lib/winapi/timezoneapi.c
+++ b/lib/winapi/timezoneapi.c
@@ -242,3 +242,30 @@ BOOL FileTimeToSystemTime (const FILETIME *lpFileTime, LPSYSTEMTIME lpSystemTime
 
     return TRUE;
 }
+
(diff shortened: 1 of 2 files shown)


lib/usb/libusbohci_xbox/xid_driver.h
```diff
@@ -36,7 +36,24 @@ typedef struct __attribute__((packed))
     uint16_t wAlternateProductIds[4];
 } xid_descriptor;
 
-typedef struct __attribute__((packed)) 
+//XINPUT defines and struct format from
+//https://docs.microsoft.com/en-us/windows/win32/api/xinput/ns-xinput-xinput_gamepad
```
JayFoxRox, line 40: These docs are probably intended for Xbox 360 and up (when MS publicly documented XInput and provided Windows drivers).
Original Xbox doesn't even have "LEFT_SHOULDER" and "RIGHT_SHOULDER" - it uses black and white face-buttons.

PR #745, opened 2025-07-09
video: Disable soften filter on start-up & fix limits for flicker filter

GXTX:
D3D init:
```
          D3DDevice_SetFlickerFilter(5);
          D3DDevice_SetSoftDisplayFilter(0);
```

FilterFilter:
```
LAB_8002fda6:
      _smbus_write(RegisterBase,bVar5,bVar4);
      return;
    }
    if (DAT_8003b3e4 != 0xe0) {
      if (Param == 0) {
        bVar5 = 200;
        HalReadSMBusValue(DAT_8003b3e4,200,'\0',&Param);
        bVar4 = (byte)Param | 0x40;
      }
      else {
        if (Param == 1) {
          bVar5 = 1;
        }
        else if (Param == 2) {
          bVar5 = 2;
        }
        else if (Param == 3) {
          bVar5 = 3;
        }
        else {
          bVar5 = 0;
        }
        HalReadSMBusValue(DAT_8003b3e4,200,'\0',(ULONG *)&Result);
        _smbus_write(RegisterBase,200,(byte)Result & 0x80 | bVar5 << 3 | bVar5);
        if (Param == 5) {
          bVar4 = 0x80;
        }
        else {
          bVar4 = 0;
        }
        bVar5 = 0x34;
      }
      goto LAB_8002fda6;
    }
```

### DIFF
--- a/lib/hal/video.c
+++ b/lib/hal/video.c
@@ -54,7 +54,7 @@ static DWORD			dwEncoderSettings 	= 0;
 static VIDEO_MODE		vmCurrent = { 0, 0, 0, 0 };
 static int			flickerLevel		= 5;
 static BOOL			flickerSet		= FALSE;
-static BOOL			softenFilter		= TRUE;
+static BOOL			softenFilter		= FALSE;
 static BOOL			softenSet		= FALSE;
 static GAMMA_RAMP_ENTRY	gammaRampEntries[256];
 
@@ -381,7 +381,7 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 	_fb = framebufferMemory;
 
 	XVideoSetFlickerFilter(5);
-	XVideoSetSoftenFilter(TRUE);
+	XVideoSetSoftenFilter(FALSE);
 
 	for(i = 0; i < 256; i++) {
 		defaultGammaRampEntries[i].red = i;


lib/hal/video.c
```diff
@@ -428,12 +428,12 @@ void XVideoSetVideoEnable(BOOL enable)
 
 void XVideoSetFlickerFilter(int level)
 {
-	if(!flickerSet || (level & 0x07) != flickerLevel)
+	if(!flickerSet || (level & 0x05) != flickerLevel)
```
Ryzee119, line 431: If this mask is intended to clamp it to be <=5, this isn't right?
This will mask out 2 and 3 aswell.

What's the intent of this?

lib/hal/video.c
```diff
@@ -381,7 +381,7 @@ void XVideoInit(DWORD dwMode, int width, int height, int bpp)
 	_fb = framebufferMemory;
 
 	XVideoSetFlickerFilter(5);
-	XVideoSetSoftenFilter(TRUE);
+	XVideoSetSoftenFilter(FALSE);
```
Ryzee119, line 384: Why did you change it? I assume you found that retail titles do this?

lib/hal/video.c
GXTX: 👍reverted, any non-valid value will set the max value.

lib/hal/video.c
Ryzee119: okay, I  booted Amped 1 and can confirm this is the case in traffic to the encoder.

Ryzee119:
@GXTX can you please reword your commit title to remove the bit about the fix limits and I'll merge

PR #747, opened 2025-07-20
Implement `CompareFileTime`, `FileTimeToSystemTime`, and `SystemTimeToFileTime`

thrimbor:
This implements `CompareFileTime`, `FileTimeToSystemTime`, and `SystemTimeToFileTime`. `CompareFileTime` only has an assert to check for null pointer parameters because it has no way to signal an error to the caller.

Also fixes a typo in the `TIME_FIELDS` struct definition.


Ryzee119:
amazing thanks!

PR #751, opened 2025-07-29
libc: Partial implementaion of `direct.h`

Ryzee119:
Supersedes https://github.com/XboxDev/nxdk/pull/636

The cwd functions remain stubbed with asserts, however I fixed this comment by JFR https://github.com/XboxDev/nxdk/pull/636#discussion_r1233221554 and return `ENOSYS` for these functions which means "Function not implemented"

Otherwise I just squashed and renamed the commit title, and did some minor formatting cleanup

### DIFF
--- a/lib/xboxrt/libc_extensions/direct.c
+++ b/lib/xboxrt/libc_extensions/direct.c
@@ -1,0 +1,131 @@
+// SPDX-License-Identifier: MIT
+
+// SPDX-FileCopyrightText: 2023 ExoSkye
+// SPDX-FileCopyrightText: 2025 Ryan Wendland
+
+// Part of Microsoft CRT
+
+#include <assert.h>
+#include <errno.h>
+#include <hal/debug.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <windows.h>
+
+// Made referencing https://www.digitalmars.com/rtl/direct.html, retrieved 2023-06-10, copyrighted 1999-2018 by Digital Mars
+
+// Only call this function in the case there actually is an error
+static int win32_error_to_errno (DWORD winerror)
+{
+    switch (winerror) {
+        case ERROR_FILE_NOT_FOUND:
+        case ERROR_PATH_NOT_FOUND:
+        case ERROR_FILENAME_EXCED_RANGE:
+        case ERROR_BAD_PATHNAME:
+            return ENOENT;
+        case ERROR_ALREADY_EXISTS:
+        case ERROR_FILE_EXISTS:
+            return EEXIST;
+        case ERROR_NOT_LOCKED:
+        case ERROR_ACCESS_DENIED:
+            return EACCES;
+        case ERROR_INVALID_PARAMETER:
+        case ERROR_INVALID_FUNCTION:
+            return EINVAL;
+        case ERROR_NOT_ENOUGH_MEMORY:
+            return ENOMEM;
+        case ERROR_DIR_NOT_EMPTY:
+            return ENOTEMPTY;
+        case ERROR_CALL_NOT_IMPLEMENTED:
+            return ENOSYS;
+        default:
+            assert(0 && "Unknown error value in win32_error_to_errno");
+            return EINVAL;
+    }
+}
+
+int _mkdir (const char *pathname)
+{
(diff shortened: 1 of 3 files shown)


JayFoxRox:
Typo in PR and commit title ("implementaion").

`ERROR_CALL_NOT_IMPLEMENTED` also appears to be intended for different errors in Windows? I'm fine with it as a temporary hack because there probably isn't a suitable error-code.

However, I wonder if the getters should be temporarily hardcoded to return `D:\` (which is typically where the current application is mounted) to avoid the situation altogether.
Even `_chdrive` could simply succeed if it's changing to the current drive index.

We'd only be lacking `_chdir`,  but misbehaviour of applications which depend on these functions is much less likely.
I believe many applications will simply check for the current drive and directory to find their data folder, so hardcoding `D:\` seems very reasonable for original Xbox.

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,135 @@
...
+        case ERROR_INVALID_FUNCTION:
+            return EINVAL;
+        case ERROR_NOT_ENOUGH_MEMORY:
+            return ENOMEM;
+        case ERROR_DIR_NOT_EMPTY:
+            return ENOTEMPTY;
+        case ERROR_CALL_NOT_IMPLEMENTED:
+            return ENOSYS;
+        default:
+            debugPrint("WARNING: Unknown win32 error code 0x%08lx, returning EINVAL", winerror);
```
thrimbor, line 43: I'd prefer if this was an assert instead, depending on the application this debugPrint might not go anywhere

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,135 @@
...
+}
+
+/*
+ * TODO: Make a working directory system?
+ * nxdk mounts the running xbe directory at D:\, so we will return that as a hardcoded current working directory and drive.
+ */
+
+int _chdir (char *path)
+{
+    if (strcmp(path, "D:\\") == 0 || strcmp(path, "D:") == 0) {
```
thrimbor, line 79: better use `_stricmp` here as lowercase is also valid

lib/xboxrt/libc_extensions/direct.h
```diff
@@ -3,3 +3,11 @@
 // SPDX-FileCopyrightText: 2020 Jannik Vogel
 
 // Part of Microsoft CRT
+
+int _chdir(char* path);
```
thrimbor, line 7: The parameter should be `const char *path` (applies to the function definition as well)

lib/xboxrt/libc_extensions/direct.h
```diff
@@ -3,3 +3,11 @@
...
+int _chdrive(int drive);
+char* _getcwd(char* buffer, size_t length);
```
thrimbor, line 9: nit: We don't have a .clang-format file applying to these files yet, but maybe give this file a quick re-format to match it to direct.c which seems to already follow the nxdk convention

lib/xboxrt/libc_extensions/direct.c
```diff
@@ -0,0 +1,135 @@
...
+            debugPrint("WARNING: Unknown win32 error code 0x%08lx, returning EINVAL", winerror);
+            return EINVAL;
+    }
+}
+
+int _mkdir (const char *pathname)
+{
+    BOOL result = CreateDirectoryA(pathname, NULL);
+
+    if (result == false) {
```
thrimbor, line 52: nit: Simplify to `!result`, or even `!CreateDirectoryA...`

thrimbor (commented):
Not seeing much wrong here, but left a bit of feedback.

There's also an improvement I'd suggest. The code has multiple places where unimplemented functionality is signaled via `assert(0)`. There's a neat trick (that I don't think we used in nxdk before) that makes the error message more useful when that is triggered, and that is to use a `logical and` with a string constant, something like this:
```c
assert(0 && "Unknown error value in win32_error_to_errno");
```
The line numbers etc. from an assert are usually enough so this is just a suggestion, but it can help narrow down an issue especially when line numbers have shifted since the last build.

Ryzee119:
Thanks @thrimbor  good pick ups. I think i have resolved these issues

PR #757, opened 2025-08-20
pbkit: deduplicate defines

pionere:
- get rid of defines in pbkit.c which are defined in pblit_dma.h (included via pbkit.h)


JayFoxRox:
nit: Commit name / description isn't exactly our style

PR #766, opened 2026-02-22
Implement LocalFileTimeToFileTime

antonic901:
This should be reverse of FileTimeToLocalFileTime, so here is imlpementation.

### DIFF
--- a/lib/winapi/timezoneapi.c
+++ b/lib/winapi/timezoneapi.c
@@ -296,3 +296,29 @@ BOOL FileTimeToLocalFileTime (const FILETIME *lpFileTime, LPFILETIME lpLocalFile
     lpLocalFileTime->dwHighDateTime = fileTime.HighPart;
     return TRUE;
 }
+
+BOOL LocalFileTimeToFileTime (const FILETIME *lpLocalFileTime, LPFILETIME lpFileTime)
+{
+    if (!lpLocalFileTime || !lpFileTime) {
+        SetLastError(ERROR_INVALID_PARAMETER);
+        return FALSE;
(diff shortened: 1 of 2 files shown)


lib/winapi/timezoneapi.c
```diff
@@ -296,3 +296,30 @@ BOOL FileTimeToLocalFileTime (const FILETIME *lpFileTime, LPFILETIME lpLocalFile
...
+    // Get the timezone offset bias in 100-nanosecond intervals
+    LARGE_INTEGER offset;
+    offset.QuadPart = timeZoneInformation.Bias;
+    offset.QuadPart *= 60LL * 10000000LL;
+
+    LARGE_INTEGER localFileTime;
+    localFileTime.LowPart = lpLocalFileTime->dwLowDateTime;
+    localFileTime.HighPart = lpLocalFileTime->dwHighDateTime;
+
+    // Adjust the file time by the timezone offset. This function does not account for DST.
```
Ryzee119: This comment doesnt apply to LocalFileTimeToFileTime 

See MSDocs https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-localfiletimetofiletime
`LocalFileTimeToFileTime uses the current settings for the time zone and daylight saving time. Therefore, if it is daylight saving time, this function will take daylight saving time into account, even if the time you are converting is in standard time.`

Ryzee119:
Generally our commit name convention is `winapi: Reason` can you change the commit title?

antonic901:
Update commit with proper message.

Removed that comment.

PR #769, opened 2026-05-01
Handle USB IRQ in DPC instead of ISR

thrimbor:
Moves USB IRQ handling from an ISR to a DPC in preparation of future changes to input handling.
Requires merging https://github.com/XboxDev/libusbohci/pull/3 first.


dracc:
Haven't tested, looks correct.

dracc (approved):

Ryzee119:
Thanks! long overdue. As I re-based the libusbohci commit it created a new commit hash so i took the liberty of fixing that here.
I think we should replace the USB  stack some day. Better options exist nowadays but this is a good fix for the moment

PR #770, opened 2026-05-20
Xbox kernel header improvements

PatrickvL:
This pull request is the result of https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/112 which fleshes out unit tests for all Xbox kernel API's, during which a few nxdk-related issues where discovered and solved via this PR;

This PR makes several updates and additions to the Xbox kernel header files to improve NTSTATUS code coverage, kernel object definitions, IRP major function handling, and function signatures for better type safety and clarity. The changes also introduce new type definitions and registry value constants, and correct some function prototypes.

I'm creating this as a draft, so reserve the liberty to rebase, squash, merge or otherwise modify the commit history. But still, each change has been committed separately, to facilitate cherry-picking if so desired, or when this PR as a whole needs discussion / adjustment / more work.

The most important changes are:

### NTSTATUS codes and kernel object constants

* Added several missing NTSTATUS codes, such as `STATUS_INVALID_DEVICE_REQUEST`, `STATUS_SHARING_VIOLATION`, `STATUS_SUSPEND_COUNT_EXCEEDED`, and `STATUS_CANCELLED` to `ntstatus.h`. [[1]](diffhunk://#diff-028c9bc3fcc55117c92153df8f6237eb6197314846586aad26a8672cc17456a2R35) [[2]](diffhunk://#diff-028c9bc3fcc55117c92153df8f6237eb6197314846586aad26a8672cc17456a2R49-R52) [[3]](diffhunk://#diff-028c9bc3fcc55117c92153df8f6237eb6197314846586aad26a8672cc17456a2R81)
* Defined new IRQL levels (`PROFILE_LEVEL`, `HIGH_LEVEL`) and added a `KOBJECTS` enum for kernel object types in `xboxkrnl.h`. [[1]](diffhunk://#diff-d21981aa83acac905ed7e8aa78402c9979eeecb06a6b94de94a91db9d2b05184R74-R75) [[2]](diffhunk://#diff-d21981aa83acac905ed7e8aa78402c9979eeecb06a6b94de94a91db9d2b05184R724-R734)

### Type and macro additions

* Introduced `VA_LIST` typedef for variadic function handling, and added `SLIST_ENTRY` as an alias for `SINGLE_LIST_ENTRY`. [[1]](diffhunk://#diff-f6daba5a213603772c00bd10b247134dc7b1ba9d25359370cc89892c2a6e1e22R51-R52) [[2]](diffhunk://#diff-d21981aa83acac905ed7e8aa78402c9979eeecb06a6b94de94a91db9d2b05184R1098-R1099)
* Added registry value type constants (e.g., `REG_SZ`, `REG_DWORD`) for use with registry-related APIs.

### IRP and driver object improvements

* Defined IRP major function codes (e.g., `IRP_MJ_CREATE`, `IRP_MJ_CLOSE`) and updated the `DRIVER_OBJECT` structure to use these codes for the `MajorFunction` array size.

### Function signature and prototype corrections

* Changed several function signatures to use more accurate types and return values, such as updating `RtlVsprintf`, `RtlVsnprintf`, `RtlSprintf`, and `RtlSnprintf` to return `INT` instead of `VOID`, and using `VA_LIST` for variadic arguments. [[1]](diffhunk://#diff-d21981aa83acac905ed7e8aa78402c9979eeecb06a6b94de94a91db9d2b05184L1856-R1892) [[2]](diffhunk://#diff-d21981aa83acac905ed7e8aa78402c9979eeecb06a6b94de94a91db9d2b05184L1988-R2019)
* Updated the return type of `NtDeleteFile` from `BOOLEAN` to `NTSTATUS`, and changed the return type of `KeIsExecutingDpc` from `BOOLEAN` to `ULONG`. [[1]](diffhunk://#diff-d21981aa83acac905ed7e8aa78402c9979eeecb06a6b94de94a91db9d2b05184L3053-R3079) [[2]](diffhunk://#diff-d21981aa83acac905ed7e8aa78402c9979eeecb06a6b94de94a91db9d2b05184L3706-R3732)

### Miscellaneous corrections

* Fixed the parameter count for `ExInterlockedAddLargeInteger` in the module definition file.
* Added the `DUPLICATE_CLOSE_SOURCE` flag for `NtDuplicateObject`.
* Fixed a parameter type in `KeRemoveEntryDeviceQueue`.
* Corrected the calling convention for `RtlUnicodeStringToInteger`.


lib/xboxkrnl/xboxkrnl.h
```diff
@@ -1237,12 +1252,21 @@ typedef NTSTATUS (NTAPI *PDRIVER_DISPATCH) (
...
 );
 
+#define IRP_MJ_CREATE 0x00
+#define IRP_MJ_CLOSE 0x01
+#define IRP_MJ_READ 0x02
+#define IRP_MJ_WRITE 0x03
+#define IRP_MJ_FLUSH_BUFFERS 0x06
+#define IRP_MJ_DEVICE_CONTROL 0x0A
+#define IRP_MJ_INTERNAL_DEVICE_CONTROL 0x0B
+#define IRP_MJ_MAXIMUM_FUNCTION 0x0D
```
JayFoxRox, line 1262: Are these the correct ones for Xbox? I believe they differ between Xbox and common Windows.
*(Edit: Maybe we should also have a code-comment about the fact that these have a different value in the original Xbox kernel)*

LoveMHz: Image

Based on ``FatxDriverObject`` these should be:

```c
#define IRP_MJ_CREATE 0x00
#define IRP_MJ_CLOSE 0x01
#define IRP_MJ_READ 0x02
#define IRP_MJ_WRITE 0x03
#define IRP_MJ_QUERY_INFORMATION 0x04
#define IRP_MJ_SET_INFORMATION 0x05
#define IRP_MJ_FLUSH_BUFFERS 0x06
#define IRP_MJ_QUERY_VOLUME_INFORMATION 0x07
#define IRP_MJ_DIRECTORY_CONTROL 0x08
#define IRP_MJ_FILE_SYSTEM_CONTROL 0x09
#define IRP_MJ_DEVICE_CONTROL 0x0A
#define IRP_MJ_??? 0x0B
#define IRP_MJ_??? 0x0C
#define IRP_MJ_CLEANUP 0x0D
#define IRP_MJ_MAXIMUM_FUNCTION 0x0D
```

---

Image

Based on ``IoBuildDeviceIoControlRequest`` it's safe to assume 0x0B is ``IRP_MJ_INTERNAL_DEVICE_CONTROL``

---

Image

The only reference I could find to 0x0C is in ``IoBuildAsynchronousFsdRequest``. Based on context I would say it's safe to assume that 0x0C is ``IRP_MJ_SHUTDOWN``.

---

Final:

```c
#define IRP_MJ_CREATE 0x00
#define IRP_MJ_CLOSE 0x01
#define IRP_MJ_READ 0x02
#define IRP_MJ_WRITE 0x03
#define IRP_MJ_QUERY_INFORMATION 0x04
#define IRP_MJ_SET_INFORMATION 0x05
#define IRP_MJ_FLUSH_BUFFERS 0x06
#define IRP_MJ_QUERY_VOLUME_INFORMATION 0x07
#define IRP_MJ_DIRECTORY_CONTROL 0x08
#define IRP_MJ_FILE_SYSTEM_CONTROL 0x09
#define IRP_MJ_DEVICE_CONTROL 0x0A
#define IRP_MJ_INTERNAL_DEVICE_CONTROL 0x0B
#define IRP_MJ_SHUTDOWN 0x0C
#define IRP_MJ_CLEANUP 0x0D
#define IRP_MJ_MAXIMUM_FUNCTION 0x0D
```

PatrickvL: Thanks for this detailed feedback!

Shall I amend the commit these symbols reside in, or push a fix-up commit? (I would prefer to amend, if you don't mind, but will wait a while for guidance.)

RadWolfie: First of all, thanks for the depth details on IRP_MJ_ defines @LoveMHz. For sure it is more than what I had documented. There's one question I have to ask, is the `FatxDriverObject` info came from your own documentation you had discovered? I'm just wondering.

LoveMHz: @RadWolfie yes, it's from my research using recovered cv meta from 5849 (and older kernels) using [ghidra-coff-process-cv](https://github.com/MakeMHz/ghidra-coff-process-cv).

RadWolfie:
image

nit: I recently notice the commit's title and body with "at" symbol usage cause link to user's profile by mistake. I recommend to drop the "at" symbol and perhaps use the quote or another method.

lib/xboxkrnl/xboxkrnl.h
thrimbor: > Shall I amend the commit these symbols reside in, or push a fix-up commit? (I would prefer to amend, if you don't mind, but will wait a while for guidance.)

Definitely amend. I'd even say that if the commits don't conflict, I'd prefer this PR be split into a few smaller ones that only contain related commits (like the fixes for the printf-style functions). I've only looked at a few commits so far and they look reasonable, but I'd like to check more closely and that would prevent the discussions from potentially becoming unwieldy.
Please also adjust the commit messages, like starting the title with `xboxkrnl: ` and keeping them short enough so they don't get cut off in GH, avoid using the @ sign or wrap it in backticks to prevent GH turning it into a tag. We don't often use commit messages that consist of more than a commit title, but I'm not against it if it provides useful detail or context.

lib/xboxkrnl/xboxkrnl.h
```diff
@@ -4240,6 +4240,16 @@ XBAPI PFN_COUNT NTAPI FscGetCacheSize (void);
...
 
+/* Registry value types */
+#define REG_NONE 0
+#define REG_SZ 1
+#define REG_EXPAND_SZ 2
+#define REG_BINARY 3
+#define REG_DWORD 4
+#define REG_DWORD_LITTLE_ENDIAN 4
+#define REG_DWORD_BIG_ENDIAN 5
+#define REG_MULTI_SZ 7
```
thrimbor, line 4251: I'm not aware of these being used with the kernel API, is there value in nxdk providing these, considering there is no registry?

ergo720: To my knowledge, `REG_BINARY` and `REG_DWORD` are the only ones used in `ExQueryNonVolatileSetting` to specify the type of the queried value. See [here](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded-legacy/blob/96aabe72e238674a22a61695fb6f5295259edf62/src/core/kernel/exports/EmuKrnlEx.cpp#L420) and [here](https://github.com/Cxbx-Reloaded/Cxbx-Reloaded-legacy/blob/96aabe72e238674a22a61695fb6f5295259edf62/src/common/EmuEEPROM.h#L48) for how they are used in cxbxr

PR #680, opened 2024-10-19
feat(winapi): add winsock2

ReenigneArcher:
This PR aims to add `winsock2` from the `winapi`, which is a requirement for using `enet`. (https://github.com/XboxDev/nxdk/issues/453)

This is basically from https://github.com/JayFoxRox/nxdk/pull/84, but I had to adjust the flags in the Makefile in order to find `#include <pktdrv.h>` in `winsock.c`. I've added Jay as a co-author so they get the proper credit.

Note: I was not 100% successful in building enet using this yet. I am facing the following error.

```
-- Configuring done (0.7s)
-- Generating done (0.1s)
-- Build files have been written to: /c/Users/ReenigneArcher/Documents/GitHub/LizardByte/mnxb/build
[  2%] Building C object third-party/moonlight-common-c/enet/CMakeFiles/enet.dir/callbacks.c.obj
In file included from C:/Users/ReenigneArcher/Documents/GitHub/LizardByte/mnxb/third-party/moonlight-common-c/enet/callbacks.c:6:
In file included from C:/Users/ReenigneArcher/Documents/GitHub/LizardByte/mnxb/third-party/moonlight-common-c/enet/include/enet/enet.h:16:
In file included from C:/Users/ReenigneArcher/Documents/GitHub/LizardByte/mnxb/third-party/moonlight-common-c/enet/include/enet/win32.h:20:
In file included from C:/Users/ReenigneArcher/Documents/GitHub/LizardByte/mnxb/third-party/nxdk/lib/winapi/ws2_32/winsock2.h:12:
In file included from C:/Users/ReenigneArcher/Documents/GitHub/LizardByte/mnxb/third-party/nxdk/lib/net/lwip/src/include/lwip/sockets.h:42:
C:/Users/ReenigneArcher/Documents/GitHub/LizardByte/mnxb/third-party/nxdk/lib/net/lwip/src/include/lwip/opt.h:51:10: fatal error:
      'lwipopts.h' file not found
   51 | #include "lwipopts.h"
      |          ^~~~~~~~~~~~
1 error generated.
make[2]: *** [third-party/moonlight-common-c/enet/CMakeFiles/enet.dir/build.make:75: third-party/moonlight-common-c/enet/CMakeFiles/enet.dir/callbacks.c.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:225: third-party/moonlight-common-c/enet/CMakeFiles/enet.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```

[...]

PR #771, opened 2026-05-20
Add SDL_net support

PaulMDemers:
## Summary

Adds SDL_net support to nxdk as an SDL2 companion library, backed by nxdk’s existing lwIP network stack.

This adds SDL_net as a submodule, builds it into `libSDL2_net.lib` when `NXDK_SDL = y`, exposes a pkg-config file, and includes a small smoke-test sample that resolves localhost and opens a UDP socket.

## Changes

- Add `lib/sdl/SDL_net` submodule
- Add nxdk/lwIP compatibility patch in the SDL_net fork
- Build `libSDL2_net.lib` from SDL_net sources
- Add `lib/pkgconfig/SDL2_net.pc`
- Add `samples/sdl_net`
- Update README SDL library summary

## Verification

Built successfully on Windows/MSYS2 with:

```bash
make -C samples/sdl_net
```

This produced:
- lib/libSDL2_net.lib
- samples/sdl_net/bin/default.xbe
- samples/sdl_net/nxdk sample - SDL2_net.iso

## Notes
The SDL_net submodule points at PaulMDemers/SDL_net, branch nxdk-sdl-net, which is based on SDL_net 2.0.1 with minimal nxdk compatibility changes for lwIP sockets.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,3 +42,7 @@
 [submodule "lib/usb/libusbohci"]
 	path = lib/usb/libusbohci
 	url = https://github.com/XboxDev/libusbohci.git
+[submodule "lib/sdl/SDL_net"]
+	path = lib/sdl/SDL_net
+	url = https://github.com/PaulMDemers/SDL_net.git
+	branch = nxdk-sdl-net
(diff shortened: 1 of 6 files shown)


JayFoxRox:
I don't think this should be merged. See reasons in https://github.com/XboxDev/nxdk/pull/601 

---

I had a port of enet at some point, and [there was work towards a dummy winsock (based on lwip)](https://github.com/XboxDev/nxdk/pull/680#issuecomment-2424001340) which was enough for enet (I believe). @thrimbor also had a winsock impl. It should probably be extended to support the upstream SDL_net without changes, too.

Such a winsock impl would be a much better fit for core nxdk.

Issue #453, opened 2021-04-11
Winsock implementation


JayFoxRox:
*(I'm very sure we discussed this in the past, but I'm unable to find anything when searching for "winsock" on this repository; please close if I missed an existing issue)*

I had started a rather dummy ws2_32.lib implementation about a year ago and I believe @thrimbor had also worked on something in that direction?
A real implementation would be nice, but a partially working dummy is also fine with me, just to get the ball rolling.

Specifically, my winsock was exclusively meant for enet, and it can be found here: https://github.com/JayFoxRox/nxdk/pull/84.
At the time of writing, the latest version is in https://github.com/JayFoxRox/nxdk/commit/f3dd88898180e6e59d6f42956c346615e335208e.
I don't think I ever really tested it, so it's probably non-functional.

It would be nice if someone could find a suitable implementation or clean up the mess from my commit.
Consider my code in that commit to be in public domain.

ReenigneArcher:
I would like this as well as I would like to use enet (fork from https://github.com/cgutman/enet/tree/moonlight) in a project I'm attempting. I had a look at your previous attempt, but for now it's a bit above my head to get to a working state.

I was wondering if it might be easier to compile enet as `UNIX` instead of `WIN32`, and then re-implement the required parts from the Linux source. It'd still be above my head, but maybe it's easier since it's open source.

JayFoxRox:
The use of `WSASendMsg` in that fork is probably problematic. It is used at least in https://github.com/cgutman/enet/commit/e790647a577fb1402b5a25a479a03d4045f5aa44#diff-14d02ab34000ddb786c1258055d552d6b6613c608e2e515096fdc8abe26e9e90R474
It's rather advanced and I'm not sure if lwip (or my code) supports it - I doubt it. It's mostly relevant for multiple network addresses, application specific routing (like tracing) or IPv6.
You could probably implement some `WSASendMsg` which just ignores some fields (for now).

Aside from that, the UNIX and WIN32 code-path should only differ slightly. Typically WIN32 networking only adds winsock specific init and cleanup and some utility functions related to timing or buffer allocations - all of which nxdk probably already supports.

I don't remember, but my winsock implementation probably also merely stubs the init/cleanup and hooks up the Winsock BSD-style socket API to the lwip BSD-style socket interface.

PR #773, opened 2026-05-27
libc: Add POSIX headers & stubs

ReenigneArcher:
These changes were required to compile https://github.com/moonlight-stream/moonlight-common-c along with the following dependencies using nxdk.

- https://github.com/cgutman/enet
- https://github.com/openssl/openssl/tree/OpenSSL_1_1_1-stable

I am using this in https://github.com/LizardByte/Moonlight-XboxOG

### DIFF
--- a/lib/net/netinet/in.h
+++ b/lib/net/netinet/in.h
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+/*
+ * Provide the BSD netinet/in.h include path expected by Unix-oriented
+ * third-party libraries while reusing nxdk's lwIP socket compatibility layer.
+ */
+
+#include <sys/socket.h>
--- a/lib/net/netinet/tcp.h
+++ b/lib/net/netinet/tcp.h
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+/*
+ * Provide the BSD netinet/tcp.h include path expected by Unix-oriented
+ * third-party libraries while reusing nxdk's lwIP socket compatibility layer.
+ */
+
+#include <sys/socket.h>
--- a/lib/xboxrt/libc_extensions/conio.h
+++ b/lib/xboxrt/libc_extensions/conio.h
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+/*
+ * Provide the tiny conio.h subset that some cross-built utility code expects.
+ * The Xbox runtime does not provide a console keyboard in these call sites, so
+ * the helpers intentionally report no pending input.
+ */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static inline int _kbhit(void)
+{
+    return 0;
+}
+
+static inline int _getch(void)
+{
+    return -1;
+}
+
+static inline int kbhit(void)
+{
+    return _kbhit();
+}
+
+static inline int getch(void)
+{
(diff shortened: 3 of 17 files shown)


glebm:
This stubs out a bunch of functions like opendir/readdir/closedir/read/write/lseek. Even if a program using these functions compiles, it probably won't work. Probably not a good idea to pretend functions are available when they're not?

ReenigneArcher:
Yet it's working here -> https://github.com/LizardByte/Moonlight-XboxOG

glebm:
I guess it's working for you because your program doesn't call these functions (for sure, never? then why are they there?).

The main issue is that programs that detect availability of these functions and use them when available would break in a bad way.

ReenigneArcher:
Correct, because moonlight-common-c needs just a tiny part of OpenSSL, not every possible thing it can do. Typically OpenSSL is just available from the distro or package managers so for Moonlight no one ever had to worry about compiling it manually... and of course you can't expect a project like that to care about supporting NXDK.

glebm:
nxdk implements many Windows APIs, it's basically a Windows platform with API level similar to Windows 9x. Consider the `ossl_time_now` function from OpenSSL:

https://github.com/openssl/openssl/blob/33482fbddcf137dde9bd84ae12f2f02cbc358c78/crypto/time.c#L15-L49

With `#if defined(_WIN32)`, it uses `GetSystemTime` and `SystemTimeToFileTime`, both of which are availble in nxdk.
On the other hand, the `gettimeofday` stub here will result in the function always returning 0 time, possibly resulting in all kinds of shenanigans.

Similarly for `LP_find_file`. While it uses `opendir` on POSIX, which is stubbed out in this PR, in the Windows implementation (https://github.com/openssl/openssl/blame/master/crypto/LPdir_win.c) it uses `FindFirstFile` and friends which are already available in nxdk. Here, you will need to add a condition to not always `#define LP_MULTIBYTE_AVAILABLE` in https://github.com/openssl/openssl/blame/master/crypto/LPdir_win32.c (e.g. by making it a configure option).

You'd have better luck starting from the Windows route and adding the missing APIs there.

This might still require patches, though it's sometimes possible to upstream them. E.g. I was able to upstream this option to libfmt because Windows 9x also lacks the corresponding API: https://github.com/fmtlib/fmt/pull/3636

I'm not opposed to a POSIX compatibility layer to make porting easier but only if it's actually working and not stubbed out. I'm not a maintainer of nxdk, just a user, but it'd be very frustrating to use an SDK that declares POSIX functions without actually implementing them.

ReenigneArcher:
> NXDK implements many Windows APIs, it's basically a Windows platform with API level similar to Windows 9x.

Unfortunately, a very incomplete Windows API. First missing thing I found when going down that route was https://github.com/XboxDev/nxdk/issues/453

This was the easier path to get my project working.

I can look into trying to complete some of the stubs in the next few weeks, but I don't want to waste time on it if a maintainer isn't going to review or consider this.

antonic901:
There is [nxio](https://github.com/astarivi/nxio) from astarivi who is helping me with XBMC development. He wrote that because we were unable to compile curl in NXDK. I don't know how much it's completed, but it was enough to compile and get curl woking NXDK.

dracc:
> There is [nxio](https://github.com/astarivi/nxio) from astarivi who is helping me with XBMC development. He wrote that because we were unable to compile curl in NXDK. I don't know how much it's completed, but it was enough to compile and get curl woking NXDK.

Neat! Thanks for sharing.

PR #775, opened 2026-05-28
pbkit: Dynamically compute PTIMER parameters

### LINKED ISSUE #536: pbkit quits if the nvclk freq is not default

thrimbor:
Split out from other pbkit changes, original commit is https://github.com/thrimbor/nxdk/commit/3a1faf33d493daa70df7110197a2e277ac9e345d

It's what I consider a proper solution for the PTIMER initialization issue, deriving the parameters from the chip register values like https://github.com/XboxDev/nxdk/pull/695 did for the performance counters, instead of only using them for checking assumptions like it is now.
On a stock Xbox it results in a 31.25MHz PTIMER frequency like the XDK has hard-coded, and it should manage to configure the same frequency on an overclocked system, though I only was able to test the former.

Fixes https://github.com/XboxDev/nxdk/issues/536
Supersedes https://github.com/XboxDev/nxdk/pull/621, which I don't consider good enough because it only removes the check without attempting to handle what it checked for.

### DIFF
--- a/lib/pbkit/pbkit.c
+++ b/lib/pbkit/pbkit.c
@@ -2180,10 +2180,38 @@ void pb_set_color_format(unsigned int fmt, bool swizzled)
     assert(swizzled == false);
 }
 
+/**
+ * Helper function to calculate numerator and denominator for PTIMER
+ *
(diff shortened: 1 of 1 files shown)


GXTX:
Looks good.

Issue #706, opened 2025-02-18
[REQUST] implementation of WideCharToMultiByte, MultiByteToWideChar and vswprintf


antonic901:
I need those function for backport of CharsetConverter from Kodi. Source code of that file is [here](https://github.com/xbmc/xbmc/blob/master/xbmc/platform/win32/CharsetConverter.cpp).

I also need `vswprintf` for wchar variant of StringUtils::Format(...). Look [here](https://github.com/xbmc/xbmc/blob/Nexus/xbmc/utils/StringUtils.cpp#L300)

- [ ] WideCharToMultiByte
- [ ] MultiByteToWideChar
- [ ] vswprintf

PR #781, opened 2026-09-18
winapi: add WideCharToMultiByte and MultiByteToWideChar (UTF-8)

MartinAlvarado1895:
Partial fix for #706.
Adds `winnls.c` / `winnls.h` with `WideCharToMultiByte` and `MultiByteToWideChar` for code page 65001 (UTF-8) and code page 1 (kodi4xbox convention). Wired into `libwinapi` via `Makefile` and `windows.h`.
Does not implement `vswprintf`, 'wcstombs' yet.
/cc @antonic901


lib/winapi/winnls.h
```diff
@@ -0,0 +1,33 @@
...
+#include <winnt.h>
+
+#ifndef CP_UTF8
+#define CP_UTF8 65001
+#endif
+
+/* kodi4xbox unimplemented.h uses 1 for UTF-8 */
+#ifndef CP_UTF8_NXDK_KODI
+#define CP_UTF8_NXDK_KODI 1
+#endif
```
antonic901: You don't need this. I used 1 only because it was placeholder for me so code compiles.

antonic901 (changes requested):
I can't say more about actual implementation, someone else from XboxDev should review this.

lib/winapi/winnls.h
MartinAlvarado1895: Thanks for the review. I misunderstood the 1: I thought NXDK had to accept kodi4xbox’s CP_UTF8 value. You used 1 only as a placeholder in unimplemented.h so Kodi would compile.

I removed CP_UTF8_NXDK_KODI. This PR now only implements WinAPI CP_UTF8 (65001). A call with code page 1 returns 0.

I tested the conversion on xemu: WideCharToMultiByte / MultiByteToWideChar with "Canción" (ó = U+00F3) produce UTF-8 43 61 6E 63 69 C3 B3 6E.

After this is merged, kodi4xbox can drop the dummy 1 and use CP_UTF8 (65001).

One question: I opened this against XboxDev/nxdk because WideCharToMultiByte is missing WinAPI, not a Kodi-only patch. I see kodi4xbox builds with the xbmc4xbox/nxdk fork (exceptions / other WIP). Should this stay only on XboxDev and get pulled into that fork later, or do you also want a PR on xbmc4xbox/nxdk?

PR #782, opened 2026-09-22
xboxkrnl: Fix ExInterlockedAddLargeInteger() missing spinlock parameter

thrimbor:
Found while looking through the commits in https://github.com/XboxDev/nxdk/pull/770

There is a mismatch between the xboxkrnl header, which declares 12 bytes of parameters for `ExInterlockedAddLargeInteger`, and our import library, which specifies 16 bytes, leading to a linker error when the function is used.
https://github.com/XboxDev/nxdk/pull/770/changes/13a3a02cf816a9c480f7c78d35a375fe622feb61 attempted to fix it by changing the import library to 12 bytes. I've double-checked this change and found it to be incorrect, verified by comparing the stack pointer before and after the call and by disassembling the exported kernel function.
image
As can be seen in the screenshot, the function return pops 16 bytes from the stack (not including the return address). This means that despite the spinlock parameter being unused (since disabling interrupts is sufficient on non-SMP systems), it is required to be passed.

This PR adds the missing parameter and the required typedef, resolving the linker error while keeping the stack intact when the function is called.


thrimbor:
cc @PatrickvL since this affects https://github.com/Cxbx-Reloaded/xbox_kernel_test_suite/pull/112

PatrickvL:
@RadWolfie have you seen this?

RadWolfie:
I had checked my local branch for this API change, which refreshed my memory that Patrick's test caused a hardware crash for an unknown reason. I had not thoroughly checked this API, so I disabled it.

Anyway, I took a look at the official doc for [ExInterlockedAddLargeInteger](https://learn.microsoft.com/en-us/previous-versions/ff545335(v=vs.85)) which indeed has 3 parameters. The first parameter is 4 bytes, the second parameter is 8 bytes, and the third parameter is 4 bytes. That sums up to 16 bytes, which matches the `@16`. It is safe to say that all of the parameters follow the doc.

Apparently, the third parameter is not used in the hardware kernel.

If you are requesting my approval without test verification, I approve this PR change. 👍

---
Nit to the screenshot: The stack does not match with the pushed parameters for some reason. The stack should be:
Stack[0x4]:4 (correct)
Stack[0x8]:8 (image says 0x14, incorrect)
Stack[0x10]:4 (image says 0x1C, incorrect)

Base on comparison from thrimbor and my Ghidra analysis results indeed shows the same thing. I had to check the custom storage and correct the error from Ghidra's analysis.

updated image ref:
image

